PR #11, opened 2021-10-19
S108 yaml rules

eschultink:
### Features
 - [connector rules to JSON or something](https://app.asana.com/0/1201039336231823/1201118215070899/f) - wip; idea is that if deployed with `rules.yaml` file, will read that in lieu of choosing java class - allowing customers additional control. 

To do:
 - [x] figure out how to test locally with a file in the file system - actually more straightforward how to do it in production, with JAR deployment (as just add it to `deployment/` directory)

### Change implications

 - dependencies added/changed? **yes** - Jackson yaml support


java/core/src/test/java/co/worklytics/psoxy/RulesTest.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @SneakyThrows
+    @Test
+    public void yaml_file() {
+        byte[] data = TestUtils.getData("rules/example.yaml");
+
+        ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory());
+
+        Rules fromYaml = objectMapper.readerFor(Rules.class)
```
jlorper, line 64: this test is basically the same: is not testing any rule but just that the object mapper can read from bytes as opposed to String like in the first test.

java/impl/gcp/src/main/java/co/worklytics/psoxy/Route.java
```diff
@@ -51,10 +55,13 @@
      * use should be overridden with something
      */
     static final String DEFAULT_SALT = "salt";
+    static final String PATH_TO_RULES_FILES = "/rules.yaml";
 
-    ConfigService config;
+    @Getter
+    ConfigService config  = new EnvVarsConfigService();;
```
jlorper, line 61: extra semicolon
```suggestion
    ConfigService config  = new EnvVarsConfigService();
```

jlorper (approved):
A couple minor things

java/core/src/test/java/co/worklytics/psoxy/RulesTest.java
eschultink: yeah, and just that no encoding issues cause weirdness with the regexes.

PR #22, opened 2021-11-16
s110 : examples of setting up GCP as terraform state backend

eschultink:
### Features
 - [example of using a GCS bucket, with CMEK, to store your state](https://app.asana.com/0/1201039336231823/1201367423324936/f)

### Change implications

 - dependencies added/changed? **yes** -- various terraform modules, most critically google/hashicorp bootstrap  : https://registry.terraform.io/modules/terraform-google-modules/bootstrap/google/latest

### DIFF
--- a/.github/workflows/ci-terraform.yaml
+++ b/.github/workflows/ci-terraform.yaml
@@ -1,7 +1,11 @@
 name: CI - infra (terraform)
 
 on:
-  [push]
+  push:
+    branches:
+      - '**' # should match all branches
+    tags-ignore:
+      - '**' # all tags; actually, believe this is default behavior when you specify filter for branches but not tags
 
 jobs:
   validate:
--- a/.github/workflows/ci-terraform.yaml
+++ b/.github/workflows/ci-terraform.yaml
@@ -23,3 +23,21 @@ jobs:
         run: |
           terraform init
           terraform validate
+
+      - name: 'Terraform - validate example-google-workspace'
+        working-directory: infra/example-google-workspace
+        run: |
+          terraform init
+          terraform validate
+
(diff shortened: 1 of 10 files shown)


.github/workflows/ci-terraform.yaml
```diff
@@ -23,3 +27,21 @@ jobs:
         run: |
           terraform init
           terraform validate
+
+      - name: 'Terraform - validate example-google-workspace'
+        working-directory: infra/example-google-workspace
+        run: |
+          terraform init
```
aperez-worklytics, line 34: Let's see how it grows, but maybe we could have a `terraformValidate.sh` shell/bash script and just do:

```
- name: 'Terraform - validate example-gcp-simple'
       working-directory: infra/example-gcp-simple
       run: ./scripts/terraformValidate.sh
```

To avoid repeting same pair of commands again and again.

infra/example-bootstrap-gcp-simple/variables.tf
```diff
@@ -0,0 +1,34 @@
+variable "project_name" {
+  type        = string
+  description = "name to give to GCP project"
+  default     = null
```
aperez-worklytics, line 4: `null` ? Maybe better without default, so it will be enforced to have a value later

aperez-worklytics (approved):

infra/example-bootstrap-gcp-simple/variables.tf
eschultink: it'll use `coalesce()` to take the `project_id` value as the name if left `null`, which I think is the right thing to do. We don't need to force customer to specify a name for the project - but it's nice to give the option.

PR #25, opened 2021-11-18
s110 : add TODOs for Worklytics side of connections

eschultink:
### Features
 - [TODO instructions for Worklytics-side of psoxy connections](https://app.asana.com/0/1201368993699465/1201385129444887)

### Change implications

### DIFF
--- a/infra/modules/worklytics-psoxy-connection/main.tf
+++ b/infra/modules/worklytics-psoxy-connection/main.tf
@@ -1,0 +1,20 @@
+
+
+
+resource "local_file" "todo-google-workspace-admin-console" {
+filename = "TODO - connect ${var.display_name} in Worklytics .md"
+content  = <<EOT
+Complete the following steps in Worklytics AFTER you have deployed the Psoxy instance for your connection:
+  1.  Visit https://intl.worklytics.co/#integrations (or login into Worklytics, and navigate to
+      Manage --> Data Connections)
+  2.  Find the connector named "${var.display_name}", click "Add new connection"
+  3.  Copy and paste `${var.psoxy_endpoint_url}` as the value for "Psoxy Base URL".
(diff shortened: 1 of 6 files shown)


infra/modules/worklytics-psoxy-connection/main.tf
```diff
@@ -0,0 +1,20 @@
...
+Complete the following steps in Worklytics AFTER you have deployed the Psoxy instance for your connection:
+  1.  Visit https://intl.worklytics.co/#integrations (or login into Worklytics, and navigate to
+      Manage --> Data Connections)
+  2.  Find the connector named "${var.display_name}", click "Add new connection"
+  3.  Copy and paste `${var.psoxy_endpoint_url}` as the value for "Psoxy Base URL".
+  4.  Review any additional settings that connector supports, adjusting values as you see fit, then
+      click "Connect".
+
+Worklytics will attempt some basic health checks to ensure your Psoxy instance is reachable and
+configured correctly. If this fails, contact support@worklytics.co for guidance.
```
aperez-worklytics, line 16: Will the health check response be exposed at someplace for the customer? If something fails, the customer could see "*ok, it is failing because the health check cannot find the URL so I'm going to review the pasted URL to see if is the one that I should copy*". Just to have some basic covered cases that we don't require contacting us through support

aperez-worklytics (approved):

jlorper (approved):

infra/modules/worklytics-psoxy-connection/main.tf
eschultink: possibly; it returns some error codes to JS client; I am not sure what David ultimately implemented to show to customer. Certainly there is nothing sophisticated to take them to a page to change the URL in response to a URL endpoint error or something like that.

PR #27, opened 2021-11-18
s110: Slack allow API method

jlorper:
Adds a new endpoint that must be reachable for org-wide installations to read all workspaces to later iterate on them.

### Features
- [slack ediscovery connector, extensible via proxy ](https://app.asana.com/0/1201290353665663/1201109001473054)

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/rules/slack/PrebuiltSanitizerRules.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/rules/slack/PrebuiltSanitizerRules.java
@@ -13,6 +13,7 @@ import static co.worklytics.psoxy.Rules.Rule;
 public class PrebuiltSanitizerRules {
 
     static final Rules SLACK = Rules.builder()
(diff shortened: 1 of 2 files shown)


aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/rules/slack/PrebuiltSanitizerRules.java
```diff
@@ -13,6 +13,7 @@
 public class PrebuiltSanitizerRules {
 
     static final Rules SLACK = Rules.builder()
+        .allowedEndpointRegex("^\\/api\\/discovery\\.enterprise\\.info(?:\\?.+)?")
```
eschultink, line 16: test it, but I think all the `\\/` can just be `/` ... I don't think `/` needs to be escaped, as it seems to be treated as a regular character unless it appears in very specific contexts

(I was wrong initially to escape it; imho should remove, as just makes the patterns more verbose)

eschultink (approved):
eh, one thought for future but I'm going to merge. we can cleanup verbose regex later

PR #30, opened 2021-11-30
Gmail rules check case-insensitive

jlorper:
Handles matching of header names in a case insensitive way so different email providers behavior is captured.

### Fixes
- [Gmail psoxy rules set case insensitive](https://app.asana.com/0/1201438072827924/1201441972191372)

### Change implications


eschultink (approved):
perf of regexes scares me, but OK

PR #36, opened 2021-12-03
s111 - jdk8 compatible version of the psoxy core

jlorper:
Maven configuration to build a jdk8 compatible psoxy-core to use with our jdk8 projects

### Features
- [java8 maven build profile for psoxy-core artifact](Search.1201449695753355_search_1201460785569618)

### Change implications

### DIFF
--- a/java/core/pom.xml
+++ b/java/core/pom.xml
@@ -11,9 +11,38 @@
     </parent>
 
     <groupId>co.worklytics.psoxy</groupId>
-    <artifactId>psoxy-core</artifactId>
+    <artifactId>psoxy-core${artifactSuffix}</artifactId>
     <packaging>jar</packaging>
     <description>Code that's core to both Psoxy clients and implementations</description>
+
+    <properties>
+        <!-- empty for java 11 builds -->
+        <artifactSuffix></artifactSuffix>
+    </properties>
+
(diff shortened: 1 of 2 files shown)


java/core/pom.xml
```diff
@@ -11,10 +11,39 @@
...
 
     <groupId>co.worklytics.psoxy</groupId>
-    <artifactId>psoxy-core</artifactId>
+    <artifactId>psoxy-core${artifactSuffix}</artifactId>
     <packaging>jar</packaging>
     <description>Code that's core to both Psoxy clients and implementations</description>
 
+    <properties>
+        <!-- empty for java 11 builds -->
+        <artifactSuffix></artifactSuffix>
```
aperez-worklytics, line 20: I'd prefer to follow always same convention: `psoxy-core-jdkXX` instead of keeping it empty for jdk11

aperez-worklytics (approved):

java/core/pom.xml
eschultink: I'm OK with it empty ... jdk8 is really a special case for use; I prefer to keep it simple for users - they don't need to consider what jdk version they're building for, it should be implementation detail

eschultink (approved):

PR #42, opened 2022-01-12
S114 jose di for discussion

eschultink:
created just for discussion purposes, as to how to incorporate DI into main branch with all the aws stuff

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/HashUtils.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/HashUtils.java
@@ -1,11 +1,14 @@
 package co.worklytics.psoxy;
 
+import lombok.NoArgsConstructor;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang3.StringUtils;
 
+import javax.inject.Inject;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 
+@NoArgsConstructor(onConstructor = @__({@Inject}))
 public class HashUtils {
 
     public String hash(String... fragments) {
--- a/java/core/src/main/java/co/worklytics/psoxy/SanitizerFactory.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/SanitizerFactory.java
@@ -1,10 +1,11 @@
 package co.worklytics.psoxy;
 
+import co.worklytics.psoxy.impl.SanitizerImpl;
 import dagger.assisted.AssistedFactory;
 
 @AssistedFactory
 public interface SanitizerFactory {
 
-    Sanitizer create(Sanitizer.Options options);
+    SanitizerImpl create(Sanitizer.Options options);
 
 }
--- a/java/core/src/main/java/co/worklytics/psoxy/SanitizerFactory.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/SanitizerFactory.java
@@ -1,0 +1,10 @@
+package co.worklytics.psoxy;
+
+import dagger.assisted.AssistedFactory;
+
+@AssistedFactory
+public interface SanitizerFactory {
+
+    Sanitizer create(Sanitizer.Options options);
+
+}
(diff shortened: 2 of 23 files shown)


java/core/src/main/java/co/worklytics/psoxy/HashUtils.java
```diff
@@ -1,11 +1,14 @@
...
 
+import lombok.NoArgsConstructor;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang3.StringUtils;
 
+import javax.inject.Inject;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 
+@NoArgsConstructor(onConstructor = @__({@Inject}))
```
eschultink, line 11: why not `@NoArgsConstructor(onConstructor_ = @Inject)` ? cleaner imho

java/core/src/main/java/co/worklytics/psoxy/SanitizerFactory.java
```diff
@@ -0,0 +1,11 @@
...
+
+import co.worklytics.psoxy.impl.SanitizerImpl;
+import dagger.assisted.AssistedFactory;
+
+@AssistedFactory
+public interface SanitizerFactory {
+
+    SanitizerImpl create(Sanitizer.Options options);
+
+}
```
eschultink, line 11: 👍  yes, I'm porting this

java/core/src/main/java/co/worklytics/psoxy/di/PsoxyModule.java
```diff
@@ -0,0 +1,40 @@
+package co.worklytics.psoxy.di;
```
eschultink, line 1: I find naming package `di` like this is rather weird ... why not just put it at top-level?

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
```diff
@@ -33,9 +35,18 @@
 import java.util.stream.Collectors;
 
 @Log
-@RequiredArgsConstructor
 public class SanitizerImpl implements Sanitizer {
 
+    @AssistedInject
+    public SanitizerImpl(HashUtils hashUtils, @Assisted Options options) {
+        this.options = options;
```
eschultink, line 42: why doesn't this assign `hashUtils`? does that happen magically in `AssistedInject`?

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
```diff
@@ -107,9 +108,6 @@ public String sanitize(@NonNull URL url, @NonNull String jsonResponse) {
             compiledPseudonymizationsWithOriginals =
                 compile(options.getRules().getPseudonymizationWithOriginals());
         }
```
eschultink, line 110: blocks above are initialization, which should move into DI factory??

java/core/src/main/java/co/worklytics/psoxy/HashUtils.java
jlorper: yeah, is confusing that Lombok's doc still points to `@__({@Inject}))`
> To put annotations on the generated constructor, you can use onConstructor=@__({@AnnotationsHere}), but be careful; this is an experimental feature. For more details see the documentation on the onX feature.

But yes, that's JDK7 style, JDK8 way cleaner. I always get confused about what's what

java/core/src/main/java/co/worklytics/psoxy/di/PsoxyModule.java
jlorper: I prefer not to mix it with the "real" code. DI is just glue for the app, you can really avoid the package while scanning contents, so I prefer to put it in their on package. But whatever.

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
jlorper: yes. With `Assisted` you just tell the factory the dependencies that the caller must satisfy. Under the hood this is just like passing "options" to a `@Provides` method.

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
jlorper: No, should be moved to constructor. I'll do it.

PR #43, opened 2022-01-16
S113 aws test call

jlorper:
I believe `curl --aws-sigv4` doesn't work on asummed roles. Inspecting headers I found out it was not generating the signature properly. I found out this tool [awscurl](https://github.com/okigan/awscurl) that performs the signature in python and works. Some policies needed to be tweaked.

### Features
 - [Invoke from local via CURL](https://app.asana.com/0/1201637024614734/1201614440806034)

### Change implications

 - dependencies added/changed? **not to the project** (install awscurl if local tests wants to be performed)


eschultink (approved):
ok thanks 👍  looks good. only doubt is that your role change seems to be opening API Gateway for the ENTIRE account, not just a *specific* API gateway as was the intention before ... but should be OK, as people *should* be using dedicated AWS account for this AND it seems like API gateway is essentially global the account anyways ... anyone desiring more restrictive settings is free to customize the rule further to make it more restrictive should their setup require it.

PR #44, opened 2022-01-17
S113 aws poc

eschultink:
### Features
 - POC to support deployment of psoxy to AWS

### Change implications

 - dependencies added/changed? **yes (explain)** - AWS bom, various libraries required for that; and indeed an entirely new module (`aws`) with separate build + pom to review


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -0,0 +1,209 @@
...
+    final int SOURCE_API_REQUEST_READ_TIMEOUT = 300_000;
+
+    @Inject ConfigService config;
+    @Inject RulesUtils rulesUtils;
+    @Inject SourceAuthStrategy sourceAuthStrategy;
+    @Inject ObjectMapper objectMapper;
+    @Inject SanitizerFactory sanitizerFactory;
+    @Inject Rules rules;
+
+    private Sanitizer sanitizer;
```
aperez-worklytics, line 42: Variable is global for the class, but only used in `handle` L48, L113 and in L114. Maybe it can be local for the method

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.*;
+import java.util.logging.Level;
+import java.util.stream.Collectors;
+
+@NoArgsConstructor(onConstructor_ = @Inject)
+@Log
+public class CommonRequestHandler {
+
+    //we have ~540 total in Cloud Function connection, so can have generous values here
+    final int SOURCE_API_REQUEST_CONNECT_TIMEOUT = 30_000;
```
aperez-worklytics, line 32: I'd add the unit here as part of the name; are they milliseconds right ?

```suggestion
    final int SOURCE_API_REQUEST_CONNECT_TIMEOUT_MILLISECONDS = 30_000;
```

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/GoogleCloudPlatformServiceAccountKeyAuthStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+
+@Log
+@NoArgsConstructor(onConstructor_ = @Inject)
+public class GoogleCloudPlatformServiceAccountKeyAuthStrategy implements SourceAuthStrategy {
+
+    @Getter
+    private final String configIdentifier = "gcp_service_account_key";
```
aperez-worklytics, line 25: It will always have same value, so I'd put it static too:

```suggestion
    private final static String configIdentifier = "gcp_service_account_key";
```

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -0,0 +1,209 @@
...
+                objectMapper.writeValueAsString(healthCheckResult) + "\r\n");
+        } catch (IOException e) {
+            log.log(Level.WARNING, "Failed to write health check details", e);
+        }
+
+        return responseBuilder.build();
+    }
+
+    boolean isSuccessFamily (int statusCode) {
+        return statusCode >= 200 && statusCode < 300;
```
aperez-worklytics, line 197: `Response.Status.Family.familyOf` from `javax.ws.rs.core` can be used here ?

aperez-worklytics (commented):
Still in review, I'll continue later. Just some feedback about format/naming. :+1:

infra/modules/aws-psoxy-instance/main.tf
```diff
@@ -0,0 +1,197 @@
...
+  memory_size      = 512 # megabytes
+
+  environment {
+    # NOTE: can use merge() to combine var map from config with additional values
+    variables = yamldecode(file(var.path_to_config))
+  }
+}
+
+
+// User: arn:aws:sts::874171213677:assumed-role/iam_for_lambda/psoxy-gdirectory is not authorized to perform:  on resource: arn:aws:ssm:us-east-1:874171213677:parameter/RULES because no identity-based policy allows the ssm:GetParameter action
```
jlorper, line 87: remove

infra/modules/aws-psoxy-instance/main.tf
```diff
@@ -0,0 +1,197 @@
...
+aws sts assume-role --duration 900 --role-arn ${var.aws_assume_role_arn} --role-session-name deploy-lambda > temporal-credentials.json
+export CALLER_ACCESS_KEY_ID=`cat temporal-credentials.json| jq -r '.Credentials.AccessKeyId'`
+export CALLER_SECRET_ACCESS_KEY=`cat temporal-credentials.json| jq -r '.Credentials.SecretAccessKey'`
+export CALLER_SESSION_TOKEN=`cat temporal-credentials.json| jq -r '.Credentials.SessionToken'`
+rm temporal-credentials.json
+aws sts get-caller-identity
+
+aws lambda update-function-code --function-name ${var.function_name} --zip-file fileb://target/psoxy-aws-1.0-SNAPSHOT.jar
+
+unset CALLER_ACCESS_KEY_ID CALLER_SECRET_ACCESS_KEY CALLER_SESSION_TOKEN
```
jlorper, line 154: oops, this is wrong... my bad. These need to be 
```suggestion
export AWS_ACCESS_KEY_ID=`cat temporal-credentials.json| jq -r '.Credentials.AccessKeyId'`
export AWS_SECRET_ACCESS_KEY=`cat temporal-credentials.json| jq -r '.Credentials.SecretAccessKey'`
export AWS_SESSION_TOKEN=`cat temporal-credentials.json| jq -r '.Credentials.SessionToken'`
rm temporal-credentials.json
aws sts get-caller-identity

aws lambda update-function-code --function-name ${var.function_name} --zip-file fileb://target/psoxy-aws-1.0-SNAPSHOT.jar

unset AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN
```
The deploy does use the AWS env variables to work.
I renamed that to CALLER_X when testing because there is no real reason to overwrite the environment (as passing it as env variables didn't work in `awscurl`)

infra/modules/aws/main.tf
```diff
@@ -0,0 +1,107 @@
...
+  assume_role_policy = jsonencode({
+    "Version" : "2012-10-17",
+    "Statement" : [
+      {
+        "Action" = "sts:AssumeRole"
+        "Effect" : "Allow"
+        "Principal" : {
+          "AWS" : "arn:aws:iam::${var.caller_aws_account_id}"
+        }
+        #TODO: add condition referencing GCP service account that will auth with Worklytics AWS account to call
```
jlorper, line 46: watch out, if adding this the test call won't work. We'd need to add another policy for the GCP case.

infra/modules/aws/variables.tf
```diff
@@ -0,0 +1,16 @@
...
+
+#eg "780C7DE5BBF9127"
+variable "caller_aws_user_id" {
+  type        = string
+  description = "id of service account that will call proxy (eg, SA of your worklytics instance)"
+}
+
+variable "aws_account_id" {
+  type        = string
+  description = "account id of the organization"
```
jlorper, line 15:
```suggestion
  description = "account id of the organization, 12-digit number"
```

infra/modules/aws/variables.tf
```diff
@@ -0,0 +1,16 @@
+variable "caller_aws_account_id" {
+  type        = string
+  description = "id of worklytics sa"
```
jlorper, line 3: 12-digit number
```suggestion
  description = "id of worklytics sa, 12-digit number and username like 123456789012:username "
```

java/core/src/main/java/co/worklytics/psoxy/PsoxyModule.java
```diff
@@ -0,0 +1,75 @@
...
+    //q: is this platform dependent?
+    @Provides
+    HttpRequestFactory providesHttpRequestFactory() {
+        return (new NetHttpTransport()).createRequestFactory();
+    }
+
+    @Provides
+    static Logger logger() {
+        return Logger.getLogger(PsoxyModule.class.getCanonicalName());
+    }
```
jlorper, line 62: why don't we just use lombok's `@Log(topic="psoxy")`  (if we want a topic)

java/core/src/main/java/co/worklytics/psoxy/PsoxyModule.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Provides
+    static SourceAuthStrategy sourceAuthStrategy(ConfigService configService, Set<SourceAuthStrategy> sourceAuthStrategies) {
```
jlorper, line 65: :warning:  Is this OK? `SourceAuthStrategy` is connector dependent, not global, isn't it?
I recall I put this in a `SourceAuthStrategyFactory` that get's called depending on the connector

java/core/src/main/java/co/worklytics/psoxy/SanitizerFactory.java
```diff
@@ -0,0 +1,27 @@
...
+    String DEFAULT_SALT = "salt";
+
+    SanitizerImpl create(Sanitizer.Options options);
+
+    //q: right place? mapping config+rules --> Sanitizer.Options isn't implementation-specific
+    default Sanitizer.Options buildOptions(ConfigService config, Rules rules) {
+        return Sanitizer.Options.builder()
+            .rules(rules)
+                    .pseudonymizationSalt(config.getConfigPropertyAsOptional(ProxyConfigProperty.PSOXY_SALT)
+                        .orElse(DEFAULT_SALT))
```
jlorper, line 21: I'd say this is critical enough to not have a DEFAULT_SALT... it should "explode" here if no PSOXY_SALT

java/core/src/main/java/co/worklytics/psoxy/gateway/HttpEventResponse.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.Builder;
+import lombok.Singular;
+import lombok.Value;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.List;
+
+/**
+ *
+ * q: split interface and implementation? What's the point?
```
jlorper, line 12: easily mock, use fake implementations, etc...

java/core/src/main/java/co/worklytics/psoxy/gateway/ProxyRequestAdapter.java
```diff
@@ -0,0 +1,27 @@
...
+     * The query part of the URI for this request. If the full URI is
+     * {@code http://foo.com/bar/baz?this=that}, then this method will return {@code this=that}.
+     * If there is no query part, the returned {@code Optional} is empty.
+     *
+     * @return the query part of the URI, if any.
+     */
+    Optional<String> getQuery(R request);
+
+
+    Optional<List<String>> getHeader(R request, String headerName);
```
jlorper, line 26: maybe simplify things with two methods
`Optional<String> getHeader(R request, String headerName);`
`Optional<List<String>> getMultiValuedHeader(R request, String headerName);`

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -0,0 +1,209 @@
...
+
+        //TODO: what headers to forward???
+
+        //setup request
+        sourceApiRequest
+            .setThrowExceptionOnExecuteError(false)
+            .setConnectTimeout(SOURCE_API_REQUEST_CONNECT_TIMEOUT)
+            .setReadTimeout(SOURCE_API_REQUEST_READ_TIMEOUT);
+
+        //q: add exception handlers for IOExceptions / HTTP error responses, so those retries
```
jlorper, line 97: No, we have the retry logic in Worklytics adapting to each case. IMHO Proxy should be _dumb_ in that regard.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -0,0 +1,209 @@
...
+        // eg, OAuth2CredentialsWithRefresh.newBuilder(), etc ..
+
+        //assume that in cloud function env, this will get ours ...
+        Optional<String> accountToImpersonate =
+           request.getHeader(ControlHeader.USER_TO_IMPERSONATE.getHttpHeader())
+                .map(values -> values.stream().findFirst().orElseThrow());
+
+        accountToImpersonate.ifPresentOrElse(
+            user -> log.info("Impersonating user"),
+            () -> log.warning("We usually expect a user to impersonate"));
```
jlorper, line 146: in GCP sources, not in others. This might be logging too much

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CompositeConfigService.java
```diff
@@ -0,0 +1,47 @@
...
+ * 'preferred' would be taken in such a scenario
+ *
+ */
+@Builder
+@RequiredArgsConstructor(access = AccessLevel.PACKAGE)
+public class CompositeConfigService implements ConfigService {
+
+    //open to feedback on these names;
+    @NonNull
+    final ConfigService preferred;
```
jlorper, line 26:
```suggestion
    final ConfigService primary;
```

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CompositeConfigService.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * checks 'preferred' first, and then fallback if not found there.
+ *
+ * although in practice we don't expect a given property to be defined in both, the value from
+ * 'preferred' would be taken in such a scenario
+ *
+ */
+@Builder
+@RequiredArgsConstructor(access = AccessLevel.PACKAGE)
+public class CompositeConfigService implements ConfigService {
```
jlorper, line 22: Not sure if this is a good idea. I get the point of simplifying the config access, but maybe the AWS implementation needs to be aware of what comes from global and what comes from lambda config to avoid misconfigurations (i.e. a global overwriting lambda, or f.e. the SALT not being in the global and lambdas with different SALTs? )

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/GoogleCloudPlatformServiceAccountKeyAuthStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+        Set<String> scopes = Arrays.stream(config.getConfigPropertyOrError(SourceAuthConfigProperty.OAUTH_SCOPES).split(","))
+            .collect(Collectors.toSet());
+        GoogleCredentials credentials;
+
+        Optional<String> key = config.getConfigPropertyAsOptional(SourceAuthConfigProperty.SERVICE_ACCOUNT_KEY);
+
+        if (key.isPresent()) {
+            credentials = ServiceAccountCredentials.fromStream(new ByteArrayInputStream(Base64.getDecoder().decode(key.get())));
+        } else {
+            credentials = GoogleCredentials.getApplicationDefault();
```
jlorper, line 42: q: this only works in GCP deployments? if so, maybe we'd need a way to check it

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
```diff
@@ -45,21 +39,19 @@
...
-            .mappingProvider(new JacksonMappingProvider())
-            .setOptions(Option.SUPPRESS_EXCEPTIONS); //we specifically want to ignore PATH_NOT_FOUND cases
+    @AssistedInject
+    public SanitizerImpl(HashUtils hashUtils, @Assisted Options options) {
+        this.hashUtils = hashUtils;
+        this.options = options;
     }
 
+    @Getter(onMethod_ = {@VisibleForTesting})
+    @Inject Configuration jsonConfiguration;
```
jlorper, line 49: hmm, maybe this should be part of the `@AssistedInject` constructor.
`public SanitizerImpl(HashUtils hashUtils, Configuration jsonConfiguration, @Assisted Options options) {`
Not sure if it's going to work when creating through the factory, I bet jsonConfiguration will be null.

java/core/src/test/java/co/worklytics/psoxy/rules/RulesBaseTestCase.java
```diff
@@ -29,12 +34,33 @@
...
+        MockModules.ForConfigService.class,
+    })
+    public interface Container {
+        void inject(RulesBaseTestCase test);
+    }
 
     @BeforeEach
     public void setup() {
-        sanitizer = new SanitizerImpl(Sanitizer.Options.builder()
+        //q: how to avoid boilerplate of this at top of every test?
```
jlorper, line 56: yeah, I don't know... that's one of the things I'd like to dig more about

java/impl/aws/src/main/java/co/worklytics/psoxy/CoreModule.java
```diff
@@ -0,0 +1,91 @@
...
+                    String source = config.getConfigPropertyOrError(ProxyConfigProperty.SOURCE);
+                    log.info("using prebuilt rules for: " + source);
+                    return PrebuiltSanitizerRules.MAP.get(source);
+                });
+            });
+
+        SanitizerImpl sanitizer = new SanitizerImpl(
+            Sanitizer.Options.builder()
+                .rules(rules)
+                .pseudonymizationSalt(config.getConfigPropertyAsOptional(ProxyConfigProperty.PSOXY_SALT)
```
jlorper, line 77: again, I think we should have the same as in Worklytics (`getConfigPropertyOrError`) for mandatory config

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/LambdaRequest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    String body;
+
+    @Override
+    public Optional<String> getQuery() {
+        if (this.getQueryParameters() == null) {
+            return Optional.empty();
+        } else {
+            String value = this.getQueryParameters().entrySet().stream()
+                .flatMap(parameter -> parameter.getValue().stream().map(v -> parameter.getKey() + "=" + v))
```
jlorper, line 43: does this work as expected for multi-valued query parameters? Is there a test to cover this?

jlorper: Probably safer to use `URIBuilder` and then extract the query string

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
```diff
@@ -0,0 +1,60 @@
...
+@Log
+@AllArgsConstructor
+public class ParameterStoreConfigService implements ConfigService {
+
+
+    String namespace;
+
+    @Inject SsmClient client;
+
+    //TODO: add caching?? guava?? apache commons??
```
jlorper, line 36: I'd love to work on all kind of improvements here.

java/impl/aws/src/test/java/co/worklytics/psoxy/HandlerTest.java
```diff
@@ -0,0 +1,105 @@
...
+        byte[] data = TestUtils.getData("request.json");
+        LambdaRequest event = objectMapper.readerFor(LambdaRequest.class)
+            .readValue(data);
+
+        Handler handler = new Handler();
+
+       // String result = handler.handleRequest(event, context);
+    }
+
+    Context context = new Context() {
```
jlorper, line 37: what's this for? Looks like a mock

java/impl/aws/src/test/resources/lambda-proxy-events/gdirectory-proxy-example.json
```diff
@@ -0,0 +1,122 @@
+{
+  "body": "",
+  "resource": "/{proxy+}",
+  "path": "/admin/directory/v1/customer/my_customer/domains",
+  "httpMethod": "POST",
+  "isBase64Encoded": true,
+  "queryStringParameters": {
+  },
+  "multiValueQueryStringParameters": {
+  },
```
jlorper, line 10: hmm, here they are split..

java/impl/cmd-line/src/main/java/co/worklytics/psoxy/FileHandler.java
```diff
@@ -0,0 +1,104 @@
...
+import javax.inject.Inject;
+import java.io.File;
+import java.io.FileReader;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+
+//@NoArgsConstructor(onConstructor_ = @Inject) //q: compile complaints - lombok annotation processing not reliable??
```
jlorper, line 24: if you have injects inside there is no need to add `(onConstructor_ = @Inject)`, it gets discovered automatically isn't it?

jlorper (commented):
Few suggestions, I want to do a full test on local too

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
eschultink: yeah - tbd; we ideally want the instance of `CommonRequestHandler` to be re-used rather than re-instantiated for each request - and `Sanitizer` as well. We need to performance test

java/impl/aws/pom.xml
```diff
@@ -0,0 +1,196 @@
...
+            </snapshots>
+        </repository>
+    </repositories>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>bom</artifactId>
+                <version>2.17.46</version>
```
aperez-worklytics, line 48: Can we have this as a variable? (same for L90, L95)

```suggestion
                <version>${aws.sdk}</version>
```

I know it is being used only here, but clear in this way IMHO

aperez-worklytics (commented):
Fine for me, let's wait untl @jlorper tests! Good job!

infra/examples/aws-google-workspace/main.tf
```diff
@@ -0,0 +1,180 @@
...
+    "gcal": {
+      display_name: "Google Calendar"
+      apis_consumed: [
+        "calendar-json.googleapis.com"
+      ]
+      oauth_scopes_needed: [
+        "https://www.googleapis.com/auth/calendar.readonly"
+      ]
+    }
+#    "gmail": {
```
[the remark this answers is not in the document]
eschultink: yeah, well - it's to provide an *example* of all the google-workspace stuff you can configure; but for example purposes, it's too much.

infra/modules/aws/variables.tf
eschultink: 👍  adding validation

java/core/src/main/java/co/worklytics/psoxy/PsoxyModule.java
eschultink: each lambda instance should map to exactly one source

java/core/src/main/java/co/worklytics/psoxy/gateway/ProxyRequestAdapter.java
eschultink: is that simpler than a single method, and always forcing caller to explicitly handle possibility of multiple values? (as opposed to lazily calling just `getHeader()` and presuming that they're correct that multi-value case can be safely ignored??)

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/LambdaRequest.java
eschultink: 👍  yes, requires adding explicit dep on apache `http-client`, but it's a transitive dep of Google http-client, so OK

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
eschultink: yeah, there's a task for you

java/impl/cmd-line/src/main/java/co/worklytics/psoxy/FileHandler.java
eschultink: it seems in dagger 2 you need both?? I get weird warnings

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -0,0 +1,209 @@
...
+            log.warning("Attempt to call endpoint blocked by rules: " + targetUrl + "; rules " + objectMapper.writeValueAsString(rules.getAllowedEndpointRegexes()));
+            return builder.build();
+        }
+
+        com.google.api.client.http.HttpRequest sourceApiRequest;
+        try {
+            sourceApiRequest =
+                requestFactory.buildGetRequest(new GenericUrl(targetUrl.toString()));
+        } catch (IOException e) {
+            builder.statusCode(500);
```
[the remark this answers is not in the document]
eschultink: we don't really know what it is ... I think safer to log rather than risking exposing sensitive info from proxy instance back to caller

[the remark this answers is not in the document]
aperez-worklytics: Yeah, then maybe we could change the message:

```suggestion
            builder.statusCode(500);
            builder.body("Failed to execute the request; review logs");
```

java/impl/aws/pom.xml
```diff
@@ -0,0 +1,202 @@
...
+    <version>${revision}</version>
+    <packaging>jar</packaging>
+
+    <properties>
+        <gcpProjectId>psoxy-dev-erik</gcpProjectId>
+        <sourceApi>gmail</sourceApi>
+        <!-- serviceAccount that function will run as. For Google Workspace use cases, should be the
+         one configured as OAuth Client and authorized for your instance. -->
+        <serviceAccount>psoxy-gmail-dwd@psoxy-dev-erik.iam.gserviceaccount.com</serviceAccount>
+    </properties>
```
jlorper, line 24: This shouldn't be here I guess

java/impl/gcp/src/main/java/co/worklytics/psoxy/Route.java
```diff
@@ -1,284 +1,40 @@
...
-        sourceApiRequest
-            .setThrowExceptionOnExecuteError(false)
-            .setConnectTimeout(SOURCE_API_REQUEST_CONNECT_TIMEOUT)
-            .setReadTimeout(SOURCE_API_REQUEST_READ_TIMEOUT);
+        //TODO: avoid rebuild graph everytime
+        DaggerGcpContainer.create().injectRoute(this);
 
-        //q: add exception handlers for IOExceptions / HTTP error responses, so those retries
-        // happen in proxy rather than on Worklytics-side?
+        HttpEventResponse abstractResponse =
```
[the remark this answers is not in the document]
eschultink: there's already a `response` in scope at that point - the `HttpResponse` from the Cloud Function SDK; the logic here is producing the abstract (or generic?) response object, then mapping it to the specific implementation version needed (the `HttpResponse` for the Cloud Function SDK)

eschultink:
@jlorper @aperez-worklytics  - ready to merge to `main`, if someone can give last review

infra/examples/aws-google-workspace/main.tf
```diff
@@ -0,0 +1,186 @@
...
+  providers = {
+    aws = aws
+  }
+}
+
+module "psoxy-package" {
+  source = "../../modules/psoxy-package"
+
+  implementation     = "aws"
+  path_to_psoxy_java = "../../../java"
```
jlorper, line 51: these relative paths usually give trouble. We might think on something, like passing a base path for the installation  and do 
```suggestion
  path_to_psoxy_java = "${var.base_dir}/java"
```

infra/modules/psoxy-package/build.sh
```diff
@@ -0,0 +1,13 @@
...
+# NOTE:
+cd $1
+cd core
+mvn package install > psoxy-package.log 2>&1
+
+cd ../impl/$2
+mvn package >> psoxy-package.log 2>&1
+
+# output back to Terraform
+OUTPUT_JSON="{\"path_to_deployment_jar\": \"$3\"}"
```
jlorper, line 12: could export a `TF_VAR_path_to_deployment_jar` easier to pass

jlorper (approved):
On suggestion.
Also, I've seen you have dropped the `deploy.sh` lambda. I think this is easy for customers to automate that way. There are no instructions on deploy right now I think.

PR #45, opened 2022-01-19
S114 aws call fix

jlorper:
Note: branches from #44 

Some improvements around AWS proxy
- End to end test call
- Use Gateway payload version 2 to handle responses properly. Simplify `Handler` code
- Log errors
- Shell scripts to ease manual tasks of building, deploy and test call.

image

### Change implications


infra/modules/aws-psoxy-instance/main.tf
```diff
@@ -19,6 +19,8 @@ resource "aws_apigatewayv2_integration" "map" {
   integration_uri           = aws_lambda_function.psoxy-instance.invoke_arn
   request_parameters        = {}
   request_templates         = {}
+  payload_format_version    = "2.0"
```
jlorper, line 22: This is the main change. Gateway has two payload formats, by default we were using 1.0 and I was having a lot of trouble with that version. Switched to 2.0 and could use directly the classes in their library.

infra/modules/aws/main.tf
```diff
@@ -25,7 +25,7 @@ resource "aws_apigatewayv2_stage" "live" {
   auto_deploy   = true
   access_log_settings {
     destination_arn = aws_cloudwatch_log_group.gateway-log.arn
-    format          = "$context.identity.sourceIp $context.identity.caller $context.identity.user [$context.requestTime] \"$context.httpMethod $context.resourcePath $context.protocol\" $context.status $context.responseLength $context.requestId $context.extendedRequestId"
+    format          = "$context.identity.sourceIp $context.identity.caller $context.identity.user [$context.requestTime] \"$context.httpMethod $context.resourcePath $context.protocol\" $context.status $context.responseLength $context.requestId $context.extendedRequestId $context.integrationErrorMessage"
```
jlorper, line 28: adds error message, helped clarify what was going on with the gateway <-> lambda mapping

java/impl/aws/src/main/java/co/worklytics/psoxy/Handler.java
```diff
@@ -1,33 +1,29 @@
...
-    @Inject
-    ObjectMapper objectMapper;
-
-    //TODO: improve this
-    //   - change this to directly take LambdaRequest as it's parameter type
-    //   - or to take InputStream, and then parse that with jackson (AWS lambda docs implies this is supported, but doesn't work)
     @SneakyThrows
     @Override
-    public String handleRequest(Map<String, Object> request, Context context) {
+    public APIGatewayV2HTTPResponse handleRequest(APIGatewayV2HTTPEvent httpEvent, Context context) {
```
jlorper, line 26: Directly use their http events, just created `APIGatewayV2HTTPEventRequestAdapter` to the interface the `CommonHandler` understands and back.

infra/modules/aws-psoxy-instance/main.tf
eschultink: ah, OK ... yeah, I was confused as to why those classes didn't match examples in AWS console / their documentation.

java/impl/impl.iml
```diff
@@ -1,9 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module org.jetbrains.idea.maven.project.MavenProjectsManager.isMavenModule="true" type="JAVA_MODULE" version="4">
-  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_11" inherit-compiler-output="true">
-    <exclude-output />
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_11">
+    <output url="file://$MODULE_DIR$/target/classes" />
+    <output-test url="file://$MODULE_DIR$/target/test-classes" />
     <content url="file://$MODULE_DIR$">
       <sourceFolder url="file://$MODULE_DIR$/target/generated-sources/annotations" isTestSource="false" generated="true" />
+      <excludeFolder url="file://$MODULE_DIR$/target" />
```
eschultink, line 8: 🤔  how does this work?? exclude `/target` but line above is including a subdirectory of it as a generated source folder??

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV2HTTPEventRequestAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Optional;
+
+/**
+ * Adapter for the APIGatewayV2HTTPEvent to the interface the {@link co.worklytics.psoxy.gateway.impl.CommonRequestHandler}
+ * understands
+ */
+@RequiredArgsConstructor
+public class APIGatewayV2HTTPEventRequestAdapter implements HttpEventRequest {
+
+    @NonNull APIGatewayV2HTTPEvent event;
```
eschultink, line 20: also `final`?

java/core/src/main/java/co/worklytics/psoxy/gateway/HttpEventResponse.java
```diff
@@ -14,12 +16,13 @@
...
 @Value
 @Builder
+@ToString
 public class HttpEventResponse {
 
     int statusCode;
 
     @Singular
-    List<Pair<String, String>> headers;
+    Map<String,String> headers;
```
eschultink, line 25: 👍  I didn't know that `lombok` `@Singular` was smart enough to work for a `Map<>`

java/core/src/main/java/co/worklytics/psoxy/gateway/HttpEventRequest.java
```diff
@@ -14,6 +14,8 @@
 
     Optional<String> getQuery();
 
-    Optional<List<String>> getHeader(String headerName);
+    Optional<String> getHeader(String headerName);
+
+    Optional<List<String>> getMultiValueHeader(String headerName);
```
eschultink, line 19: I don't think it's logical to provide *two* access points to the same thing.  a single-value header is just a special case of a multi-valued one. Callers should always us the `multi-value` case, to ensure they're handling it properly - rather than calling `getHeader()` and having ambiguous behavior in multi-valued case

so imho, `Optional<List<String>> getHeader(String headerName)` alone

eschultink (approved):
some suggestions, no blockers

java/impl/impl.iml
jlorper: In theory the only sources we want are the ones under `generated-sources`. Not sure about he order, but works for me.

java/core/src/main/java/co/worklytics/psoxy/gateway/HttpEventRequest.java
jlorper: :smile: not gonna agree. 99.9% of cases single values headers, adding the burden of picking first to the client doesn't make any sense. Call it syntactic sugar if you want, but simplifies code and the caller must know if it's dealing with a multi-valued header or not explicitly.

PR #47, opened 2022-01-25
S114 msft connectors

eschultink:
### Features
 - [example configs of MSFT connectors](https://app.asana.com/0/1201703347247995/1201639210470097)

### Change implications

 - dependencies added/changed? **yes** - azuread terraform provider


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -0,0 +1,144 @@
...
+                ((RequiresConfiguration) refreshHandler).getRequiredConfigProperties().stream());
+        }
+        return propertyStream.collect(Collectors.toSet());
+    }
+
+
+    @Override
+    public Credentials getCredentials(Optional<String> userToImpersonate) {
+        return OAuth2CredentialsWithRefresh.newBuilder()
+            //TODO: pull an AccessToken from some cached location or something? otherwise will
```
aperez-worklytics, line 70: :+1:  Yeah, a cache with the token expiration as expiration; it could be invalidated if a new access token is requested as well

infra/modules/azuread-connection/main.tf
```diff
@@ -0,0 +1,39 @@
...
+
+terraform {
+  required_providers {
+    azuread = {
+      version = "~> 2.15.0"
+    }
+  }
+}
+
+data "azuread_application_published_app_ids" "well_known" {}
```
aperez-worklytics, line 13: Main problem of this, as Terraform docs says as that it comes from an *unofficial* list. 

The id of `MicrosoftGraph` application has been always the same as far I remember (`00000003-0000-0000-c000-000000000000`). Of course it could change without warning in the future, but note that same app is not only used for any third party you could imagine but also for a lot of MSFT services.

There is a [link with official Ids](https://docs.microsoft.com/en-us/troubleshoot/azure/active-directory/verify-first-party-apps-sign-in#application-ids-for-commonly-used-microsoft-applications) that may be more accurate to refer on this case.

For this, I'd consider to have `application_id` in L16 as constant with a comment on the official documentation.

aperez-worklytics (approved):
Nice! Just some comments about Azure usage and documentation

README.md
```diff
@@ -51,9 +51,12 @@ And, depending on your scenario, you may also need:
     your psoxy instances in GCP *OR* if you plan to connect Google Workspace as a data source. It
     should be configured for the GCP project that will host your psoxy instance(s) and/or your
     connectors.
+  - [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli) Required to connect to
+    Microsoft 365 sources.
```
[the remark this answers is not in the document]
eschultink: yeah, i just think customers reading this view those as Microsoft 365 sources (although need enterprise app in azure to connect them)

infra/examples/aws-msft-365/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+
+## Example Configuration
+
+Example `terraform.tfvars`:
+```terraform
+aws_account_id                = "123456789"
+aws_assume_role_arn           = "arn:aws:iam::123456789:role/InfraAdmin"
+environment_name              = "dev-aws"
+connector_display_name_suffix = " Psoxy Dev AWS - erik"
```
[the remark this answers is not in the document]
eschultink: intent is for it to be a suffix; including the space is correct; adding code to magically add space if `connector_display_name_suffix` is non-null/non-empty would be tricky, and not necessarily more obvious

infra/modules/azuread-connection/main.tf
eschultink: yeah .... not sure what to do here. That the terraform code is doing is, in effect, the same as what MSFT tells you to do to look up permission IDs (https://docs.microsoft.com/en-us/graph/permissions-reference#retrieving-permission-ids) - except that they're querying by `appDisplayName =='Microsoft Graph'` . imho, that seems like a worse practice, as that's nothing but a marketing name

aperez-worklytics: Just FYI yeap, that's not a good practice. Display name can be repeated and the safest way is to use the proper GUID because is the only unique value (id) that identifies an AAD application. 

I've just created a couple AAD applications with an SP named `Microsoft Graph` in my personal AAD:

image

If we know the GUID best approach is to use it. ID will not change unless the application is destroyed, something that in this case is hard to happen. But if we don't know the ID and we need to perform a lookup as the MSFT docs says, we need more else to check -as they say, checking things in the properties as permissions, paths, whatever- and then order them by their expiration date of their keys or something, because we *could* have more than one entry and then get the wrong ID to use.

PR #51, opened 2022-01-28
S115 aws psoxy improvements

jlorper:
Changes to support end to end run of a connector (tested with GDirectory)
- Implement HEAD calls, as used in validation calls
- Fix linking between Gateway and Lambda functions
- Sanitization rules "caching" as singleton
- URL handling improvements 

### Features
 - [AWS end-to-end connector install](https://app.asana.com/0/1201703347247995/1201732924537315)

### Change implications


infra/modules/aws/main.tf
```diff
@@ -70,7 +71,12 @@ resource "aws_iam_role" "api-caller" {
           "Effect": "Allow",
           "Action": "execute-api:Invoke",
           "Resource": "arn:aws:execute-api:*:${var.aws_account_id}:*/*/GET/*",
-        }
+        },
+        {
+          "Effect": "Allow",
+          "Action": "execute-api:Invoke",
+          "Resource": "arn:aws:execute-api:*:${var.aws_account_id}:*/*/HEAD/*",
```
eschultink, line 78: I don't understand the point in allowing `HEAD` ... seems to complicate things, without adding any practical value

infra/modules/aws-psoxy-instance/main.tf
```diff
@@ -23,25 +44,34 @@ resource "aws_apigatewayv2_integration" "map" {
...
-  function_name = var.function_name
+  function_name = aws_lambda_function.psoxy-instance.function_name
   principal     = "apigateway.amazonaws.com"
 
-  # The /*/*/* part allows invocation from any stage, method and resource path
+
+  # The /*/*/ part allows invocation from any stage, method and resource path
   # within API Gateway REST API.
-  source_arn = "${var.api_gateway.arn}/*/*/*"
+  source_arn = "${var.api_gateway.execution_arn}/*/*/${var.function_name}/{proxy+}"
```
jlorper, line 71: This was what wasn't connecting the lambda with the gateway propertly

infra/modules/aws/main.tf
jlorper: Well, HEAD is slightly faster than GET, but other than that analogous and harmless from the proxy POV.
I could have changed the HEAD calls in Worklytics, but we should adapt proxy to Worklytics, not the other way around.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -37,13 +42,26 @@
     @Inject SanitizerFactory sanitizerFactory;
     @Inject Rules rules;
 
-    private Sanitizer sanitizer;
+    private volatile Sanitizer sanitizer;
+    private final Object $writeLock = new Object[0];
+
+
+    private Sanitizer loadSanitizerRules() {
```
jlorper, line 49: load rules as singleton

eschultink (approved):

PR #53, opened 2022-01-29
S115 msft auth

eschultink:
### Features
  - support OAuth 2.0 access_token via `client_credentials` assertion grant flow, to support MSFT sources

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/CanonicalOAuthAccessTokenResponseDto.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/CanonicalOAuthAccessTokenResponseDto.java
@@ -18,7 +18,7 @@ import java.util.TreeMap;
  * "access_token": "BWjcyMzY3ZDhiNmJkNTY",
  * "refresh_token": "Srq2NjM5NzA2OWJjuE7c",
  * "token_type": "bearer",
- * "expires": 3600
+ * "expires_in": 3600
  * }
  */
 @JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
@@ -30,7 +30,8 @@ public class CanonicalOAuthAccessTokenResponseDto {
     String accessToken;
     String refreshToken;
     String tokenType;
-    Integer expires;
+    //OAuth 2.0 https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.3
+    Integer expiresIn;
 
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     @Getter(onMethod_ = {@JsonAnyGetter})
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestPayloadBuilder.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestPayloadBuilder.java
@@ -12,6 +12,7 @@ import com.google.common.annotations.VisibleForTesting;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.SneakyThrows;
+import org.apache.commons.codec.binary.Hex;
 
 import javax.inject.Inject;
 import javax.inject.Provider;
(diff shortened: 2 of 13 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestPayloadBuilder.java
```diff
@@ -69,48 +80,81 @@
...
+    @SneakyThrows
+    String buildJwtAssertion(String clientId, String audience) {
 
+        //https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-certificate-credentials
         JsonWebSignature.Header header = new JsonWebSignature.Header();
+        header.setType("JWT");
         header.setAlgorithm("RS256");
-        header.setX509Thumbprint(config.getConfigPropertyOrError(ConfigProperty.PRIVATE_KEY_ID));
-        //header.setType("JWT"); //this IS a JWT, but MSFT examples don't include this
+        header.setX509Thumbprint(encodeKeyId(config.getConfigPropertyOrError(ConfigProperty.PRIVATE_KEY_ID)));
```
eschultink, line 117: our main codebase generated certs in powershell, and just took thumbprint value from there ... in proxy, we're using the HEX value as the `PRIVATE_KEY_ID` (which is also the "thumbprint" value that Azure shows in the console for the certificate) - and then per MSFT docs, we have to base64-urlencode that to put it in this JWT header (unclear to me why, given the this value is set on the JWT header, which will be serialized to JSON and then base64-encoded as the first part of the JWT before it's included in the URL ...)

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/CanonicalOAuthAccessTokenResponseDto.java
```diff
@@ -30,7 +30,8 @@
     String accessToken;
     String refreshToken;
     String tokenType;
-    Integer expires;
+    //OAuth 2.0 https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.3
```
jlorper, line 33: Maybe add a `@JsonAlias("expires")` ? What was the use case of "expires" ?

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestPayloadBuilder.java
```diff
@@ -69,48 +80,81 @@
...
+    //Utility to parse HEX string --> byte[]
+    //TODO: replace with java.util.HexFormat.of().parseHex(s) ... but that's only from Java 17
+    private byte[] hexStringToByteArray(String s) {
+        int len = s.length();
+        byte[] data = new byte[len / 2];
+        for (int i = 0; i < len; i += 2) {
+            data[i / 2] = (byte) ((Character.digit(s.charAt(i), 16) << 4)
+                + Character.digit(s.charAt(i+1), 16));
+        }
+        return data;
```
jlorper, line 154: I think we have apache commons-codec transitively (not sure why). It has `Hex.decodeHex` that does this.
Also it need to throw / handle when string length is not even.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -134,8 +132,11 @@ public AccessToken refreshAccessToken() throws IOException {
         }
 
         AccessToken asAccessToken(CanonicalOAuthAccessTokenResponseDto tokenResponse) {
+            //expires_in is RECOMMENDED, not REQUIRED in response; if omitted, we're supposed to
+            // assume a default value for service OR retrieve via some other means
+            Integer expiresIn = Optional.ofNullable(tokenResponse.getExpiresIn()).orElse(3600);
```
jlorper, line 137: constant DEFAULT_TOKEN_EXPIRATION_SECONDS or similar (better as `Duration` object)

jlorper (approved):
A couple of suggestions, but looks good to me

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/CanonicalOAuthAccessTokenResponseDto.java
eschultink: i don't know. it's not in the spec. did you see a data source that returns it? I think maybe some return a timestamp for `expires`, instead of `expires_in`, which is TTL in seconds.

PR #54, opened 2022-02-04
S115 psoxy improvements

jlorper:
Related [AWS proxy end-to-end](https://app.asana.com/0/1201703347247995/1201732924537315)

### Fixes
- Do not break startup if function does not uses OAuth refresh
- Double encoding parameters

### Features
- IDE cleanup
- Move HealthCheck to own class, better separate responsibilities
- Use constants

### Change implications

### DIFF
--- a/infra/dev-personal/main.tf
+++ b/infra/dev-personal/main.tf
@@ -33,101 +33,126 @@ module "psoxy-gcp" {
   ]
 }
 
-module "gmail-connector" {
-  source = "../modules/google-workspace-dwd-connection"
-
-  project_id                   = var.project_id
-  connector_service_account_id = "psoxy-gmail-dwd"
-  display_name                 = "Psoxy Connector - GMail${var.connector_display_name_suffix}"
-  apis_consumed                = [
-    "gmail.googleapis.com"
-  ]
-  oauth_scopes_needed          = [
-    "https://www.googleapis.com/auth/gmail.metadata"
-  ]
-
-  depends_on = [
-    module.psoxy-gcp
-  ]
-}
-
-# setup gmail to auth using a secret (not specific to Cloud Function)
-module "gmail-connector-auth" {
-  source = "../modules/gcp-sa-auth-key-secret-manager"
-
-  secret_project     = var.project_id
-  service_account_id = module.gmail-connector.service_account_id
-
-  # TODO: recommend migrate this to `PSOXY_{{function_name}}_SERVICE_ACCOUNT_KEY`
-  secret_id          = "PSOXY_SERVICE_ACCOUNT_KEY_gmail"
-}
-
-# for local dev, write the key to flat file on your machine (shouldn't do this for prod configs)
-resource "local_file" "gmail-connector-sa-key" {
-  filename = "gmail-connector-sa-key.json"
-  content_base64 = module.gmail-connector-auth.key_value
-}
-
-module "psoxy-gmail" {
-  source = "../modules/gcp-psoxy-cloud-function"
-
-  project_id            = var.project_id
-  function_name         = "psoxy-gmail"
-  source_kind           = "gmail"
-  service_account_email = module.gmail-connector.service_account_email
-
-  secret_bindings = {
-    PSOXY_SALT = {
-      secret_name    = module.psoxy-gcp.salt_secret_name
-      version_number = module.psoxy-gcp.salt_secret_version_number
-    },
-    SERVICE_ACCOUNT_KEY = {
-      secret_name    = module.gmail-connector-auth.key_secret_name
-      version_number = module.gmail-connector-auth.key_secret_version_number
+locals {
+  # Google Workspace Sources; add/remove as you wish
+  google_workspace_sources = {
+    # GDirectory connections are a PRE-REQ for gmail, gdrive, and gcal connections. remove only
+    # if you plan to directly connect Directory to worklytics (without proxy). such a scenario is
+    # used for customers who care primarily about pseudonymizing PII of external subjects with whom
+    # they collaborate in GMail/GCal/Gdrive. the Directory does not contain PII of subjects external
+    # to the Google Workspace, so may be directly connected in such scenarios.
+    "gdirectory": {
+      deploy: 1
+      display_name: "Google Directory"
+      apis_consumed: [
+        "admin.googleapis.com"
+      ]
+      oauth_scopes_needed: [
+        "https://www.googleapis.com/auth/admin.directory.user.readonly",
+        "https://www.googleapis.com/auth/admin.directory.user.alias.readonly",
+        "https://www.googleapis.com/auth/admin.directory.domain.readonly",
+        "https://www.googleapis.com/auth/admin.directory.group.readonly",
+        "https://www.googleapis.com/auth/admin.directory.group.member.readonly",
+        "https://www.googleapis.com/auth/admin.directory.orgunit.readonly",
+        "https://www.googleapis.com/auth/admin.directory.rolemanagement.readonly"
+      ],
+      worklytics_connector_name: "Google Workspace Directory via Psoxy"
+    }
+    "gcal": {
+      deploy: 0
+      display_name: "Google Calendar"
+      apis_consumed: [
+        "calendar-json.googleapis.com"
+      ]
+      oauth_scopes_needed: [
+        "https://www.googleapis.com/auth/calendar.readonly"
+      ]
+    }
+    "gmail": {
+      deploy: 1
+      display_name: "GMail"
+      apis_consumed: [
+        "gmail.googleapis.com"
+      ]
+      oauth_scopes_needed: [
+        "https://www.googleapis.com/auth/gmail.metadata"
+      ]
+    }
+    "google-chat": {
+      deploy: 0
+      display_name: "Google Chat"
+      apis_consumed: [
+        "admin.googleapis.com"
+      ]
+      oauth_scopes_needed: [
+        "https://www.googleapis.com/auth/admin.reports.audit.readonly"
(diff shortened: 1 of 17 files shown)


java/core/src/main/java/co/worklytics/psoxy/PsoxyModule.java
```diff
@@ -73,13 +75,28 @@ static SourceAuthStrategy sourceAuthStrategy(ConfigService configService, Set<So
...
-            configService.getConfigPropertyOrError(OAuthRefreshTokenSourceAuthStrategy.ConfigProperty.GRANT_TYPE);
-        return payloadBuilders
-            .stream()
-            .filter(impl -> Objects.equals(identifier, impl.getGrantType()))
-            .findFirst()
-            .orElseThrow(() -> new Error("No TokenRequestPayloadBuilder impl supporting oauth grant type: " + identifier));
+        // Final instantiation of configs are per-function. Grant type is dependent of type of auth
+        // strategy so might not exist for certain functions.
+        // If it is mis-configured will throw NPE at some point
+        Optional<String> grantTypeOptional = configService.getConfigPropertyAsOptional(OAuthRefreshTokenSourceAuthStrategy.ConfigProperty.GRANT_TYPE);
```
jlorper, line 81: change `getConfigPropertyOrError`, if function does not use GRANT_TYPE shouldn't err.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -244,8 +210,9 @@ public URL buildTarget(HttpEventRequest request) {
         URIBuilder uriBuilder = new URIBuilder();
         uriBuilder.setScheme("https");
         uriBuilder.setHost(config.getConfigPropertyOrError(ProxyConfigProperty.TARGET_HOST));
-        uriBuilder.setPath(request.getPath());
-        uriBuilder.setCustomQuery(request.getQuery().orElse(""));
+        // URL comes encoded, decode it prior to perform call to API origin to avoid double encoding issues
+        uriBuilder.setPath(URLDecoder.decode(request.getPath(), StandardCharsets.UTF_8));
+        uriBuilder.setCustomQuery(URLDecoder.decode(request.getQuery().orElse(""), StandardCharsets.UTF_8));
```
jlorper, line 215: Double encoding fix

aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -11,22 +11,23 @@
 import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.auth.Credentials;
 import com.google.auth.http.HttpCredentialsAdapter;
-import com.google.common.net.MediaType;
 import lombok.NoArgsConstructor;
 import lombok.SneakyThrows;
 import lombok.extern.java.Log;
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.Pair;
+import org.apache.http.HttpHeaders;
```
eschultink, line 18: k, i don't think this is a direct dependency of ours, right? so possibly need to add it to pom??

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
eschultink: actually, is this the right place??  I think it should be done within `HttpEVentRequest::getPath()`. Seems like parsing request --> nice POJO is responsibility of that class rather than the `CommonRequestHandler`

My particular concern is that GCP, AWS implementations each transform (or decorate) their native request format --> our generic `HttpEventRequest`.  Do they BOTH get double-encoded path values, such that this decoding is needed in both cases? or should only AWS's impl of `HttpEventRequest` have it?

eschultink (changes requested):
I think where the decoding happens is the wrong place

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
jlorper: I do not agree, this is transversal to all. "=" in this particular case, like any other reserved character in the URL must travel encoded. Is not something that the AWS Api Gateway does, but standard behavior per [RFC](https://www.ietf.org/rfc/rfc1738.txt).
It is handler responsibility to know what to do with the URL, not each implementation to choose if decode or not.

jlorper: To be clear, up to this point the URL is perfectly valir. Problem is that we do another request with it, so it double encodes it.
We need to set it back to their "true self" before encoding, so the underlying transport takes care of the final encoding.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
jlorper: good point, might be transitive. I'll add it.

infra/dev-personal/main.tf
```diff
@@ -33,119 +33,149 @@ module "psoxy-gcp" {
...
+locals {
+  # Google Workspace Sources; add/remove as you wish
+  google_workspace_sources = {
+    # GDirectory connections are a PRE-REQ for gmail, gdrive, and gcal connections. remove only
+    # if you plan to directly connect Directory to worklytics (without proxy). such a scenario is
+    # used for customers who care primarily about pseudonymizing PII of external subjects with whom
+    # they collaborate in GMail/GCal/Gdrive. the Directory does not contain PII of subjects external
+    # to the Google Workspace, so may be directly connected in such scenarios.
+    "gdirectory": {
+      deploy: 1
```
jlorper, line 45: this is just the test example. I put it up to date with your recent changes. I added this `deploy: 1` just to test.
Instead of removing the ones you don't want, just mark it as deployable or not.

java/impl/aws/pom.xml
```diff
@@ -87,12 +87,6 @@
             <artifactId>commons-lang3</artifactId>
             <version>${dependency.apache-commons-lang3.version}</version>
         </dependency>
-        <!-- explicit direct dep on apache HTTP Client (psoxy-core has transitive dep on it) -->
```
jlorper, line 90: I did remove this, is not used by us anywhere (is a transitive dep of `google-http-client`) but same way we don't specify its other transitive dependencies.
image

If we explicitly use it, we add it. That makes sense.

jlorper:
@eschultink please re-review. I don't agree with the change you suggest to the encoding place, but open to discussion.

java/impl/aws/pom.xml
eschultink: ok, i probably put it here thinking we did, or at some point we did. thanks for the cleanup!

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
eschultink: OK, I guess I'm confused.  For some reason, I thought the behavior differs by platform , which is why we didn't hit this bug with GCP but do now with AWS. Specifically, if request is

`your-psoxy-host/gmail/v1/messages?pageToken=123123` then the POJO passed in by GCP looks like:

```json
{
  "path": "gmail/v1/messages",
  "query": "pageToken=123123"
}
```

We should be just re-assumably that into `gmail/v1/messages?pageToken=123123" to 1) test with the REGEX as to whether it's allowed by our rules and then 2) forward on to source. 

Your change suggests `getQuery()` returns `pageToken%3D123412`, whereas I expect it to return `pageToken=123123`.

Am I misunderstanding something?

eschultink: see
image

jlorper: No, the problem comes when there are `=` in the parameter value. Page token is base64 so can have padding in the form `abc123=`

that when doing the call gets encoded to
`your-psoxy-host/gmail/v1/messages?pageToken=abc123%3D`

if we let the values as they come, the call that the proxy does to the endpoint will be re-encoded to
`your-psoxy-host/gmail/v1/messages?pageToken=abc123%253D`

so before that happens we need to set the token back to `abc123=`

eschultink: mmm.... AWS docs suggest they aren't double-encoding either:

https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

so is this something we're doing on the Worklytics side? If so, I don't think it's the right thing. We want psoxy to be drop-in replacement for the data source's REST API. if our request to psoxy looks like `gmail/v1/messages?pageToken%3D123123` that is wrong

jlorper: Again, this is not anything the Api Gateway does. Is the underlying URLFetchService encoding the url.
When we do call directly GDirectory the call goes like this `your-psoxy-host/gmail/v1/messages?pageToken=abc123%3D`
That's perfectly valid and the way to go, the `=` needs to be encoded.
The thing is that we are chaining calls, our lambda does ANOTHER call to the api, so that's where the double encoding may happen-

eschultink (approved):

PR #66, opened 2022-04-01
fix pseudonymization of 'null' properties

eschultink:
### Fixes
 - [azure ad proxy returning empty JSON array for users endpoint ](https://app.asana.com/0/1201991831825945/1202014631790084/f). - it's actually that the JsonPath implementation seems to blow up when applying a `MapFunction` to a JSON structure, if that function throws an exception. Resulting document after the map is `[]` though, instead of either throwing the error up OR just knocking out (or leaving unchanged) the input to the `MapFunction`.  This affected our test MSFT org, which has users without email addresses - so rules were pseudonymizing `null`, which was unexpected.

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
@@ -132,6 +132,11 @@ public class SanitizerImpl implements Sanitizer {
                     .delete(document, jsonConfiguration);
             }
 
+            //TODO: error handling within the map functions. any exceptions thrown within the map
+            //      function seem to be suppressed, and an empty [] left as the 'document'.
+            // ideas:
+            // jsonConfiguration.addEvaluationListeners(); -->
+
             for (JsonPath pseudonymization : pseudonymizationsToApply) {
                 document = pseudonymization
                     .map(document, this::pseudonymizeToJson, jsonConfiguration);
@@ -146,7 +151,6 @@ public class SanitizerImpl implements Sanitizer {
                 document = pseudonymization
                     .map(document, this::pseudonymizeWithOriginalToJson, jsonConfiguration);
             }
-
 
             return jsonConfiguration.jsonProvider().toJson(document);
         }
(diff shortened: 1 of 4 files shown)


java/core/src/test/resources/api-response-examples/microsoft-365/directory/users.json
```diff
@@ -67,6 +67,15 @@
       "preferredLanguage": "en-US",
       "surname": "Allen",
       "userPrincipalName": "Paul@worklytics.onmicrosoft.com"
+    },
+    {
+      "businessPhones": [],
+      "jobTitle": null,
+      "mail": null,
```
eschultink, line 74: this is the problematic case

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
```diff
@@ -132,6 +132,11 @@ public String sanitize(@NonNull URL url, @NonNull String jsonResponse) {
                     .delete(document, jsonConfiguration);
             }
 
+            //TODO: error handling within the map functions. any exceptions thrown within the map
+            //      function seem to be suppressed, and an empty [] left as the 'document'.
```
aperez-worklytics, line 136: Yep, it seems that  flag `Option.SUPPRESS_EXCEPTIONS` is suppressing everything, no matter where the exception is thrown.

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
```diff
@@ -132,6 +132,11 @@ public String sanitize(@NonNull URL url, @NonNull String jsonResponse) {
...
+            // ideas:
+            // jsonConfiguration.addEvaluationListeners(); -->
```
aperez-worklytics, line 138: Not sure, I mean it could work but at the moment I don't see an improvement in comparison with current solution and using evaluation listeners. At the end, `EvaluationListener` will be like this (some dirty code):

```
static class X implements EvaluationListener {

    @Override
    public EvaluationContinuation resultFound(FoundResult found) {
        if (found.result() == null) {
            return EvaluationContinuation.CONTINUE;
        }
        if (Objects.equals(found.path(), "my-path")) {
            return EvaluationContinuation.CONTINUE;
        }
        
        return EvaluationContinuation.ABORT;
    }
}
```

aperez-worklytics (approved):
Good catch!

PR #71, opened 2022-04-12
S120 bulk flow improvements

eschultink:
### Features
 - better bucket names
 - cleanup/naming of resources

### Change implications

---
  - https://app.asana.com/0/0/1202053086212667


aperez-worklytics (approved):
Both tried; I've added a couple of commits for updating the GCP bucket name and readme files

PR #75, opened 2022-04-21
S121 more example fixes

eschultink:
### Fixes
 - docs wrong
 - calendar scopes wrong in examples
 - `SOURCE` env variable wrong for stock msft config

### Change implications


infra/examples/README.md
```diff
@@ -59,3 +59,12 @@ backend "s3" {
   region = "us-east-1"
 }
 ```
+
+## Disclaimer
+All examples in this directory and its subdirectories are provided as-is, for reference only, with
+no warranty. You should review all resulting Terraform plans carefully before applying them to your
+environment. Additionally, we strongly recommend that you make example curl requests to any proxy
+instance you deploy to verify you are satisified with the behavior of the proxy with regard to the
+transofmration of your data.
```
jlorper, line 68:
```suggestion
transformation of your data.
```

jlorper (approved):
one typo

aperez-worklytics (approved):

PR #83, opened 2022-05-04
S122 more example fixes

eschultink:
### Features
 - make GCP billing account optional (and similar)
 - avoid examples failing bc of empty `""` secret; avoid changes causing problems

### Change implications

---
  - https://app.asana.com/0/0/1202214476274918
  - https://app.asana.com/0/0/1202214476274930


jlorper (changes requested):
Looks good, but typos and diff merge marks left that need to be cleaned

PR #90, opened 2022-05-06
S122 direct lambda call fixes

jlorper:
- Grant permissions to proxy caller to invoke lambdas
- Removes gateway configuration
- Removes concat call in resources names
- Test function for AWS sign calls with "lambda" service
- NPE in handler when switching from API Gateway
- Last commit is also tag "v0.3.0-beta.1", pointed everything to this, so passes CI

### Change implications

---
  - https://app.asana.com/0/0/1202219382641155

### DIFF
--- a/infra/examples-dev/aws-msft-365/main.tf
+++ b/infra/examples-dev/aws-msft-365/main.tf
@@ -177,11 +177,11 @@ module "psoxy-msft-connector" {
 
   function_name        = "psoxy-${each.key}"
   source_kind          = each.value.source_kind
-  api_gateway          = module.psoxy-aws.api_gateway
   path_to_function_zip = module.psoxy-package.path_to_deployment_jar
   function_zip_hash    = module.psoxy-package.deployment_package_hash
   path_to_config       = "../../../configs/${each.value.source_kind}.yaml"
   api_caller_role_arn  = module.psoxy-aws.api_caller_role_arn
+  api_caller_role_arn_name = module.psoxy-aws.api_caller_role_name
   aws_assume_role_arn  = var.aws_assume_role_arn
   example_api_calls    = each.value.example_calls
 
--- a/infra/examples/aws-google-workspace/main.tf
+++ b/infra/examples/aws-google-workspace/main.tf
@@ -171,13 +171,14 @@ module "psoxy-google-workspace-connector" {
 
   source = "git::https://github.com/worklytics/psoxy//infra/modules/aws-psoxy-instance?ref=v0.3.0-beta.1"
 
-  function_name        = "psoxy-${each.key}"
-  source_kind          = each.key
-  path_to_function_zip = module.psoxy-package.path_to_deployment_jar
-  function_zip_hash    = module.psoxy-package.deployment_package_hash
-  path_to_config       = "../../../configs/${each.key}.yaml"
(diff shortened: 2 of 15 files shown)


aperez-worklytics (approved):

infra/examples/aws-google-workspace/main.tf
```diff
@@ -49,7 +49,7 @@ module "psoxy-package" {
 
 # holds SAs + keys needed to connect to Google Workspace APIs
 resource "google_project" "psoxy-google-connectors" {
-  name            = "Worklytics Connectors${var.environment_name == null ? "" : concat(" - ", var.environment_name)}"
```
eschultink, line 52: thanks, my fault.

tools/test-psoxy.sh
```diff
@@ -45,7 +45,7 @@ then
   log "Calling proxy..."
   log "Request: $TEST_URL"
   log "Waiting Response:"
-  awscurl "${curlparams[@]}" --service execute-api --access_key $CALLER_ACCESS_KEY_ID --secret_key $CALLER_SECRET_ACCESS_KEY --security_token $CALLER_SESSION_TOKEN $TEST_URL
```
eschultink, line 48: 👍 ; ok, yeah, I see that the auth signature is coupled to the service, not just the URL.  That's very strange to me (why should caller be aware of the AWS service used the implement a REST API??), but it's AWS fault not ours.

eschultink: direct calls to the lambdas is the KISS solution that I think will work for 90% of customers, but customer call from Friday does get me worried that API gateways will be appealing to some enterprise customers precisely BC of the complexity and the ability to put a bunch of additional rules/restrictions on the mappings.

infra/examples-dev/aws-msft-365/main.tf
```diff
@@ -177,11 +177,11 @@ module "psoxy-msft-connector" {
 
   function_name        = "psoxy-${each.key}"
   source_kind          = each.value.source_kind
-  api_gateway          = module.psoxy-aws.api_gateway
   path_to_function_zip = module.psoxy-package.path_to_deployment_jar
   function_zip_hash    = module.psoxy-package.deployment_package_hash
   path_to_config       = "../../../configs/${each.value.source_kind}.yaml"
   api_caller_role_arn  = module.psoxy-aws.api_caller_role_arn
+  api_caller_role_arn_name = module.psoxy-aws.api_caller_role_name
```
eschultink, line 184: style (`terraform fmt -recursive`)

infra/examples/aws-google-workspace/main.tf
```diff
@@ -33,23 +33,23 @@ provider "aws" {
 }
 
 module "psoxy-aws" {
-  source = "git::https://github.com/worklytics/psoxy//infra/modules/aws?ref=v0.2.0-beta.1"
+  source = "git::https://github.com/worklytics/psoxy//infra/modules/aws?ref=v0.3.0-beta.1"
```
eschultink, line 36: +1

eschultink (approved):
👍  looks good. one idea is to add a `CHANGELOG` at top level to clarify that `0.2` is initial tagged version, and `0.3` does the direct lambda change.

PR #94, opened 2022-05-10
S122 lifecycle value ignore

jlorper:
- Lifecycle ignore for certain parameters
- Fail on error when building jar
- Update tag

### Change implications

### DIFF
--- a/infra/examples-dev/aws-msft-365/main.tf
+++ b/infra/examples-dev/aws-msft-365/main.tf
@@ -147,6 +147,12 @@ resource "aws_ssm_parameter" "client_id" {
   name  = "PSOXY_${upper(replace(each.key, "-", "_"))}_CLIENT_ID"
   type  = "String"
   value = module.msft-connection[each.key].connector.application_id
+
+  lifecycle {
+    ignore_changes = [
+      value
+    ]
+  }
 }
 
 resource "aws_ssm_parameter" "refresh_endpoint" {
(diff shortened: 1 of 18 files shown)


infra/examples-dev/aws-msft-365/main.tf
```diff
@@ -156,6 +162,12 @@ resource "aws_ssm_parameter" "refresh_endpoint" {
   type      = "String"
   overwrite = true
   value     = "https://login.microsoftonline.com/${var.msft_tenant_id}/oauth2/v2.0/token"
+
+  lifecycle {
+    ignore_changes = [
+      value
+    ]
+  }
```
eschultink, line 170: I don't think you want this here ...

infra/examples/aws-msft-365/main.tf
```diff
@@ -156,13 +162,19 @@ resource "aws_ssm_parameter" "refresh_endpoint" {
   type      = "String"
   overwrite = true
   value     = "https://login.microsoftonline.com/${var.msft_tenant_id}/oauth2/v2.0/token"
+
+  lifecycle {
+    ignore_changes = [
+      value
+    ]
+  }
```
eschultink, line 170: doesn't make sense here. this isn't something they change

aperez-worklytics (approved):

infra/modules/aws/main.tf
```diff
@@ -120,6 +120,12 @@ resource "aws_ssm_parameter" "salt" {
   type        = "SecureString"
   description = "Salt used to build pseudonyms"
   value       = sensitive(random_password.random.result)
+
+  lifecycle {
+    ignore_changes = [
+      value
+    ]
+  }
```
eschultink, line 128: OK, although a bit weird.  should certainly comment this. I think it's entirely legitimate for us to pre-generate a value, and then they overwrite it if they don't trust ours - and if so, our terraform doesn't replace it - but still should comment that that's what's happening.

eschultink (approved):

eschultink:
tech debt ticket to follow-up: https://app.asana.com/0/1202204100399673/1202256552566197

PR #97, opened 2022-05-17
S123 msft keys not updated

eschultink:
### Fixes
 - msft keys not updated
 - misleading log message

### Features
 - improve examples for paths

### Change implications


infra/modules/private-key-aws-parameter/variables.tf
```diff
@@ -12,3 +12,9 @@ variable "private_key" {
   type        = string
   description = "value of the key"
 }
+
+variable "user_managed_value" {
```
jlorper, line 16: I don't see where this is used. I assume is for discriminating use of `private-key-aws-parameter` vs `private-key-aws-parameter-user-managed` modules

jlorper (approved):

PR #104, opened 2022-05-24
s123 : fix aws role arn to be consistent with other examples

eschultink:
### Fixes
  - fix call ARN in google example

### Change implications


aperez-worklytics (approved):
Needs to be changed as well in L180: `aws_assume_role_arn = var.aws_assume_role_arn`

PR #106, opened 2022-06-08
S124 new rules

eschultink:
### Fixes
  - various urls containing `joinUrl`s with `pwd` params

### Change implications

 - dependencies added/changed? **o**


java/core/src/main/java/co/worklytics/psoxy/rules/RuleSet.java
```diff
@@ -0,0 +1,7 @@
+package co.worklytics.psoxy.rules;
+
+public interface RuleSet {
+
+    @Deprecated // migrate to use transform-level ids
+    String getDefaultScopeIdForSource();
```
eschultink, line 6: thought is to define `scope` on the transform itself, as we have cases that mix scopes other than email (specifically the directory integrations have phone numbers, screen names, employee ids, etc)

java/core/src/main/java/co/worklytics/psoxy/rules/RuleSet.java
```diff
@@ -0,0 +1,7 @@
+package co.worklytics.psoxy.rules;
+
+public interface RuleSet {
```
eschultink, line 3: introduce where want to bind interfaces to values of EITHER `Rules1` or `Rules2`

java/core/src/main/java/co/worklytics/psoxy/Sanitizer.java
```diff
@@ -29,9 +31,9 @@ class Options implements Serializable {
         String defaultScopeId;
 
         //q: add regexes to whitelist endpoints that we actually use??
-        Rules rules;
-
+        Rules1 rules;
 
+        Rules2 rules2;
```
eschultink, line 36: backwards compatibility; support for `Rules1` should be maintained, while adding more intuitive `Rules2` with support for more transforms

java/core/src/main/java/co/worklytics/psoxy/ConfigRulesModule.java
```diff
@@ -3,35 +3,30 @@
...
 
-        BiFunction<Optional<Rules>, String, Optional<Rules>> loadAndLog = (o, msg) -> {
+        BiFunction<Optional<RuleSet>, String, Optional<RuleSet>> loadAndLog = (o, msg) -> {
             if (o.isPresent()) {
                 log.info(msg);
             }
             return o;
         };
 
-        return loadAndLog.apply(rulesUtils.getRulesFromFileSystem(PATH_TO_RULES_FILES), "Rules: loaded from file system")
```
eschultink, line 33: dropped support for explicit `rules.yaml` file. in practice, I think that's no easier than using env variable OR AWS SystemParameter (and those are quite easy, as can change rules without re-deploy)

java/core/src/main/resources/rules/microsoft-365/outlook-mail.yaml
```diff
@@ -1,84 +1,86 @@
-allowedEndpointRegexes:
-  - "^/(v1.0|beta)/(groups|users)/?[^/]*"
-  - "^/(v1.0|beta)/(groups|users)\\?.*"
-  - "^/(v1.0|beta)/groups/[^/]*/members[^/]*"
-  - "^/(v1.0|beta)/users/[^/]*/messages/[^/]*"
-  - "^/(v1.0|beta)/users/[^/]*/mailFolders(/SentItems|\\('SentItems'\\))/messages.*"
-  - "^/(v1.0|beta)/users/[^/]*/mailboxSettings"
```
eschultink, line 7: main advantage is the inversion of the structure; define endpoints and then transform(s) to apply to each.  

by default, only explicitly allowed endpoints can be called

aperez-worklytics (approved):
I think is fine, new structure looks much better than previous one

java/core/src/test/java/co/worklytics/psoxy/rules/JavaRules2TestBaseCase.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * base test case for rules written in java (although also expected to have yaml-encoded equivalent
+ * in repo under `src/main/resources/rules/`)
+ */
+public abstract class JavaRules2TestBaseCase extends Rules2BaseTestCase {
+
+
+    //TODO: we could avoid this if java rules serialized to YAML as part of build process or
+    // something, rather than under version control themselves; or flip so that master copy is the
+    // yaml and java ones read from file system??
```
jlorper, line 19: As a future thing, to me the java ones allows use to do probably more complex transformations, negative selections, subsets, etc. so verbosity reduced and rules easier to understand enriching the jsonPaths and transforming then to yaml. So maybe a tool that generates them... just throwing the idea.

jlorper (approved):
OK, some improvements spotted, no blockers though

java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
```diff
@@ -0,0 +1,73 @@
...
+    public Rules2 withAdditionalEndpoints(Endpoint... endpointsToAdd) {
+        Rules2Builder builder = this.toBuilder();
+        Arrays.stream(endpointsToAdd).forEach(builder::endpoint);
+        return builder.build();
+    }
+
+    @JsonPropertyOrder(alphabetic = true)
+    @Builder
+    @AllArgsConstructor //for builder
+    @NoArgsConstructor //for Jackson
```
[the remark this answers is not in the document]
eschultink: cool, something good to look at but need to be very careful as could definitely change some behaviors if we do non-standard stuff in Jackson case (eg, if setters do anything that builders don't, such as populating id fields)

one thing I miss in lombok builder is some sort of a "hook", where you could do work as the last step of `build()`.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -85,7 +87,7 @@ public HttpEventResponse handle(HttpEventRequest request) {
         } else {
             builder.statusCode(HttpStatus.SC_FORBIDDEN);
             builder.header(ResponseHeader.ERROR.getHttpHeader(), ErrorCauses.BLOCKED_BY_RULES.name());
-            log.warning(String.format("%s. Blocked call by rules %s", callLog, objectMapper.writeValueAsString(rules.getAllowedEndpointRegexes())));
+            log.warning(String.format("%s. Blocked call by rules %s", callLog, objectMapper.writeValueAsString(rules)));
```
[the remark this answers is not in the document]
eschultink: we return the sha of the rules as a header, so we can at least see if they're running a known version or not. 

returning encoded rules themselves feels like it may open security concerns. Eg, someone opened some API endpoint for some reason, but didn't robustly test the behavior.  From SHA alone, you don't necessarily know this; but from SHA you do know which well-known rule set they're using.

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
```diff
@@ -65,17 +73,30 @@ List<JsonPath> applicablePaths(@NonNull List<Pair<Pattern, List<JsonPath>>> rule
...
+                compiledAllowedEndpoints = options.getRules2().getEndpoints().stream()
+                    .map(Rules2.Endpoint::getPathRegex)
                     .map(regex -> Pattern.compile(regex, CASE_INSENSITIVE))
                     .collect(Collectors.toList());
             }
             String relativeUrl = URLUtils.relativeURL(url);
-            return compiledAllowedEndpoints.stream().anyMatch(p -> p.matcher(relativeUrl).matches());
+
+            return options.getRules2().getAllowAllEndpoints() ||
+                compiledAllowedEndpoints.stream().anyMatch(p -> p.matcher(relativeUrl).matches());
```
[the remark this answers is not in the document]
eschultink: imho, general best practice is to keep constructors as light as possible.  This is "just in time" initialization, which I think of as best practice.

also, exceptions being thrown out of constructors seems very likely to cause problems and not be well-handled, as constructors often invoked by frameworks, such as dagger.  If something blows up in the object graph at run time bc of malformed rules, I think it will be hard to hunt down why

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
```diff
@@ -91,7 +112,93 @@ public String sanitize(@NonNull URL url, @NonNull String jsonResponse) {
...
+
+        } else if (transform instanceof Transform.PseudonymizeEmailHeader) {
+            for (JsonPath path : paths) {
+                document = path.map(document, this::pseudonymizeEmailHeaderToJson, jsonConfiguration);
+            }
+        } else if (transform instanceof Transform.RedactRegexMatches) {
+            MapFunction f = getRedactRegexMatches((Transform.RedactRegexMatches) transform);
+            for (JsonPath path : paths) {
+                document = path.map(document, f, jsonConfiguration);
+            }
```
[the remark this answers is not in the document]
eschultink: i considered that, but without (huge) refactor it would couple the transforms to our implementation (in particular, the `JsonPath` library. 

eg, redact is
```java
            for (JsonPath path : paths) {
                document = path.delete(document, jsonConfiguration);
            }
```
so would need to compile the paths, as well as be aware of the `jsonConfiguration` thing being passed in as argument ...

java/core/src/main/java/co/worklytics/psoxy/rules/Validator.java
```diff
@@ -11,25 +10,48 @@
...
+            rules1.getAllowedEndpointRegexes().forEach(Pattern::compile);
         }
     }
 
-    static void validate(@Nullable List<Rules.Rule> rules) {
+    static public void validate(@NonNull Rules2 rules) {
+        rules.getEndpoints().forEach(Validator::validate);
+    }
+    static void validate(@Nullable Rules2.Endpoint endpoint) {
+        if (endpoint != null) {
```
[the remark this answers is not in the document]
eschultink: 🤔  not sure why i would have done this.  possibly let GitHub copilot write it for me ...

java/core/src/test/java/co/worklytics/psoxy/rules/JavaRules2TestBaseCase.java
eschultink: yeah, i think there's a future in which we do something like that.  Have a web ux in Worklytics that generates rule sets based on policies (eg, "include employee ids in plain") to allow  high level customization without granularly customizing the rule sets.

PR #109, opened 2022-06-10
S124 path handling

eschultink:
### Fixes
  - possibly OOM if not disconnect
  - thread-safety of sanitizer

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
@@ -1,6 +1,7 @@
 package co.worklytics.psoxy.impl;
 
 import co.worklytics.psoxy.*;
+import co.worklytics.psoxy.rules.RuleSet;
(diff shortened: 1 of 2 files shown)


java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
```diff
@@ -388,4 +390,38 @@ public PseudonymizedIdentity pseudonymize(@NonNull Number value) {
         return pseudonymize((Object) value);
     }
 
+    List<Pair<Pattern, List<JsonPath>>>  getCompiledPseudonymizations() {
+        if (compiledPseudonymizations == null) {
+            synchronized ($writeLock){
```
jlorper, line 395: ideally you should re-check the null, in case it's been done while getting the lock
See double-checked locking pattern
https://www.baeldung.com/java-singleton-double-checked-locking
https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

jlorper (approved):
do the double checked locking and good to go

PR #110, opened 2022-06-15
s125 : more cases where zoom plugin might stick zoom URLs

eschultink:
### Fixes
  - display names seem to sometimes be URLs

### Change implications


java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
```diff
@@ -202,7 +202,9 @@ MapFunction getRedactRegexMatches(Transform.RedactRegexMatches transform) {
        List<Pattern> patterns = transform.getRedactions().stream().map(Pattern::compile).collect(Collectors.toList());
        return (s, jsonConfiguration) -> {
            if (!(s instanceof String)) {
-               log.warning("value matched by " + transform + " not of type String");
+               if (s != null) {
+                   log.warning("value matched by " + transform + " not of type String");
```
aperez-worklytics, line 206: Why not keep the warning if the string is null too?

aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java
eschultink: bc that's not really an error case. For example, content could be `{"onlineMeetingUrl": null}`, which is valid JSON, our rule would match the value of the `onlineMeetingUrl` - it's just `null` in this case, which is OK.  My concern is if rule is "redact portions of the value that match this regex", but the value isn't a `String`  - then we can't safely apply the rule.

PR #116, opened 2022-06-22
S124 response gzip compression

jlorper:
Enables compression between worklytics and AWS proxy instances.

### Features
[aws: compression bw worktlytics AND proxy](https://app.asana.com/0/1201039336231823/1201738052069994)

### Change implications

image


java/impl/aws/src/main/java/co/worklytics/psoxy/Handler.java
```diff
@@ -60,12 +89,67 @@ public APIGatewayV2HTTPResponse handleRequest(APIGatewayV2HTTPEvent httpEvent, C
...
+        boolean compressionOutweighOverhead(String body) {
+            return StringUtils.length(body) >= MIN_BYTES_TO_COMPRESS;
+        }
+
+        /**
+         * Compresses content as binary base64
+         * @param body
+         * @return optional with content if compression has been applied
+         */
+        Optional<String> compressBodyAndConvertToBase64(String body) {
```
eschultink, line 120:
```suggestion
        Optional<String> compressBodyAndEncodeToBase64(String body) {
```

also, shouldn't this be `@VisibleForTesting`? I think intent is that the other method is the one called in practice from outside.

eschultink (approved):

PR #128, opened 2022-08-22
s130 : drop v1 rule format

eschultink:
### Features
  - drop v1 rules format; all customers on v2 rules

### Change implications


java/core/src/main/java/co/worklytics/psoxy/rules/Validator.java
```diff
@@ -10,23 +11,16 @@
 public class Validator {
 
     static public void validate(@NonNull RuleSet rules) {
-        if (rules instanceof Rules1) {
-            validate((Rules1) rules);
+        if (rules instanceof CsvRules) {
+            validate((CsvRules) rules);
         } else if (rules instanceof Rules2) {
             validate((Rules2) rules );
         }
```
aperez-worklytics, line 18: Mmm, worth to add here one more case to ensure we don't support other thing? 

```suggestion
        } else {
          throw new NotImplementedException("Set not supported!");
        }
```

aperez-worklytics (approved):

PR #130, opened 2022-08-22
s130 : support for transiently reversible pseudonyms (eg, encrypted as well as a hashed values returned)

eschultink:
### Features
  - add option to rules to encrypt with transiently reversible pseudonyms

### Change implications

 - dependencies added/changed? **yes - common OSS libs updated; use of Avaulta lib for pseudonyms, adding support for transiently reversible ones**

if legacy psoxy instances are updated with this code, they should invoked with the header (`X-Psoxy-Pseudonym-Implementation: LEGACY`) to ensure that the legacy hashes (including `scope` value) are returned


eschultink:
ugh, not ready; tried to reconstitute some stuff i had been messing with, but it is not ready for prime time

PR #138, opened 2022-08-30
S130 id handling improvements

eschultink:
### Features
 - methods to aid parsing of pseudonyms in clients
 - match artifact version to repo release versions
 - examples with multiple instances for 'bulk' case

### Change implications

 - dependencies added/changed? **no**

PR #139, opened 2022-08-30
S130 plain yaml rules

eschultink:
### Features
 - rules as base64 *or* plain text yaml
 - support env vars passed into ALL proxy modules where it makes sense
 - use env vars to set PSEUDONYMIZE_APP_IDS on all of a customers instances

### Change implications


java/core/core.iml
```diff
@@ -13,7 +13,6 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
-    <orderEntry type="module" module-name="gateway-core" />
```
aperez-worklytics, line 16: Ah, that has been removed too in https://github.com/Worklytics/psoxy/pull/138/files#diff-311e2f8f3a14f67cf4b6bb78c8f7133e8a9ddf832fd907c0857f8d478f584733

java/core/src/main/java/co/worklytics/psoxy/rules/RulesUtils.java
```diff
@@ -32,23 +33,36 @@ public String sha(RuleSet rules) {
...
+                yamlEncodedRules = configuredRules.get();
+            }
+            return Optional.of(parse(yamlEncodedRules));
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    @VisibleForTesting
+    RuleSet parse(String yamlString) {
```
aperez-worklytics, line 52:
```suggestion
    RuleSet parse(@NotNullString yamlString) {
```

otherwise `reader` will complain with its internal null check

aperez-worklytics (approved):

PR #142, opened 2022-08-31
s130 : set rules via terraform

eschultink:
### Features
  - configure rules via Terraform

### Change implications

### DIFF
--- a/infra/examples-dev/aws-msft-365/main.tf
+++ b/infra/examples-dev/aws-msft-365/main.tf
@@ -41,10 +41,14 @@ locals {
   bulk_sources = {
     "hris" = {
       source_kind = "hris"
+      rules = {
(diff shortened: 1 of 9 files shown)


infra/examples-dev/aws-msft-365/main.tf
```diff
@@ -41,10 +41,14 @@ locals {
   bulk_sources = {
     "hris" = {
       source_kind = "hris"
+      rules = {
+        columnsToRedact = []
+        columnsToPseudonymize = [
+          "email",
```
aperez-worklytics, line 47: I'd clarify somehow if when adding here rules, we are overriding the rules created or if they are going to be concatenated to the existing ones

aperez-worklytics (approved):

PR #148, opened 2022-09-07
S130 node test suite

davidfq:
**TODO**
- [x] Documentation
- [x] ~Create `.env` file for "multiple psoxy calls" runner~

### Fixes
.

### Features
- Save psoxy call results to file
- Run multiple psoxy calls, i.e. `npm run test:gcal` -> will run all API endpoints sequentially

### Change implications

 - dependencies added/changed? **yes** npm dependencies


tools/psoxy-test/cli.js
```diff
@@ -31,6 +31,7 @@ const theme = {
     .option('-t, --token <token>', 'Authorization token for GCP')
     .option('-v, --verbose', 'Verbose output', false)
     .option('-z, --gzip', 'Add gzip compression header', false)
+    .option('--save-to-file', 'Save test results to file', false)
```
eschultink, line 34: so I would probably prefer if `-s` is alias to this, rather than `--skip`; saving output to a file is a very common use case for our customers, whereas `--skip` is really just for us (dev purpose)

tools/psoxy-test/data-sources/runner.js
```diff
@@ -0,0 +1,53 @@
...
+    console.log(`${chalk.blue(dataSource.name)}, fetching ${chalk.blue(endpoint.name)}: ${url}`);
+    result = await psoxyTest({
+      ...options,
+      url: url,
+    })
+
+    if (result?.error) {
+      console.error(chalk.bold.red(result.error));
+    } else if (result.status === 200) {
+      console.log(chalk.green('OK'));
```
eschultink, line 45: imho, I would just put the full HTTP status line in all cases that there is one (eg, assuming error isn't some other IO thing), colored RED for non-2xx response codes. Eg,

200 OK
500 Internal Server Error
...

tools/psoxy-test/README.md
```diff
@@ -1,13 +1,78 @@
...
+```
+Zoom example:
+```shell
+node cli.js -u https://us-central1-acme.cloudfunctions.net/v2/users -t <IDENTITY_TOKEN>
+```
+
+(*) You can obtain it by running `gcloud auth print-identity-token` (using Google Cloud SDK)
+
+## Testing all endpoints for a given data source
+We also provide some npm scripts that allow you to test all the endpoints for a given data source. These scripts rely on the same Node.js module that runs the CLI script described above, but instead of providing the options via CLI, you have to create a `.env` file with the following contents:
```
eschultink, line 40: I'm not sure I like this `.env` approach. I worry this will be trickier for customers than just passing these as arguments.

tools/psoxy-test/README.md
```diff
@@ -1,13 +1,78 @@
...
+PSOXY_URL=<url>
+PSOXY_ROLE=<role ARN>
+PSOXY_IMPERSONATE=<user email> # only for Google Workspace data sources
+PSOXY_SAVE_TO_FILE=true
+PSOXY_VERBOSE=true
+```
+```shell
+# For GCP deploys
+PSOXY_URL=<url>
+PSOXY_TOKEN=<identity token>
```
eschultink, line 52: I probably prefer just asking them to pipe it in  (showing them the right bash command to do it, eg 

```bash
node cli.js -token `gcloud auth print-identity-token` 
```

additionally, I think the tokens are only good for an hour; running `gcloud auth print-identity-token`, copying the contents, opening a flat file, pasting it in there, saving it - that's going to be really tedious. Especially for people not familiar with a terminal and/or not doing it on their local machine (such as via gcloud Cloud Shell), so that using copy-paste and a file editor is not trivial. I would say that *MOST* of our users are not savvy editing files with vim/emacs, and use IDE instead. That gets really tedious to do regularly. 

alternatively could us a real environment variable, and tell them to run
```bash
export PSOXY_TOKEN=`gcloud auth print-identity-token`
```

every hour.  That's going to be easier than editing a file.

tools/psoxy-test/README.md
```diff
@@ -1,13 +1,78 @@
...
+node cli.js -u https://us-central1-acme.cloudfunctions.net/v2/users -t <IDENTITY_TOKEN>
+```
+
+(*) You can obtain it by running `gcloud auth print-identity-token` (using Google Cloud SDK)
+
+## Testing all endpoints for a given data source
+We also provide some npm scripts that allow you to test all the endpoints for a given data source. These scripts rely on the same Node.js module that runs the CLI script described above, but instead of providing the options via CLI, you have to create a `.env` file with the following contents:
+```shell
+# For AWS deploys
+PSOXY_URL=<url>
```
eschultink, line 43: i think this DEFINITELY needs to be an argument, because it will differ for each proxy invocation.

eg, gcal, gmail, gdrive psoxy instances are all distinct cloud functions in GCP, so will have a different value for  `PSOXY_URL` when you run the suite. Hence I think little value in defining in a flat file

tools/psoxy-test/README.md
```diff
@@ -1,13 +1,78 @@
...
+PSOXY_ROLE=<role ARN>
+PSOXY_IMPERSONATE=<user email> # only for Google Workspace data sources
+PSOXY_SAVE_TO_FILE=true
+PSOXY_VERBOSE=true
```
eschultink, line 47: `verbose` for the test suite case isn't that useful imho, but OK. again, I think `-v` arg suffices; this is very much a debug thing, rather than something I expect customers to use themselves

eschultink (changes requested):

eschultink (approved):

PR #149, opened 2022-09-08
S130 (1/3) Dropbox support: new rules

aperez-worklytics:
- Added rules for Dropbox connector
- Updated intelliJ idea settings for fixing GCP module -it was ignored-

NOTE: Most of the lines included are due JSON test files

### Features
[Psosy: rules + config in proxy](https://app.asana.com/0/0/1202654062511673)

### Change implications


configs/dropbox-business.yaml
```diff
@@ -0,0 +1,5 @@
+SOURCE: dropbox-business
+TARGET_HOST: api.dropboxapi.com
+SOURCE_AUTH_STRATEGY_IDENTIFIER: oauth2_refresh_token
+GRANT_TYPE: refresh_token
```
aperez-worklytics, line 4: Connector config uses `refresh_token` handler; it will raise write in the log errors because Dropbox API it will also provides a new refresh_token during token call. 
See https://github.com/Worklytics/psoxy/blob/9c64b7e506d47e500a23ba7461be6013b30cfa22/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java#L144

In this case particular case for Dropbox, we don't need to update that token -the one that we initially is enough- even each time the API returns a new one. But in the other hand it is true that having log errors for this are weird. 

Options? 
  - a) add new handler to ensure we skip new refresh_token 
  - b) include a parameter to skip that part  in `OAuthRefreshTokenSourceAuthStrategy` -so no error will be raised
  - c) ?

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV2HTTPEventRequestAdapter.java
```diff
@@ -2,9 +2,11 @@
 
 import co.worklytics.psoxy.gateway.HttpEventRequest;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayV2HTTPEvent;
+import com.google.api.client.util.IOUtils;
```
eschultink, line 5: doesn't see this used ... and doesn't seem like the kind of thing we should depend on from an API client library.  I'd likely prefer using apache commons equivalent

eschultink (approved):
cool; one suggestion, no blockers

PR #162, opened 2022-09-22
S132 misc fixes

eschultink:
### Features
  - add'l slack field that can be pseudonymized
  - expose gcal calendarList endpoint

### Change implications


java/core/src/main/java/co/worklytics/psoxy/Sanitizer.java
```diff
@@ -37,6 +39,13 @@ class ConfigurationOptions implements Serializable {
 
         @Builder.Default
         PseudonymImplementation pseudonymImplementation = PseudonymImplementation.DEFAULT;
+
+
+        /**
+         * if set, sanitizer should only include domains in this list in response (domains not in
+         * this list should be hashed)
+         */
+        Optional<Set<String>> domainsToInclude;
```
aperez-worklytics, line 48: I'd include it when fully implemented in sanitizer as well, at the moment it seems only to have this new field as an option

aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/Sanitizer.java
eschultink: 🤦 yeah, this was intended for a different branch

java/core/src/main/java/co/worklytics/psoxy/rules/google/PrebuiltSanitizerRules.java
```diff
@@ -8,9 +8,11 @@
 import com.avaulta.gateway.pseudonyms.PseudonymEncoder;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import org.checkerframework.checker.regex.qual.Regex;
```
jlorper, line 11: what's this? I think this comes from a transitive dep

jlorper (approved):
Just one import that shouldn't be there

PR #165, opened 2022-09-27
S132 store short lived access token

jlorper:
AWS only
WIP: as main examples not yet changed, just development ones.
Features storing the short lived access token in the AWS parameter store for connectors that support shared tokens.

### Features
- [shared access token](https://app.asana.com/0/1203002526906932/1203033925140360)
- Grant lambdas read access to only their parameters and global ones (also write if needed)

### Change implications


infra/modules/aws-psoxy-lambda/main.tf
```diff
@@ -70,7 +70,25 @@ resource "aws_iam_role" "iam_for_lambda" {
...
-resource "aws_iam_policy" "policy" {
+# "PSOXY_${upper(replace(each.value.connector_name, "-", "_"))}_${upper(each.value.secret_name)}"
+data "aws_ssm_parameters_by_path" "psoxy_parameters" {
+  path            = "/"
+  with_decryption = false # we just want the arns, not the values
+}
+
+locals {
+  prefix          = "PSOXY_${upper(replace(var.source_kind, "-", "_"))}_"
+  prefix_writable = "PSOXY_${upper(replace(var.source_kind, "-", "_"))}_WRITABLE"
```
jlorper, line 81: WRITABLE parameters by convention.
This block reads all parameters and filters by prefix to grant read/write to specific parameters of the function.
The external ones needs to be passed in `parameters`

infra/modules/aws-psoxy-rest/variables.tf
```diff
@@ -81,4 +81,12 @@ variable "environment_variables" {
   type        = map(string)
   description = "Non-sensitive values to add to functions environment variables; NOTE: will override anything in `path_to_config`"
   default     = {}
-}
\ No newline at end of file
+}
+
+# TODO: values passed around but not used yet. Intended use is to grant access only to those listed here
+# TODO (cont): Right now lambdas have read grants to all ssm parameters
```
jlorper, line 87: I need to clean all these TODOs

infra/modules/psoxy-package/build.sh
```diff
@@ -7,14 +7,31 @@ set -e
...
+FOLDERS=("gateway-core" "core" "impl/$2")
 
-cd $1/core
-mvn package install > ${LOG_FILE} 2>&1
+touch ${LOG_FILE}
 
-cd $1/impl/$2
-mvn package >> ${LOG_FILE} 2>&1
+cd $1
+for index in "${!FOLDERS[@]}"
```
jlorper, line 15: this can be reverted if you want, still needs some tweaks, but basically allows for faster turnaround when testing lambdas but changes only affect terraform. It avoids regenerating the jars.
Pending: regenerate all if any is stale.

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -239,9 +239,10 @@ EOT
       secured_variables : [
         "CLIENT_SECRET",
         "CLIENT_ID",
-        "ACCOUNT_ID"
+        "ACCOUNT_ID",
+        "WRITABLE_ACCESS_TOKEN"
```
jlorper, line 243: WRITABLE, by convention, so instead of defining a read set and a write set of params, we can use just one.

eschultink: seems in correct - you had `WRITABLE` as suffix

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
```diff
@@ -70,7 +69,7 @@ public String load(String key) throws AwsServiceException {
                                     .build();
                                     GetParameterResponse parameterResponse = client.getParameter(parameterRequest);
                                     return parameterResponse.parameter().value();
-                                } catch (ParameterNotFoundException | ParameterVersionNotFoundException nfe) {
+                                } catch (SsmException ignore) {
```
eschultink, line 70: mmm ... I don't see why broaden this. You're now catching cases beyond the two "not found"s

infra/modules/psoxy-package/build.sh
eschultink: yeah, this is definitely a pain point that's worth solving; but perhaps decouple from this SSM stuff, as should be totally separate but we need to be absolutely certain it's working correctly before release

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -217,6 +243,32 @@ public Set<ConfigService.ConfigProperty> getRequiredConfigProperties() {
...
+            }
+        }
+
+        private void storeSharedAccessToken(@NonNull AccessToken accessToken) {
+            Preconditions.checkState(payloadBuilder.useSharedToken(), "Shared token not supported");
+            try {
+                config.putConfigProperty(ConfigProperty.WRITABLE_ACCESS_TOKEN, objectMapper.writerFor(AccessToken.class).writeValueAsString(accessToken));
+                log.log(Level.INFO, "New token stored in config");
+            } catch (JsonProcessingException e) {
+                log.log(Level.SEVERE, "Could not write token into JSON", e);
```
eschultink: for clarity, `Could not serialize token as JSON` (just because this is distinct case from other IO involved in overall "write" operation

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/AccessTokenDto.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Custom class that maps 1:1 with {@link com.google.auth.oauth2.AccessToken}
+ * Used for storage of the token in JSON. Google's is immutable and can't easily configure
+ * Jackson to do it.
+ */
+@NoArgsConstructor //for jackson
+@AllArgsConstructor
+@Getter
+public class AccessTokenDto {
```
eschultink, line 19: `Serializable`? `@Data`?

eschultink (approved):
eh, a couple comments.The more I think about it, the less I like the hackery of naming params with `WRITABLE` to apply a policy that allows writable policy to them. Seems to magic.  I think better just to pass two parameter lists to the lambdas.

infra/modules/aws-psoxy-lambda/main.tf
jlorper: changed for an object with a writeable attribute

infra/modules/psoxy-package/build.sh
jlorper: reverted, I'll work on a new branch with what I had so far

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
jlorper: Yeah, had issues with errors related to policies when splitting into read/write and granting params per function.
I've split those and log in the non-general case

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/AccessTokenDto.java
jlorper: Made in immutable. No need to be serializable as we do json back and forth

infra/modules/aws-psoxy-lambda/variables.tf
```diff
@@ -79,3 +74,18 @@ variable "log_retention_in_days" {
   description = "number of days to retain logs in CloudWatch for this psoxy instance"
   default     = 7
 }
+
+variable "global_parameters" {
```
eschultink, line 78: if it's arns, I'd just call it `global_parameter_arns` to be clear on the expected type.

eschultink (approved):

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
jlorper: To be clear: the issue is the nature of the fallback configuration approach, we try to read everything until we find it.
F.e. an environment variable, now with my policy restrictions, the error is not that is not found, but that there is no policy allowing to read that parameter, and that isn't an specific sub-classed `SsmException`.

jlorper:
@eschultink ready for final review. AWS examples done, Release data updated to 0.4.5 (tag not created yet)

infra/modules/aws/variables.tf
```diff
@@ -48,7 +48,7 @@ variable "caller_aws_arns" {
 variable "psoxy_version" {
   type        = string
   description = "version of psoxy to deploy"
-  default     = "0.4.4"
+  default     = "0.4.5-jrc"
```
eschultink, line 51: the `-jrc` thing confuses me

eschultink (approved):

aperez-worklytics (approved):
:+1: A couple of naming suggestions and following @eschultink , `-jrc` suffix thing

infra/modules/aws/variables.tf
jlorper: to clarify: was a "jose release candidate" to not collide with any other, did that temporarily to check validity of terraform files.

PR #166, opened 2022-09-28
s132  msft client id / refresh endpoint as env vars, rather than SSm

eschultink:
### Features
  - msft client id / refresh endpoint as env var, rather than ssm parameter

### Change implications

 - customers will see destroy of SSM parameters; update to lambdas - but nothing they need to actually do on next apply

---
  - https://app.asana.com/0/0/1203062039134020

### DIFF
--- a/infra/examples-dev/aws-msft-365/main.tf
+++ b/infra/examples-dev/aws-msft-365/main.tf
@@ -111,35 +111,6 @@ module "msft-connection-auth" {
   certificate_subject   = var.certificate_subject
 }
 
-resource "aws_ssm_parameter" "client_id" {
-  for_each = module.worklytics_connector_specs.enabled_msft_365_connectors
-
-  name  = "PSOXY_${upper(replace(each.key, "-", "_"))}_CLIENT_ID"
-  type  = "String"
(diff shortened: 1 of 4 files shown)


infra/examples-dev/aws-msft-365/main.tf
```diff
@@ -168,8 +139,10 @@ module "psoxy-msft-connector" {
   aws_account_id        = var.aws_account_id
   path_to_repo_root     = var.psoxy_base_dir
   environment_variables = {
-    PSEUDONYMIZE_APP_IDS = tostring(var.pseudonymize_app_ids)
     IS_DEVELOPMENT_MODE  = "true"
+    CLIENT_ID            = module.msft-connection[each.key].connector.application_id
+    REFRESH_ENDPOINT     = "https://login.microsoftonline.com/${var.msft_tenant_id}/oauth2/v2.0/token"
```
aperez-worklytics, line 144: Probably this URL could be part of connector specs module output for MSFT instead of being hardcoded here and in the other example

aperez-worklytics (approved):

PR #179, opened 2022-10-12
S133 limit query params

eschultink:
### Features
  - limit query params to an explicit subset (with regexes, it's hard to do anything but allow all)

### Change implications

---
  - https://app.asana.com/0/0/1203152530731887

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
@@ -1,6 +1,7 @@
 package co.worklytics.psoxy.rules;
 
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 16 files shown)


java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
```diff
@@ -90,6 +92,17 @@ public static class Endpoint {
 
         String pathRegex;
 
+        //if provided, only query params in this list will be allowed
+        @JsonInclude(JsonInclude.Include.NON_NULL)
+        List<String> allowedQueryParams;
```
aperez-worklytics, line 97: I'd renamed to something like `onlyAllowedQueryParams`; its absence will enforce that everything is allowed; its inclusion will mean that *only* the ones defined are allowed

aperez-worklytics (approved):
Some suggestion about allowedQueryParam name

PR #188, opened 2022-10-18
S134 asana support

aperez-worklytics:
As part of [asana 2.0](https://app.asana.com/0/302559783667519/1202400497708926), rules have been updated to use the ones we use in the new jobs and:
- Include `gid` and `email` fields for users as pseudonymized. 
- Include rules for *task search by workspace*; not yet implemented in the jobs but it will require to use that endpoint.

### Features
[Update rules for Proxy](https://app.asana.com/0/0/1203176250867029)

### Change implications


eschultink (approved):
ok, unfortunately cust wanting this has already deployed without this endpoint open; and 'gid' pseudonymization will be breaking change. Many standards (including APIs of vendors such as Atlassian/Asana) don't consider the gid to be PII, as it doesn't identify a natural person directly. You need a customer's asana data to do a lookup.

PR #190, opened 2022-10-19
reversibly encode asana user gids

eschultink:
### Features
  - recover `/workspaces` endpoint; useful for testing if nothing else
  - reversibly encode user gids - not sure we need this, but in case

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/rules/asana/PrebuiltSanitizerRules.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/rules/asana/PrebuiltSanitizerRules.java
@@ -23,9 +23,12 @@ public class PrebuiltSanitizerRules {
                     .jsonPath("$.data[*].photo")
                     .build())
             .transform(Transform.Pseudonymize.builder()
-                    .jsonPath("$.data[*].gid")
                     .jsonPath("$.data[*].email")
                     .build())
+            .transform(Transform.Pseudonymize.builder()
+                .includeReversible(true)
+                .jsonPath("$.data[*].gid")
+                .build())
             .build();
 
     static final Rules2.Endpoint TEAMS = Rules2.Endpoint.builder()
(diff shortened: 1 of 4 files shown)


java/core/src/main/java/co/worklytics/psoxy/rules/asana/PrebuiltSanitizerRules.java
```diff
@@ -23,9 +23,12 @@ public class PrebuiltSanitizerRules {
                     .jsonPath("$.data[*].photo")
                     .build())
             .transform(Transform.Pseudonymize.builder()
-                    .jsonPath("$.data[*].gid")
                     .jsonPath("$.data[*].email")
                     .build())
+            .transform(Transform.Pseudonymize.builder()
+                .includeReversible(true)
```
aperez-worklytics, line 29: Is it reversible only for `user` endpoints, or does it needs to be reversible for any `gid` belong to the user? Because in that case, that needs to be included in other rules too

java/core/src/main/resources/rules/asana/asana.yaml
```diff
@@ -53,6 +57,7 @@ endpoints:
           - "$.data[*].custom_fields[*].people_value[*].gid"
           - "$.data[*].followers[*].gid"
```
aperez-worklytics, line 58: For example, this 2 are users's gid too.

aperez-worklytics (commented):
Not sure the reversible scope; if needs to be applied to all users objects don't worry, I'll create a PR over this one

java/core/src/main/java/co/worklytics/psoxy/rules/asana/PrebuiltSanitizerRules.java
eschultink: no, just for user endpoints (where we might need it as arg to some follow-up call)

aperez-worklytics (approved):

PR #192, opened 2022-10-20
modularize aws 365 example

eschultink:
### Features
  - improve example initialization
  - modularize `aws-msft-365` example

### Change implications


infra/examples-dev/aws-msft-365/variables.tf
```diff
@@ -82,7 +82,6 @@ variable "enabled_connectors" {
 
   default = [
     "azure-ad",
-    "dropbox-business",
```
aperez-worklytics, line 85: Can we keep it ?

aperez-worklytics (approved):

infra/examples-dev/aws-msft-365/variables.tf
eschultink: eh, i don't think it makes a ton of sense by default, given how rare usage is

PR #194, opened 2022-10-21
support restricting HTTP method in rules

eschultink:
### Features
 - support for restricting http methods by endpoint

### Change implications

---
  - https://app.asana.com/0/0/1203208515850410

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
@@ -210,7 +210,7 @@ public class CommonRequestHandler {
                             .withPseudonymImplementation(pseudonymImplementation.get()));
                     }
 
-                    proxyResponseContent = sanitizer.sanitize("GET", targetUrl, responseContent);
+                    proxyResponseContent = sanitizer.sanitize(request.getHttpMethod(), targetUrl, responseContent);
                     String rulesSha = rulesUtils.sha(sanitizer.getConfigurationOptions().getRules());
                     builder.header(ResponseHeader.RULES_SHA.getHttpHeader(), rulesSha);
                     log.info("response sanitized with rule set " + rulesSha);
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
@@ -131,7 +131,7 @@ public class CommonRequestHandler {
         String callLog = String.format("%s %s", request.getHttpMethod(), relativeURL);
         if (skipSanitization) {
             log.info(String.format("%s. Skipping sanitization.", callLog));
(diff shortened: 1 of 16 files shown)


java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
```diff
@@ -103,6 +103,18 @@ public Optional<List<String>> getAllowedQueryParamsOptional() {
             return Optional.ofNullable(allowedQueryParams);
         }
 
+
+        //if provided, only http methods in this list will be allowed
+        @JsonInclude(JsonInclude.Include.NON_NULL)
+        List<String> allowedMethods;
```
aperez-worklytics, line 109:
```suggestion
        Set<String> allowedMethods;
```

java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
```diff
@@ -103,6 +103,18 @@ public Optional<List<String>> getAllowedQueryParamsOptional() {
...
+
+        @JsonIgnore
+        public Optional<List<String>> getAllowedMethods() {
```
aperez-worklytics, line 112:
```suggestion
        public Optional<Set<String>> getAllowedMethods() {
```

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -210,7 +210,7 @@ public HttpEventResponse handle(HttpEventRequest request) {
                             .withPseudonymImplementation(pseudonymImplementation.get()));
                     }
 
-                    proxyResponseContent = sanitizer.sanitize(targetUrl, responseContent);
+                    proxyResponseContent = sanitizer.sanitize("GET", targetUrl, responseContent);
```
aperez-worklytics, line 213: :yellow_circle:  Use the original http method:
```suggestion
                    proxyResponseContent = sanitizer.sanitize(request.getHttpMethod(), targetUrl, responseContent);
```

All the APIs we have work with GET, but Dropbox works with POST for all methods

aperez-worklytics (approved):
Some comments, the important one is marked with :yellow_circle: , after that change OK for me

aperez-worklytics (changes requested):
Provide the method used in the request instead of hardcoded GET

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
eschultink: :face_palm

jlorper (approved):
Ok, but we could have used  the `com.google.api.client.http.HttpMethods` constants everywhere.

eschultink:
> Ok, but we could have used the `com.google.api.client.http.HttpMethods` constants everywhere.

I want to migrate that to apache, bc Apache supports more sophisticated control of connections (https ciphers, etc)

PR #196, opened 2022-10-24
Updated Asana rules

aperez-worklytics:
Restrict query parameters for Asana

### Features
[restrict query params in asana rules](https://app.asana.com/0/1203169873202743/1203185839611254)

### Change implications

### DIFF
--- a/java/core/src/main/resources/rules/asana/asana.yaml
+++ b/java/core/src/main/resources/rules/asana/asana.yaml
@@ -2,7 +2,19 @@
 allowAllEndpoints: false
 endpoints:
   - pathRegex: "^/api/1.0/workspaces[?]?[^/]*$"
+    allowedQueryParams:
+      - "limit"
+      - "offset"
+      - "opt_pretty"
+      - "opt_fields"
   - pathRegex: "^/api/1.0/users[?]?[^/]*"
+    allowedQueryParams:
+      - "limit"
+      - "offset"
+      - "opt_pretty"
+      - "opt_fields"
+      - "workspace"
+      - "team"
     transforms:
       - !<redact>
         jsonPaths:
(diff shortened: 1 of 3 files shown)


java/core/src/main/resources/rules/asana/asana.yaml
```diff
@@ -39,6 +62,17 @@ endpoints:
           - "$.data[*].completed_by"
           - "$..name"
   - pathRegex: "^/api/1.0/tasks[?][^/]*"
+    allowedQueryParams:
+      - "limit"
+      - "offset"
+      - "opt_pretty"
+      - "opt_fields"
+      - "assignee"
```
eschultink, line 70: `assignee` is what I would *block*, to avoid allowing re-identification in practice (by passing id of known user through here, to get tasks with their pseudonym back)

java/core/src/main/resources/rules/asana/asana.yaml
```diff
@@ -2,7 +2,19 @@
 allowAllEndpoints: false
 endpoints:
   - pathRegex: "^/api/1.0/workspaces[?]?[^/]*$"
+    allowedQueryParams:
+      - "limit"
+      - "offset"
+      - "opt_pretty"
```
eschultink, line 8: if we don't need this, don't include it

java/core/src/main/resources/rules/asana/asana.yaml
```diff
@@ -2,7 +2,19 @@
...
+      - "opt_fields"
   - pathRegex: "^/api/1.0/users[?]?[^/]*"
+    allowedQueryParams:
+      - "limit"
+      - "offset"
+      - "opt_pretty"
+      - "opt_fields"
+      - "workspace"
+      - "team"
```
eschultink, line 17: are we using this?

eschultink (changes requested):
actually limit query params to what we really use (or are reasonably likely to use)

eschultink (approved):

eschultink (approved):

PR #202, opened 2022-11-02
S135 npm: bulk + logs

davidfq:
### Fixes
.

### Features
 - [npm-based test util for bulk file case](https://app.asana.com/0/1201039336231823/1203118181520484/f)
 - [npm-based log viewer](https://app.asana.com/0/1201039336231823/1203238032459548/f)

### Change implications

 - dependencies added/changed? **yes** Npm AWS sdk dependencies added


tools/psoxy-test/lib/aws.js
```diff
@@ -75,7 +89,230 @@ async function call(options = {}) {
...
+}
+
+
+/**
+ * Only for testing: List all available buckets
+ * Ref: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/classes/listbucketscommand.html
+ * Req: options.role requires "s3:ListAllMyBuckets" permissions
+ * @param {object} options 
+ */
+async function listBuckets(options) {
```
aperez-worklytics, line 208: Just as double-check, *ListAllMyBuckets* perm is enabled in bulk use case? Because if not, I'd drop this method.

aperez-worklytics (approved):

tools/psoxy-test/lib/aws.js
davidfq: It must be, as I've been using these methods as manual tests against your Psoxy instances...

eschultink: we shouldn't count on it. some customers deploy into shared accounts (even though we recommend against it)

PR #205, opened 2022-11-03
Fix environment variables resolution

davidfq:
.

### Fixes
- Environment variables resolution: 

Screenshot 2022-11-03 at 17 58 58

### Features
.

### Change implications


eschultink (approved):
if this works, looks good to me. certainly useful to be more robust.

eschultink:
WFM, so I'm merging this as I think it's good improvement and will be defensive against errors that will otherwise be uncaught.

PR #212, opened 2022-11-15
Fix default values for `environment_name`

davidfq:
### Fixes
- Default values for `environment_name` to "" instead of `null` that causes interpolation errors

### Features
.

### Change implications


aperez-worklytics (approved):
Just curiosity, what is the issue with null? Is it just for displaying names or other stuff?

davidfq:
Terraform errors when interpolating it:
```
resource "google_project" "psoxy-google-connectors" {
  name            = "Worklytics Connect%{if var.environment_name != ""} - ${var.environment_name}%{endif}"
```

if not filled-in, the value was `null` and Terraform would accept the `if` as truthy, then fails when printing `null`

eschultink (approved):

PR #221, opened 2022-12-20
s138 : include instructions re Google Workspace Service Account setup

eschultink:
### Features
  - suggest that customers create Google user to be a 'Service Account' for their Workspace directory connections that require it

### Change implications

---
  - https://app.asana.com/0/0/1203457917096945


eschultink:
created from `main` for cleaner PR. and can go either to `main` or next release bc no functional/breaking changes. But given that examples ref releases, rather than `main` - no prod users will get this until we do a release, and they update version numbers in their configs.

aperez-worklytics (approved):

PR #223, opened 2022-12-23
Canonical equivalence for emails with dots

jlorper:
If provider allows it add configuration to be able to make canonical equivalents of emails with dots.

### Change implications


jlorper:
Dismissed as wasn't the source of issue with identities

PR #225, opened 2023-01-05
Pseudonym implementation as a config setting

eschultink:
### Fixes
 - no way to configure legacy pseudonym implementation on non-REST connectors

### Change implications

---
  - https://app.asana.com/0/0/1203651923592450

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
@@ -32,6 +32,7 @@ import javax.inject.Inject;
 import java.io.IOException;
 import java.net.URL;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Supplier;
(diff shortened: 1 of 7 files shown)


java/gateway-core/src/main/java/com/avaulta/gateway/pseudonyms/PseudonymImplementation.java
```diff
@@ -25,4 +29,13 @@ public static PseudonymImplementation parseHttpHeaderValue(String httpHeaderValu
...
     }
+
+    public static PseudonymImplementation parseConfigPropertyValue(String configPropertyValue) {
+        for (PseudonymImplementation impl : values()) {
+            if (impl.getHttpHeaderValue().equals(configPropertyValue)) {
+                return impl;
+            }
+        }
+        throw new IllegalArgumentException("Unknown pseudonym implementation: " + configPropertyValue);
+    }
```
jlorper, line 40: methods are identical, maybe merge both into `parseValue` ??

jlorper: A functional approach
```suggestion
    public static PseudonymImplementation parseValue(String configPropertyValue) {
        return Arrays.stream(values())
            .filter( p -> p.getHttpHeaderValue().equals(httpHeaderValue))
            .findFirst()
            .orElseThrow( () -> new IllegalArgumentException("Unknown pseudonym implementation: " + httpHeaderValue));
    }
```

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -224,6 +221,23 @@ public HttpEventResponse handle(HttpEventRequest request) {
         }
     }
 
+    /**
+     * encapsulates dynamically configuring Sanitizer based on request (to support some aspects of
+     * its behavior being controlled via HTTP headers)
+     *
+     * @param request
+     */
+    void dynamicallyConfigureSanitizer(HttpEventRequest request) {
```
jlorper, line 230: I'd make this return the sanitizer so it is explicit there is a change, rather than recreate internally, it is hidden side-effect.
I'd do something like this:

```
Optional<Sanitizer> perRequestSanitizer = requestBasedSanitizer(request);
if (perRequestSanitizer.isPresent()) {
  sanitizer = perRequestSanitizer.get();
}
```

jlorper (approved):
Works as is, but I think it can be improved.

Issue #200, opened 2022-10-29
Make Java builds reproducible to reduce Lambda churn


patricksanders:
When deploying psoxy to AWS using the provided Terraform, the function code bundles get updated with every `plan`/`apply`. I believe this can be avoided by [making builds reproducible](https://maven.apache.org/guides/mini/guide-reproducible-builds.html). This looks fairly straightforward, assuming you can require versions of dependencies that support reproducible builds.

eschultink:
thanks @patricksanders  - definitely something we'd like to deal with. Gave it a quick shot and found that a Maven plugin we're using (to "flatten" the `pom.xml`) isn't compatible with the approach you link for reproducible builds. 

We need to investigate if we can avoid using the `flatten` plugin, which may avoid this.

eschultink:
hey @patricksanders - didn't actually do this, but #226 adds simple check for the build artifact and avoids rebuilding it if it exists, unless you set `force_bundle = true` in your `terraform.tfvars`.  Forcing the bundle should really only be needed if you're changing Java code, which for customer use cases we wouldn't expect.

eschultink:
we don't utilize reproducible builds; but now terraform modules by default will not re-build on every terraform plan/apply. additionally modules support a prebuilt bundle being passed in. we also produce bundles that are consumable from gcs/s3, which can be used; or customers may build their own bundle and reference an gcs/s3 url for cloud run / lambda to use as the bundle.

PR #226, opened 2023-01-06
avoid lots of logs, java compliation

eschultink:
### Features
  - avoid tons of `.log` files generated by re-deploys
  - avoid re-build of proxy JAR, *unless* version change *or* forced with a parameter (only developers really need to force re-builds; everyone else should be able to depend on version number changing)

NOTE: this likely renders #200 moot; deterministic builds would avoid unnecessary re-deploys of lambda, but actual Java compilation would still occur. This avoids compilation of Java code entirely.

### Change implications

---
  - https://app.asana.com/0/0/1203644805511066
  - https://app.asana.com/0/0/1203246719435623


jlorper:
Looks good, maybe add a commented sample setting in `terraform.tfvars.example` ??

jlorper (approved):

PR #228, opened 2023-01-06
polish for release of v0.4.9

eschultink:
### Fixes
  - add default values to avoid type problems, reference problems
  - use `try` rather than `coalesce` in most places
  - fix bug with secret references that was breaking GCP deployment

### Change implications

---
  - https://app.asana.com/0/0/1203665072663826

### DIFF
--- a/infra/modular-examples/gcp-google-workspace/main.tf
+++ b/infra/modular-examples/gcp-google-workspace/main.tf
@@ -126,7 +126,7 @@ module "psoxy-google-workspace-connector" {
   secret_bindings = merge({
     # as SERVICE_ACCOUNT_KEY rotated by Terraform, reasonable to bind as env variable
     SERVICE_ACCOUNT_KEY = {
-      secret_id      = module.google-workspace-key-secrets[each.key].secret_ids["PSOXY_${replace(upper(each.key), "-", "_")}_SERVICE_ACCOUNT_KEY"]
+      secret_id      = module.google-workspace-key-secrets[each.key].secret_secret_ids["PSOXY_${replace(upper(each.key), "-", "_")}_SERVICE_ACCOUNT_KEY"]
       version_number = module.google-workspace-key-secrets[each.key].secret_version_numbers["PSOXY_${replace(upper(each.key), "-", "_")}_SERVICE_ACCOUNT_KEY"]
     }
   }, module.psoxy-gcp.secrets)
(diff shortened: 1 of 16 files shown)


infra/modular-examples/gcp-google-workspace/main.tf
```diff
@@ -126,7 +126,7 @@ module "psoxy-google-workspace-connector" {
   secret_bindings = merge({
     # as SERVICE_ACCOUNT_KEY rotated by Terraform, reasonable to bind as env variable
     SERVICE_ACCOUNT_KEY = {
-      secret_id      = module.google-workspace-key-secrets[each.key].secret_ids["PSOXY_${replace(upper(each.key), "-", "_")}_SERVICE_ACCOUNT_KEY"]
+      secret_id      = module.google-workspace-key-secrets[each.key].secret_secret_ids["PSOXY_${replace(upper(each.key), "-", "_")}_SERVICE_ACCOUNT_KEY"]
```
jlorper, line 129: secret_secret_ids is kind of confusing

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -227,15 +227,19 @@ public HttpEventResponse handle(HttpEventRequest request) {
      *
      * @param request
      */
-    void dynamicallyConfigureSanitizer(HttpEventRequest request) {
+    Sanitizer getSanitizerForRequest(HttpEventRequest request) {
         Optional<PseudonymImplementation> pseudonymImplementation = parsePseudonymImplementation(request);
         if (pseudonymImplementation.isPresent()) {
+            loadSanitizerRules(); // ensure sanitizer is loaded
```
jlorper, line 233: you can remove this, `loadSanitizerRules` is one of the very first calls in the handler, so no point in doing it again.

jlorper (approved):

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
eschultink: tests req'd it

infra/modular-examples/gcp-google-workspace/main.tf
eschultink: yeah, agreed; unfortunately was introduced in earlier version so a breaking change to just rename, but I'll mark deprecated, duplicate it with a better name, and use that.

PR #229, opened 2023-01-07
google projects in examples

eschultink:
### Features
  - stop managing GCP projects in Terraform; makes it harder to configure for most customers, as 80% case is that the terraform user can't create projects, so needs to get project_id/org_id/billing_account_id/folder_id from someone else, then import. 

### Fixes
  - changelogs duplicated

### Change implications


CHANGELOG.md
```diff
@@ -19,9 +19,44 @@ Working tracking of changes, updated as work done prior to release.  Please revi
   - keys/salts per value kind (PII, item id, etc)
 
 
+## v0.4.9
+
+Breaking Changes: (at least, somewhat breaking)
```
jlorper, line 24: just an idea to improve readability is to split this by deployment kind, so customers in AWS don't need to read GCP stuff and viceversa. Like:

## v0.4.9
- Breaking changes (if any, for common stuff)
- Features (if any, for common stuff)
### Changes for AWS Deployments
- Breaking changes (if any)
- Features (if any)
### Changes for GCP Deployments
- Breaking changes (if any)
- Features (if any)

jlorper (approved):
Just one suggestion

CHANGELOG.md
eschultink: well, it's a 2x2 matrix kinda - host platform X data source. So in this case breaking changes if EITHER deploying to GCP *or* using Google Workspace.

PR #230, opened 2023-01-09
avoid caching accessTokens in process memory

eschultink:
### Features
  - support to mark `ConfigProperty` so won't be cached
  - use this to avoid caching accessToken, which are potentially subject to refreshing. Main point of this is to avoid one lambda instance repeatedly attempting to reuse a token that has been refreshed by another process. Eg 

      - 0: A reads token, caches it
      - 30s: B reads token, sees needs refresh; refreshes and writes to config; 
      - 45s: A invoked again, still sees original token in its cache, so sees need to refresh; does so and writes to config.
      - _if A's refresh causes B's token to be invalidated by Resource Server, we could enter perpetually broken state. B's next use fails with 401 invalid token, so it refreshes - invalidating A's token again; now in loop_

### Change implications

---
  - https://app.asana.com/0/0/1203651346414871


java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
```diff
@@ -11,18 +11,55 @@ interface ConfigProperty extends Serializable {
...
+
+        /**
+         * @return whether cached value for property must be revalidated with origin before re-use
+         *  (equivalent to HTTP `Cache-Control: no-cache` semantics as defined by RFC 7234)
+         *
+         * @see 'https://www.rfc-editor.org/rfc/rfc7234#section-5.2.2'
+         */
+        default Boolean noCache() {
+            return false;
+        }
```
eschultink, line 23: This is actually too conservative for the ACCESS_TOKEN use case. If we had memcache / redis caching at proxy- instance level (rather than lambda/cloud function instance level), we could use that just fine.  

The problem is caching at process/thread memory such that all lambda/cloud function instances don't see changes. 

so could drop reliance on standard definition of `no-cache` and call this `noLocalCache` to allow for future use of memcache/redis as caching solution in front of SSM ParameterStore / GCP Secret Manager / etc.

java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
```diff
@@ -11,18 +11,55 @@ interface ConfigProperty extends Serializable {
...
+     * get property as defined in this ConfigService, wrapped in Optional
+     *
+     * @see Optional
+     *
+     * @param property to retrieve value for
+     * @return filled Optional if defined; empty otherwise
+     */
     Optional<String> getConfigPropertyAsOptional(ConfigProperty property);
 
+    @Deprecated // use EnvVarsConfigService::isDevelopment
```
aperez-worklytics, line 62: If you want to include it as part of javadoc:

```suggestion
     /**
     *
     * @deprecated use EnvVarsConfigService::isDevelopment
     */
    @Deprecated
```

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CachingConfigServiceDecorator.java
```diff
@@ -24,9 +26,11 @@ public class CachingConfigServiceDecorator implements ConfigService {
 
     private final Object $writeLock = new Object[0];
 
-    private final String NEGATIVE_VALUE = "##NO_VALUE##";
+    @VisibleForTesting
+    static final String NEGATIVE_VALUE = "##NO_VALUE##";
```
aperez-worklytics, line 30: Just personal taste, but each time I see this I'm thinking in a *pure negative* value (<0); I understand that maybe what we want to represent is a value not defined in cache, so maybe... ?

```suggestion
    static final String NO_CACHE_VALUE = "##NO_VALUE##";
```

java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
aperez-worklytics: Just personal taste and I know the RFC, but maybe clear if the property is just `cache` so when defining the properties in *ConfigService.ConfigProperty* implementations we don't need to use the double negation ( `REFRESH_ENDPOINT(false)` -> `REFRESH_ENDPOINT(true)` )

aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CachingConfigServiceDecorator.java
```diff
@@ -53,33 +57,39 @@ public boolean supportsWriting() {
...
-                throw e.getCause();
+        if (property.noCache()) {
+            return delegate.getConfigPropertyAsOptional(property);
+        } else {
+            try {
+                String value = getCache().get(property);
+                if (Objects.equals(NEGATIVE_VALUE, value)) {
+                    return Optional.empty();
+                } else {
+                    return Optional.of(value);
```
jlorper, line 84:
```suggestion
                    return Optional.ofNullable(value);
```

eschultink: `value` should be `##NEGATIVE_VALUE##` if the value isn't in cache; I don't think it would be possible for `value` to ever be `null`.  I think you would have to configure it as `""` or something.

PR #231, opened 2023-01-10
persist `refresh_token` if changed; randomize proactive `access_token` refreshes to avoid race cases

eschultink:
### Features
 - support persisting new `refresh_token` if rotated by a token request, and if stored one is older than 7 days (avoid token churn, for sources which seem to ALWAYS send us new refresh token)
 - randomize proactive token refreshes, so refresh-races less likely (eg, if 40 shards running, expect 7.5s before concurrent refresh attempted; so very likely that first one wins, and even worst case likely to be much better than current situation)

### Change implications

 - dependencies added/changed? **junit, mockito upgrade to latest versions** --> as these are `test` dep, implications should be minimal assuming tests actually pass

---
  - https://app.asana.com/0/0/1203677915687493


jlorper (approved):

aperez-worklytics (approved):

java/core/src/test/java/co/worklytics/test/MockModules.java
```diff
@@ -23,6 +26,19 @@ static ConfigService configService() {
...
+    @Module
+    public interface ForRandomNumberGenerator {
+        @Provides @Singleton
+        static RandomNumberGenerator random() {
+            //NOTE: only works for jdk17+ with Mockito versions that include fix for:
+            // https://github.com/mockito/mockito/issues/2589
+            // --> possibly not coming until Mockito 5????
+            //return mock(new Random());
+
+            return mock(RandomNumberGenerator.class);
```
eschultink, line 38: introduced this interface, bc couldn't mock `Random` with mockito in java 17+

java/core/src/main/java/co/worklytics/psoxy/utils/RandomNumberGenerator.java
```diff
@@ -0,0 +1,7 @@
+package co.worklytics.psoxy.utils;
+
+public interface RandomNumberGenerator {
```
eschultink, line 3: interface used to hide `Random`, which can't be mocked with mockito in jdk17+ (until Mockito 5)

eschultink:
marked for re-review, as added a couple other features to PR

aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -183,7 +194,7 @@ public AccessToken refreshAccessToken() throws IOException {
                 }
             }
 
-            if (isCurrentTokenValid(this.currentToken, clock.instant())) {
+            if (needsRefresh(this.currentToken, clock.instant())) {
```
jlorper, line 197: sorry I missed this change, semantics are inverted: if needs refresh then return current token
Method should be negated or kept as `isCurrentTokenValid`

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -209,22 +220,42 @@ private CanonicalOAuthAccessTokenResponseDto exchangeRefreshTokenForAccessToken(
...
+         * @param tokenResponse the token response
+         */
+        @VisibleForTesting
+        void storeRefreshTokenIfRotated(CanonicalOAuthAccessTokenResponseDto tokenResponse) {
+            if (!StringUtils.isBlank(tokenResponse.getRefreshToken())) {
+                //if a refresh_token came back from server, potentially update it
+                config.getConfigPropertyWithMetadata(RefreshTokenTokenRequestBuilder.ConfigProperty.REFRESH_TOKEN)
+                    .filter(storedToken -> !Objects.equals(storedToken.getValue(), tokenResponse.getRefreshToken()))
+                    .filter(storedToken -> storedToken.getLastModifiedDate().isEmpty()
+                        || storedToken.getLastModifiedDate().get()
```
jlorper, line 242: are you sure this makes sense??? if a new refresh token comes I'd just assume the old one has been revoked. Id' store it anyway.

PR #232, opened 2023-01-10
expose metadata from config service

eschultink:
### Features
 - remove 'ACCESS_TOKEN' parameter from payload builder implementation; not coupled to that
 - clean up 'required parameters' logic

### Change implications

---
  - https://app.asana.com/0/0/1203696578378049


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -221,6 +227,37 @@ public HttpEventResponse handle(HttpEventRequest request) {
...
+                .map(param -> Pair.of(param, config.getConfigPropertyWithMetadata(param)
+                        .map(ConfigService.ConfigValueWithMetadata::getLastModifiedDate)
+                        .map(lastModified -> lastModified.map(Instant::toString)
+                                            .orElse("not-available") // config service that doesn't support it
+                        )
+                        .orElse("")))
+                .map(p -> p.getKey() + "=" + p.getValue())
+                .collect(Collectors.joining(";")));
+
+        //q: make this all conditional on some control header being passed (e.g. `X-Psoxy-Debug-Config`)?
```
jlorper, line 258: I was about to comment this, no real need to do a full request, just configuration.
We should include more stuff in the response
- proxy version (should be a constant in the deployment somehow)
- basic config check on all parameters, list all and return true/false if blank, so we can quickly see misconfigurations.

eschultink: expanded a bit; but w/o reflection, there's not a full list of ConfigParameters anywhere.

we could dump keys from env, but would likely be large and potentially unsafe to do so (if someone adds something weird to env on all their lambdas or something).

java/core/src/main/java/co/worklytics/psoxy/HealthCheckResult.java
```diff
@@ -12,17 +17,39 @@
 @AllArgsConstructor
 public class HealthCheckResult {
 
+    // TODO: get this from pom.xml or something
+    @Builder.Default
+    String version = "v0.4.10";
+
+    //q: terraform module version?? (eg, have terraform deployment set its version number as ENV
```
aperez-worklytics, line 24: :+1:

java/core/src/main/java/co/worklytics/psoxy/ResponseHeader.java
```diff
@@ -16,7 +16,17 @@ public enum ResponseHeader {
...
+
+    /**
+     * how source authentication is configured to be performed (non-sensitive config values)
+     */
+    SOURCE_AUTH("Source-Auth"),
+
+    /**
+     * when various config values were filled (timestamps, not actually config values)
+     */
+    CONFIG_FILLED("Config-Filled");
```
aperez-worklytics, line 29: Then.. ?

```suggestion
    CONFIG_FILLED_TIMESTAMP("Config-Filled");
```

aperez-worklytics (approved):

PR #233, opened 2023-01-12
fix ssm namespace support; add support in terraform examples

eschultink:
### Fixes
 - SSM config behavior uses PATH_TO_INSTANCE_CONFIG / PATH_TO_SHARED_CONFIG in consistent manner (currently, adds `_` to it). 
 - tweaked behavior so works with values that include `/`; AWS handles any SSM param name containing `/` differently, and requires that it also begin with a `/`. So added both validation and defensive checks that add `/` in necessary cases to address this.

### Features
 - AWS examples for MSFT + Google both updated to support passing in a path to use as a "root" for creating SSM parameters other than default (which implicitly would be `""` or `/`, however you think of it). These are tested for both default and pathed cases (Eg, `""`, `"/corp_it/worklytics/"` equivalents.

### Change implications

---
  - https://app.asana.com/0/0/1203715901381851


jlorper (approved):
Looks good, but we should be careful and make sure all config paths work as expected.

eschultink:
> Looks good, but we should be careful and make sure all config paths work as expected.

yeah, it's deeper bc of IAM policies. I was struggling to get this to work in my env, then realized it was bc we ALSO have to set in the IAM policies.

infra/modules/aws-psoxy-rest/main.tf
```diff
@@ -11,18 +11,19 @@ terraform {
...
+  handler_class                   = "co.worklytics.psoxy.Handler"
+  path_to_function_zip            = var.path_to_function_zip
+  function_zip_hash               = var.function_zip_hash
+  memory_size_mb                  = var.memory_size_mb
+  timeout_seconds                 = 55
+  reserved_concurrent_executions  = var.reserved_concurrent_executions
+  path_to_config                  = var.path_to_config
+  source_kind                     = var.source_kind
+  function_parameters             = var.function_parameters
+  path_to_instance_ssm_parameters = var.path_to_instance_ssm_parameters
```
eschultink, line 24: the real change. re-aligns rest of columns though ... 🤷

infra/modules/aws-ssm-secrets/main.tf
```diff
@@ -3,18 +3,26 @@
 #   eg, GCP Secret Manager, AWS SSM Parameter Store, Hashicorp Vault, etc.
 #  but is this good Terraform style? clearly in AWS case, this module doesn't do much ...
 
+locals {
+  # AWS SSM param name must be fully qualified if contains `/`;
+  # so test for that case, and prefix with `/` if needed
+  non_empty_path           = length(var.path) > 0
+  non_fully_qualified_path = length(regexall("/", var.path)) > 0 && !startswith(var.path, "/")
+  path_prefix              = local.non_empty_path && local.non_fully_qualified_path ? "/${var.path}" : var.path
```
eschultink, line 11: this is defensive, as ultimately added validation at the upper levels of the (eg `examples` and `modular-examples/`); but still this *should* do the right thing if someone uses our modules directly

infra/modules/aws-psoxy-lambda/main.tf
```diff
@@ -8,6 +8,16 @@ terraform {
...
 }
 
+locals {
+  instance_ssm_prefix = coalesce(var.path_to_instance_ssm_parameters, "${upper(replace(var.function_name, "-", "_"))}_")
+
+  # parse PATH_TO_SHARED_CONFIG in super-hacky way
+  # expect something like:
+  # arn:aws:ssm:us-east-1:123123123123:parameter/PSOXY_SALT
+  salt_arn              = [for l in var.global_parameter_arns : l if endswith(l, "PSOXY_SALT")][0]
+  path_to_shared_config = regex("arn.+parameter/(.*)PSOXY_SALT", local.salt_arn)[0]
```
eschultink, line 18: so this is aggressively hacky ... could have added a variable to pass the `path_to_shared_config` in as a variable, but then what do if it's inconsistent with these arns? 

Could add validation check that `path_to_shared_config` is indeed in the ARNs, but that would ultimately look much like the regex I'm doing here. So imho this is the best thing to do until a `v0.5` where we avoid passing explicit ARNs, and require `path_to_shared_config` to be a path that each proxy instance should be able to read ALL descendants of.

infra/modules/vault-access-token/main.tf
```diff
@@ -10,28 +10,28 @@
...
 
 resource "vault_token" "periodic_service_token" {
-  role_name        = vault_token_auth_backend_role.lambda_role.role_name # name to give token's role in Vault
-  period           = "${14*24}h" # 2 weeks
-  renewable        = true
-  renew_min_lease  = 8*24*60*60  # 8 days
-  renew_increment  = 14*24*60*60 # 2 weeks
-  policies         = [var.vault_policy_name]
+  role_name       = vault_token_auth_backend_role.lambda_role.role_name # name to give token's role in Vault
+  period          = "${14 * 24}h"                                       # 2 weeks
```
aperez-worklytics, line 25: Too much spaces for this value?
```suggestion
  period          = "${14 * 24}h" # 2 weeks
```

aperez-worklytics (approved):

infra/modules/vault-access-token/main.tf
eschultink: must have been `terraform fmt`

PR #236, opened 2023-01-13
deep link todos

eschultink:
### Features
 - integrate deep-linked URLs into TODOs for connections to Worklytics

### Change implications

---
  - https://app.asana.com/0/0/1203715826050293

### DIFF
--- a/infra/modules/worklytics-connector-specs/main.tf
+++ b/infra/modules/worklytics-connector-specs/main.tf
@@ -35,6 +35,7 @@ locals {
     },
     "gcal" : {
       source_kind : "gcal",
+      worklytics_connector_id : "gcal-psoxy",
       display_name : "Google Calendar"
       apis_consumed : [
         "calendar-json.googleapis.com"
@@ -52,6 +53,7 @@ locals {
     },
     "gmail" : {
       source_kind : "gmail",
+      worklytics_connector_id : "gmail-meta-psoxy",
       display_name : "GMail"
       apis_consumed : [
         "gmail.googleapis.com"
@@ -67,6 +69,7 @@ locals {
     },
     "google-chat" : {
       source_kind : "google-chat",
+      worklytics_connector_id : "google-chat-psoxy",
       display_name : "Google Chat"
       apis_consumed : [
         "admin.googleapis.com"
@@ -82,6 +85,7 @@ locals {
     },
     "google-meet" : {
       source_kind : "google-meet",
+      worklytics_connector_id : "google-meet-psoxy",
       display_name : "Google Meet"
       apis_consumed : [
         "admin.googleapis.com"
@@ -97,6 +101,7 @@ locals {
     },
     "gdrive" : {
       source_kind : "gdrive",
+      worklytics_connector_id : "gdrive-psoxy",
       display_name : "Google Drive"
       apis_consumed : [
         "drive.googleapis.com"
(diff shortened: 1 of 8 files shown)


infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -409,6 +424,7 @@ EOT
     }
     "qualtrics" = {
```
davidfq, line 425: I didn't notice this before: is this source kind OK? `Integration.json` seed doesn't define `sourceKindId` for this connector

davidfq (approved):
OK, but I'm not sure about `https://intl.worklytics.co/...` references. Shouldn't we always use central for this kind of link? Fewer redirections will happen for non-logged-in users (in this context, we should assume they aren't logged in).

non-logged-in wil be like: `intl` -> `app` -> `intl`

infra/modules/worklytics-psoxy-connection-generic/main.tf
```diff
@@ -22,6 +34,28 @@ Complete the following steps in Worklytics AFTER you have deployed the Psoxy ins
   5. Select `${var.psoxy_host_platform_id}` for "Proxy Instance Type".${local.per_setting_instructions_text}
   ${local.last_step}. Review any additional settings that connector supports, adjusting values as you see fit, then
      click "Connect".
+EOT
+
+
+  deep_link_instructions = <<EOT
+  1. Ensure you're authenticated with Worklytics. Either sign-in at `https://app.worklytics.co` with
```
davidfq, line 41: We should have a clear pattern for this: always link to central, deep linking feature should persist path and params

aperez-worklytics (approved):

infra/modules/worklytics-connector-specs/main.tf
eschultink: it uses the generic importer (`survey-import-psoxy`), with parser=SURVEY or something like that. Calling it `qualtrics` just distinguishes it, and makes it clearer what kind of survey data the customer is dealing with. 

https://github.com/Worklytics/psoxy/pull/236/files/872e3e5980ff562fed1002af56a791706533909a..fb1c9e17eaea0e165b87452cd5ab366f9a66d8b2#diff-a95861f4521ed65205892ca0024ea417b0cdf21bd828a3eb3126811b7556bdeeR426-R429

all the connectorIds are explicitly defined there.

the instance id in the proxy isn't used for anything but distinguishing the different lambdas on the AWS side, really.  In theory, customer could have multiple instances mapping to same connector on Worklytics-side (different connections), with same source Id. eg (`gcal-1`, `gcal-2`, etc); our examples are organized for the usual case of single connection per so, but nothing about the proxy modules themselves constrains this.

infra/modules/worklytics-psoxy-connection-generic/main.tf
eschultink: yeah, you're right; I'll take this as a tech-debt task. In practice, customers no longer care about EU if they're proxy customer though.  Although if we have large EU-based customer they might.  Since the proxy, we no longer seem to get the case from customers where even US based customers are interested in EU for GDPR compliance purposes; people now recognize it doesn't really matter, and that proxy does more towards GDPR compliance than EU hosting would.

eschultink: (as an aside, I think we can drop EU support; but there's a strong case to keep ability to have central auth and then redirect people out a Google project specific to them. eg, instead of `app-eu.worklytics.co`, we'd have `acme.worklytics.co` and we could put that project wherever they want, and the data/processing/etc would all be totally segregated by customer.  A lot of overhead to do that atm, but in future it could make sense if a customer is big enough that the efficiencies of sharing some resources is less than benefits of segregation.  A real GCP project for production use has a fixed base cost of ~$1500/m though, once we activate all the security stuff it requires - so not really efficient to do unless customer really values it.)

PR #242, opened 2023-01-26
S136 msft apps with federation

aperez-worklytics:
Supports *federation identity* workflow between AAD <-> AWS and AAD <--> GCP.

This authentication workflow is a `client_credentials` but the client assertion claim is obtained externally instead of using a credential o certificate; being that client assertion a JWT generated from the other part. That JWT needs to have same values in *issuer* (iss) , *audience* (aud) and *subject* (sub) claims to match the value defined in AAD as federated credentials. 

From AAD perspective, adding a federated credential is just a different type of credentials associated to an application. It needs to include an issuer, audience, subject and some administrative fields like name and description.

From AWS/GCP side, *issuer* has same value (different for AWS than for GCP) but the rest are a bit different due implementation details:
- For GCP: To obtain the token is just a call to an internal *identity* endpoint where we pass the audience we want to use. That endpoint returns a JWT with that audience; being the subject the id of the SA which requested the token. *audience* can be anything meanwhile it matches the AAD federated credential, but just rely on defaults suggested by MSFT documentation.
- For AWS: This is the more complicated stuff, as it involves the creation of a `Cognito Identity` resource with a pool where an identity is created matching the AAD federated credential: 
  - The *audience* is the *id* of the *identity pool* created in *Cognito*
  - The *subject* is the value of the *IdentityId* of the *identity* created in Cognito for that pool, which has been created based on the developer name + the AAD clientId

See my comments on my review for further details

Testing:
Through `example-dev/aws-msft-365` and `example-dev/gcp-workspace` using `outlook-mail` connector; testing tool returns valid result in both.

### Features
[use identity federation to auth with MSFT instead of certificates](https://app.asana.com/0/180812557884176/1202200981451897)

### Change implications

   - `cognitoidentity` for AWS; to get a token to use as a JWT as bearer claim


configs/azure-ad.yaml
```diff
@@ -1,6 +1,6 @@
 SOURCE: azure-ad
 TARGET_HOST: graph.microsoft.com
 SOURCE_AUTH_STRATEGY_IDENTIFIER: oauth2_refresh_token
-GRANT_TYPE: client_credentials # by default, assumed to be of type 'urn:ietf:params:oauth:client-assertion-type:jwt-bearer'
+GRANT_TYPE: workload_identity_federation # by default, assumed to be of type 'urn:ietf:params:oauth:client-assertion-type:jwt-bearer'
```
aperez-worklytics, line 4: Name I've used instead of `client_credentials`; not sure it fits well as it is how its named in GCP and AAD but not in AWS (just federated credentials would fit better?)

infra/.gitignore
```diff
@@ -5,3 +5,5 @@
 # PEM files used in local cert generation; don't want to commit them
 **.pem
 
+# Cognito identity files generation
+cognito_identity_*.json
```
aperez-worklytics, line 9: Files used to get the identity from cognito; see that module

infra/examples-dev/aws-msft-365/main.tf
```diff
@@ -8,7 +8,7 @@ terraform {
 
     # for API connections to Microsoft 365
     azuread = {
-      version = "~> 2.0"
+      version = "~> 2.3"
```
aperez-worklytics, line 11: Need to bump it because otherwise `audience` will require to be a valid URI when creating a federated credential; true for GCP but not for AWS.

infra/examples-dev/aws-msft-365/variables.tf
```diff
@@ -66,11 +66,6 @@ variable "connector_display_name_suffix" {
   default     = ""
 }
 
-variable "certificate_subject" {
```
aperez-worklytics, line 69: Drop any certificate usage and its module

infra/modules/aws-cognito-identity-cli/addIdentity.sh
```diff
@@ -0,0 +1,22 @@
...
+
+
+export $(printf "AWS_ACCESS_KEY_ID=%s AWS_SECRET_ACCESS_KEY=%s AWS_SESSION_TOKEN=%s" \
+$(aws sts assume-role \
+ --role-session-name="session_for_$CONNECTOR_ID" \
+ --role-arn=$ROLE \
+ --query "Credentials.[AccessKeyId,SecretAccessKey,SessionToken]" \
+ --output text))
+
+aws cognito-identity get-open-id-token-for-developer-identity --identity-pool-id $IDENTITY_POOL_ID --logins $LOGIN_ID --region $REGION > $OUTPUT/cognito_identity_$CONNECTOR_ID.json
```
aperez-worklytics, line 22: There are no command in TF for doing this, so I need to extract it as a shell script

infra/modules/aws-cognito-identity-cli/addIdentity.sh
```diff
@@ -0,0 +1,22 @@
...
+IDENTITY_POOL_ID=$1
+LOGIN_ID=$2 # expected to be an AAD clientId, etc
+REGION=$3 # expected us-east-1, etc
+ROLE=$4 # ARN of the role to assume
+CONNECTOR_ID=$5 # connector id, such "outlook-mail", etc
+OUTPUT=$6 # Path where the file with the identity will be dropped
+
+
+export $(printf "AWS_ACCESS_KEY_ID=%s AWS_SECRET_ACCESS_KEY=%s AWS_SESSION_TOKEN=%s" \
+$(aws sts assume-role \
```
aperez-worklytics, line 16: To ensure that proper *assume_role* is used; have to do it because in my case I don't have perms with my user -but *infra admin* have it-
Tried to do it in the module with provider and assume role, but it is not working with a variable.

infra/modules/aws-cognito-identity-cli/main.tf
```diff
@@ -0,0 +1,27 @@
...
+}
+
+data "local_file" "identity-result" {
+  for_each = var.login-ids
+
+  filename   = "${path.module}/cognito_identity_${each.key}.json"
+  depends_on = [null_resource.identity-command]
+}
+
+output "identity_id" {
```
aperez-worklytics, line 19: So the approach here is the following:
- Invoking the shell as a resource; not as a data. This is to ensure that the script is executed as part of *apply* and not as part of *plan*; it will also ensure I can have a better handling of results. Once TF supports this, we could use the proper command.
- The output of the file is dropped as a file, needed to have the *IdentityId* from JSON response there. We don't need the file for more things and as it includes a *token* we could use for authentication it is marked as ignored in *.gitignore.

infra/modules/aws-cognito-pool/main.tf
```diff
@@ -0,0 +1,45 @@
+resource "aws_cognito_identity_pool" "main" {
```
aperez-worklytics, line 1: Creating a cognito identity pool; we will use it to create an identity and then use that identity to request a token from lambda side.

infra/modules/azuread-federated-credentials/variables.tf
```diff
@@ -0,0 +1,33 @@
...
+  type        = string
+  description = "The URL of the external identity provider, which must match the issuer claim of the external token being exchanged. The combination of the values of issuer and subject must be unique on the app."
+}
+
+variable "audience" {
+  type = string
+  # This is the recommended value from MSFT, as it is what AAD expects to be in the "aud"
+  # property token value
+  # See https://learn.microsoft.com/en-us/azure/active-directory/develop/workload-identity-federation-create-trust?pivots=identity-wif-apps-methods-azp#important-considerations-and-restrictions
+  default = "api://AzureADTokenExchange"
```
aperez-worklytics, line 27: GCP uses default; AWS needs to put here a value

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -134,7 +134,7 @@ locals {
         "Group.Read.All"
       ]
       environment_variables : {}
-      example_calls : [
+      example_api_calls : [
```
aperez-worklytics, line 137: Unified the property name as other connectors

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/WorkloadIdentityFederationGrantTokenRequestBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+ * claims "aud" and "iss" should be the same between two parts.
+ *
+ * @see <a href="https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow#third-case-access-token-request-with-a-federated-credential">...</a>
+ * as an example of MSFT implementation
+ *
+ * - <a href="https://datatracker.ietf.org/doc/html/rfc7521">...</a>
+ * - <a href="https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2">...</a>
+ */
+public abstract class WorkloadIdentityFederationGrantTokenRequestBuilder
+        implements OAuthRefreshTokenSourceAuthStrategy.TokenRequestBuilder, RequiresConfiguration {
```
aperez-worklytics, line 27: Now I see; worth to do some stuff and inherit from `ClientCredentialsGrantTokenRequestBuilder` to reuse properties and claims? Both are the same, except `ClientCredentialsGrantTokenRequestBuilder` builds the JWT with the private key provided meanwhile here it is responsible from implementation do obtain it somewhere

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AWSWorkloadIdentityFederationGrantTokenRequestBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+    public Set<ConfigService.ConfigProperty> getAllConfigProperties() {
+        return Streams.concat(super.getAllConfigProperties().stream(),
+                        Stream.of(ConfigProperty.values()))
+                .collect(Collectors.toSet());
+    }
+
+    @Override
+    @SneakyThrows
+    protected String getClientAssertion() {
+        GetOpenIdTokenForDeveloperIdentityResponse response = cognitoIdentityClient
```
aperez-worklytics, line 56: Getting a token from cognito; we provide `identityId` of the identity created from TF, its pool and the name of the login (*azure-access=client id of the AAD app*)

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AWSWorkloadIdentityFederationGrantTokenRequestBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Federation for AWS using a Cognito instance to have an IDENTITY_ID previously created
+ * identity to obtain its security JWT.
+ * The identity created should match login link DEVELOPER_NAME_ID=CLIENT_ID value created
+ */
+// NOTE: For some reason, this annotation is not working with @NoArgsConstructor(onConstructor_ = @Inject) syntax
+@NoArgsConstructor(onConstructor = @__(@Inject))
```
aperez-worklytics, line 26: Weird, works in this way but no in the other

java/impl/gcp/src/main/java/co/worklytics/psoxy/GCPWorkloadIdentityFederationGrantTokenRequestBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+             InputStreamReader inputStreamReader = new InputStreamReader(inputStream);
+             BufferedReader bufferedReader = new BufferedReader(inputStreamReader)) {
+
+            String inputLine;
+
+            while ((inputLine = bufferedReader.readLine()) != null)
+                content.append(inputLine);
+        }
+
+        return content.toString();
```
aperez-worklytics, line 74: GCP case is simpler than AWS; just invoke that URL which its response is the token to use

configs/azure-ad.yaml
eschultink: so this is actually from RFC for OAuth 2.0, and is a value to be passed with the oauth token request:
https://www.rfc-editor.org/rfc/rfc6749#section-4.4.2

Different values depend on what you're requesting an how.  In a sense, the "certificate" was the "client credential", hence this was the type. But it's the most common `grant_type` for most OAuth flows. 

see: https://oauth.net/2/grant-types/

infra/examples-dev/aws-msft-365/main.tf
eschultink: OK, add note in `CHANGELOG` that customers will need to `terraform init --upgrade` if coming from older version.

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AWSWorkloadIdentityFederationGrantTokenRequestBuilder.java
eschultink: could be my IDE being flaky

configs/azure-ad.yaml
eschultink: i need to review whole PR in more detail, but my thought is that it might make more sense to use a different value for `SOURCE_AUTH_STRATEGY_IDENTIFIER: oauth2_refresh_token` (that is our own invention)

infra/modular-examples/aws-msft-365/main.tf
```diff
@@ -74,34 +81,33 @@ module "msft-connection" {
...
+  description           = "AWS federation to be used for psoxy Connector - ${each.value.display_name}${var.connector_display_name_suffix}"
+  issuer                = "https://cognito-identity.amazonaws.com"
+  audience              = module.cognito-identity-pool.pool_id
+  subject               = module.cognito-identity.identity_id[each.key]
 
-  private_key_id = module.msft-connection-auth[each.key].private_key_id
-  private_key    = module.msft-connection-auth[each.key].private_key
-  ssm_path       = var.aws_ssm_param_root_path
-  kms_key_id     = var.aws_ssm_key_id
+  depends_on = [module.cognito-identity]
```
eschultink, line 106: i would hope TF smart enough that this explicit dependency isn't needed; while maybe `pool_id` known at the time of plan, I would think `identity_id` may not be known until that `cognito-identity` module is applied, right?

infra/modular-examples/aws-msft-365/main.tf
```diff
@@ -130,7 +136,7 @@ module "psoxy-msft-connector" {
   function_zip_hash               = module.psoxy-aws.deployment_package_hash
   api_caller_role_arn             = module.psoxy-aws.api_caller_role_arn
   aws_assume_role_arn             = var.aws_assume_role_arn
-  example_api_calls               = each.value.example_calls
+  example_api_calls               = each.value.example_api_calls
```
eschultink, line 139: 👍

infra/modular-examples/gcp-google-workspace/main.tf
```diff
@@ -250,6 +251,95 @@ module "worklytics-psoxy-connection-long-auth" {
 }
 # END LONG ACCESS AUTH CONNECTORS
 
+# BEGIN MSFT Connectors
```
eschultink, line 254: This can be dropped from this example. In practice, we don't have any one who wants to host in GCP and consume data from MSFT. Hence there's `aws` examples for `msft` and `google`; but only a `gcp` example for `google`.

YAGNI. Let's leave it out until we have a need, rather than maintain it.

The MSFT case we do have is a `infra/examples/msft-365`, which is example to create MSFT app registrations alone - without all the corresponding AWS infra. This is to support the case where the person managing the AWS infra lacks permissions to do the MSFT Azure AD stuff. That's the other example we may want to modify to support this auth scheme - but I think leave it for later PR.

infra/modular-examples/gcp-google-workspace/variables.tf
```diff
@@ -124,3 +124,9 @@ variable "google_workspace_example_admin" {
   description = "user to impersonate for Google Workspace API calls (null for value of `google_workspace_example_user`)"
   default     = null # will failover to user
 }
+
+variable "msft_tenant_id" {
+  type        = string
+  default     = ""
+  description = "ID of Microsoft tenant to connect to (req'd only if config includes MSFT connectors)"
+}
```
eschultink, line 132: drop

infra/modules/aws-cognito-identity-cli/addIdentity.sh
eschultink: 👍 it's the right way to do it I think

infra/modules/aws-cognito-identity-cli/addIdentity.sh
eschultink: are you cleaning up this output file somewhere? 

perhaps better to return the identities' json to STDOUT, and I think you can consume it in Terraform from that (rather than reading in a file that's presumed to exist). 

My worry is extraneous files being left around, ESPECIALLY if they're tokens.

infra/modules/aws-cognito-identity-cli/main.tf
```diff
@@ -0,0 +1,21 @@
...
+    command = "bash ${path.module}/addIdentity.sh ${var.identity_pool_id} ${each.value} ${var.aws_region} ${var.aws_role} ${each.key} ${path.module}"
+  }
+}
+
+data "local_file" "identity-result" {
+  for_each = var.login-ids
+
+  filename   = "${path.module}/cognito_identity_${each.key}.json"
+  depends_on = [null_resource.identity-command]
+}
```
eschultink, line 17: so yeah, my preference is a single `data` block that invokes the script, parses STDOUT as JSON and gets the value from there - rather than writing a file. I think 1) will be more concise/clear in terraform code, 2) will avoid extraneous file persisted to disk. I think it's similar to what we did in the certificate case previously

java/core/core.iml
```diff
@@ -10,6 +10,7 @@
       <sourceFolder url="file://$MODULE_DIR$/src/test/resources" type="java-test-resource" />
       <sourceFolder url="file://$MODULE_DIR$/target/generated-sources/annotations" isTestSource="false" generated="true" />
       <sourceFolder url="file://$MODULE_DIR$/target/generated-test-sources/test-annotations" isTestSource="true" generated="true" />
+      <excludeFolder url="file://$MODULE_DIR$/target" />
```
eschultink, line 13: ugh, thanks - I don't know why my IDE keeps taking this off

eschultink (approved):
some thoughts, no blockers.

configs/azure-ad.yaml
eschultink: after looking through java code, I'm OK with `workload_identity_federation`, although would do some research if there's any convention/standard on this.  Eg, when the grant is authenticated by this trusted-based federation approach - is it still a `client_credential`? or is there a different identifier used? I'd default to what you pass as `grant_type` in the token request, which unless I'm misunderstanding flow, I think is still `client_credentials`, right?

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -176,8 +176,10 @@ locals {
         "Group.Read.All",
         "User.Read.All"
       ],
-      environment_variables : {}
-      example_calls : [
+      environment_variables : {
+        REFRESH_ENDPOINT = "https://login.microsoftonline.com/${var.msft_tenant_id}/oauth2/v2.0/token"
```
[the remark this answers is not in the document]
aperez-worklytics: Ouch, I should update it in all of them - thanks!

configs/azure-ad.yaml
aperez-worklytics: It is `client_credentials`, here in MSFT docs are explained the uses cases. This is the [third one](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow#third-case-access-token-request-with-a-federated-credential) where the JWT assertion is created from a different identity provider rather than the proper application.

The thing is for GCP and MSFT it seems they refer to this auth flow as *workload identity federation* or *identity federation*; but I see nothing similar from AWS or nothing as standard in RFC; this is why not sure how to differentiate it from a "normal client credentials"

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AWSWorkloadIdentityFederationGrantTokenRequestBuilder.java
aperez-worklytics: Not at all, it happens just running maven outside of the IDE, but no idea why...

aperez-worklytics (commented):
Thanks, feedback applied!

eschultink (approved):

PR #246, opened 2023-01-27
more defensive header handling

eschultink:
### Fixes
 - avoid possibility that header names don't reach Java logic with expected casing

### Change implications

### DIFF
--- a/java/impl/gcp/src/main/java/co/worklytics/psoxy/CloudFunctionRequest.java
+++ b/java/impl/gcp/src/main/java/co/worklytics/psoxy/CloudFunctionRequest.java
@@ -7,7 +7,9 @@ import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


davidfq (approved):

java/impl/gcp/src/main/java/co/worklytics/psoxy/CloudFunctionRequest.java
```diff
@@ -39,14 +44,33 @@ public Optional<String> getQuery() {
...
     }
 
+    /**
+     * @return view of Headers with lower-case names
+     *
+     * this is kinda defensive; while AWS seems to pass headers in lower-case, GCP does not. (or
+     * at least docs, and prior practice, suggest they do not). But really we probably want to
+     * presume case-insensitivity, giving clients leeway in how they send headers.
+     */
+    private Map<String, List<String>> getCaseInsensitiveHeaders() {
```
davidfq, line 54: No worth having this duplicated method in a utils class or something?

eschultink: it's not exactly duplicated, bc for gcp its `Map<String, List<String>>`; for aws it's `Map<String, String>`

PR #249, opened 2023-02-09
S142 api gateway v1 support

eschultink:
### Features
 - API gateway v1 payload support
 - throw more meaningful exceptions if payloads lack expected fields
 - example of API gateway terraform

### Change implications

---
  - https://app.asana.com/0/0/1203913618176685


java/impl/aws/src/main/java/co/worklytics/psoxy/Handler.java
```diff
@@ -9,21 +9,20 @@
...
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.http.HttpHeaders;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.nio.charset.StandardCharsets;
-import java.util.Optional;
-import java.util.zip.GZIPOutputStream;
+import static co.worklytics.psoxy.ResponseCompressionHandler.isCompressionRequested;
```
aperez-worklytics, line 18: Just personal taste, I'd prefer to not import it and directly use `ResponseCompressionHandler.isCompressionRequested(...)` on L68

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV1ProxyEventRequestAdapter.java
```diff
@@ -0,0 +1,92 @@
...
+    @Override
+    public Optional<List<String>> getMultiValueHeader(String headerName) {
+        return getHeader(headerName.toLowerCase()).map(s -> Splitter.on(',').splitToList(s));
+    }
+
+    @Override
+    public String getHttpMethod() {
+        if (event.getRequestContext().getHttpMethod() == null) {
+            //q: better exception to throw here???
+            throw new IllegalStateException("Psoxy expects API Gateway V1 REST API proxy payload here. If using API Gateway V2 or Lambda function, please set the handler as co.worklytics.psoxy.Handler");
```
aperez-worklytics, line 52: I think the text is fine; however not sure if worth to return a 502 too (bad gateway? not sure if very accurate, as we might to represent that psoxy config is wrong)

aperez-worklytics (approved):

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV1ProxyEventRequestAdapter.java
eschultink: I think any uncaught exception (which is what this will end up as) will return a 502 if a lambda url, and maybe a 500 through API gateway - from my experience.

I think just relying on that behavior is fine for now; people need to grab the logs if they see any error case.

jlorper (approved):
OK, just one thing regarding the query string parameters

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV1ProxyEventRequestAdapter.java
```diff
@@ -0,0 +1,92 @@
...
+        return StringUtils.prependIfMissing(event.getPath(), "/");
+    }
+
+    @Override
+    public Optional<String> getQuery() {
+        if (event.getQueryStringParameters() == null || event.getQueryStringParameters().isEmpty()) {
+            return Optional.empty();
+        } else {
+            return Optional.of(event.getQueryStringParameters().entrySet().stream()
+                .map(e -> e.getKey() + "=" + e.getValue())
```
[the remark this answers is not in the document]
jlorper: Not sure why I had something very similar in of my scratches, so this should do the work:

```java
return Optional.of(
  Stream.concat(
    event.getQueryStringParameters().entrySet().stream().map(k -> Pair.of(k.getKey(), k.getValue())),
    event.getMultiValueQueryStringParameters().entrySet().stream().flatMap(k -> k.getValue().stream().map(v -> Pair.of(k.getKey(), v)))
)
.map( pair -> pair.getLeft() + "=" + pair.getRight())
.collect(Collectors.joining("&")));
```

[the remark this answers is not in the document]
eschultink: doesn't look null-safe to me ...

PR #259, opened 2023-02-15
fix msft 365 example

eschultink:
### Fixes
 - msft-365 example uses legacy local_cert module, which isn't preferred
 - local_cert module was removed in 0.4.11, but probably don't want to do that in case we break things for people. 

### Change implications

### DIFF
--- a/infra/modules/aws-cognito-identity-cli/variables.tf
+++ b/infra/modules/aws-cognito-identity-cli/variables.tf
@@ -13,6 +13,9 @@ variable "aws_role" {
   description = "Role to assume during script execution"
 }
 
-variable "login-ids" {
+# TODO: expect value format "${module.cognito-identity-pool.developer_provider_name}=${module.msft-connection[k].connector.application_id}"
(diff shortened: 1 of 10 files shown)


infra/modules/aws-cognito-identity-cli/variables.tf
```diff
@@ -13,6 +13,9 @@ variable "aws_role" {
   description = "Role to assume during script execution"
 }
 
-variable "login-ids" {
+# TODO: expect value format "${module.cognito-identity-pool.developer_provider_name}=${module.msft-connection[k].connector.application_id}"
+# --> add validation of that OR split those components in properties of an object??
+variable "login_ids" {
```
eschultink, line 18: @aperez-worklytics changed this to `_`, as I think better style to support (potential) invocation of terraform configurations via command line with variables, eg https://developer.hashicorp.com/terraform/language/values/variables#variables-on-the-command-line

jlorper (approved):
Looks good to me

PR #262, opened 2023-02-16
demo scaffolding

eschultink:
### Features
  - a standard location for creating demo configuration examples for customers; allows us to test a scenario for a specific customer and send them some valid code, while reducing change we inadvertently commit it

### Change implications

### DIFF
--- a/infra/.gitignore
+++ b/infra/.gitignore
@@ -4,3 +4,8 @@
 
 # PEM files used in local cert generation; don't want to commit them
 **.pem
+
+
+# standard demos directory, so we can create demo examples for customers without mistakenly in a
+# consistent location that won't be inadvertently committed to repo
+examples-demo/**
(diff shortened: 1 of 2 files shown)


infra/.gitignore
```diff
@@ -3,4 +3,9 @@
...
 
 # PEM files used in local cert generation; don't want to commit them
-**.pem
\ No newline at end of file
+**.pem
+
+
+# standard demos directory, so we can create demo examples for customers without mistakenly in a
+# consistent location that won't be inadvertently committed to repo
+examples-demo/**
```
aperez-worklytics, line 11: With this, even `README.md` included in this PR will be out of source control once this ignore is merged... not sure if something like the suggestion might work to only include that file:

```suggestion
examples-demo/**
!examples-demo/README.md
```

aperez-worklytics (approved):

infra/.gitignore
eschultink: i think it works bc already under version control, but yeah that's clearer

PR #264, opened 2023-02-18
build tool, to directly build bundle outside terraform module

eschultink:
### Features
  - shell scripts to directly build proxy deployment bundle from examples - eases debugging, should build under terraform fail (arguably may just add this to `init` as well, rather than using terraform for it at all in usual case)

### Change implications

### DIFF
--- a/tools/build.sh
+++ b/tools/build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+#see ../infra/modules/psoxy-package/build.sh
+# this is similar, but outputs errors directly instead of to log file; and doesn'
+
+IMPLEMENTATION=$1 # 'aws' or 'gcp'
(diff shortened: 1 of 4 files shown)


tools/build.sh
```diff
@@ -0,0 +1,25 @@
...
+# this is similar, but outputs errors directly instead of to log file; and doesn'
+
+IMPLEMENTATION=$1 # 'aws' or 'gcp'
+JAVA_SOURCE_ROOT=$2
+
+# set to fail on erors
+set -e
+
+
+#TODO: validate prereqs?? (mvn??)
```
aperez-worklytics, line 13: Not sure, maybe in such case just let `mvn` fails for that reason

aperez-worklytics (approved):

PR #261, opened 2023-02-16
S142 msft docs

aperez-worklytics:
Document about roles and process for running MSFT examples with AAD

### Features
[document min Azure perms/roles needed to run our msft terraform examples](https://app.asana.com/0/1201039336231823/1203982535369527)

### Change implications

 - dependencies added/changed? **no**

PR #265, opened 2023-02-25
TODOS via output

eschultink:
### Features
  - TODOs exposed as `output`, rather than *only* written to file system; improves flexibility to use these modules via Terraform Cloud / Enterprise / etc

### Change implications

---
  - https://app.asana.com/0/0/1203943782878949

### DIFF
--- a/infra/modular-examples/aws-msft-365/main.tf
+++ b/infra/modular-examples/aws-msft-365/main.tf
@@ -411,7 +411,13 @@ output "todos_1" {
   )
 }
 
-
+output "todos_2" {
+  description = "List of todo steps to complete 3rd."
+  value       = concat(
+    values(module.psoxy-msft-connector)[*].todo,
+    values(module.aws-psoxy-long-auth-connectors)[*].todo,
+  )
+}
 
 output "todos_3" {
   description = "List of todo steps to complete 3rd."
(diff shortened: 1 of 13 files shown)


infra/modules/azuread-grant-all-users/main.tf
```diff
@@ -44,6 +39,16 @@ ${join("\n", concat(var.app_roles, var.oauth2_permission_scopes))}
 EOT
 }
 
+resource "local_file" "todo" {
```
aperez-worklytics, line 42: In past I had this automated; it seems now can be done through [az cli](https://learn.microsoft.com/en-us/cli/azure/ad/app/permission?view=azure-cli-latest#az-ad-app-permission-admin-consent) and since https://github.com/Worklytics/psoxy/pull/261 we have roles required as part of documentation , so we could get rid of this *todo* and just have the command executed as part of terraform flow.
If worth it, please create a task and assign it to me!

infra/modular-examples/aws-msft-365/main.tf
```diff
@@ -400,4 +400,30 @@ output "cognito_identities" {
 
 output "cognito_identity_pool_id" {
   value = module.cognito-identity-pool.pool_id
-}
\ No newline at end of file
+}
+
+//q: better to invert control, let calling modules write the files??
```
aperez-worklytics, line 405: I think so; so the caller will decide how to proceed with the outputs (by default, writing them into files)

aperez-worklytics (approved):

PR #267, opened 2023-02-28
S143 api gateway paths

eschultink:
### Features
  - strip leading `stage`/`route` portions of paths for API Gateway v2 requests. Not sure exactly how this worked when we initially supported API Gateway v2, but believe we should now deal with payloads that come *either* from API Gateway V2 proxy routes *or* via Lambda URLs.

### Change implications

---
  - https://app.asana.com/0/0/1204068354727093

### DIFF
--- a/java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV2HTTPEventRequestAdapter.java
+++ b/java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV2HTTPEventRequestAdapter.java
@@ -32,10 +32,27 @@ public class APIGatewayV2HTTPEventRequestAdapter implements HttpEventRequest {
 
     @Override
     public String getPath() {
-        //remove stage portion from path
+        //remove stage portion from path, if any
+        //  - for Lambda URL deployments, we expect 'stage' to be '$default', and not match any portion
+        //    of rawPath; this will be no-op
+        //  - for API Gateway deployments, this should strip the stage
         String rawPath = event.getRawPath().replace("/" + event.getRequestContext().getStage(), "");
 
-        //TODO: route portion
+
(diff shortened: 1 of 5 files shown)


java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV2HTTPEventRequestAdapter.java
```diff
@@ -32,7 +32,30 @@ public class APIGatewayV2HTTPEventRequestAdapter implements HttpEventRequest {
...
+
+        // in API gateway deployments, multiple lambdas are behind single gateway with routes like
+        //  - ANY /gcal/{proxy+} --> psoxy-gcal lambda
+        //  - ANY /gdrive/{proxy+} --> psoxy-gcal-settings lambda
+        // eg ANY /gcal/{proxy+} -> /gcal/...
+
+        //NOTE: we're *assuming* routeKey of structure "ANY /gcal" for route "ANY /gcal/{proxy+}",
+        // but haven't found an example or documentation that explicitly states this
+
+        if (event.getRequestContext().getRouteKey() != null &&
```
aperez-worklytics, line 50: > Not sure exactly how this worked when we initially supported API Gateway v2

FYI https://github.com/Worklytics/psoxy/blob/113195b4b0d642d0e25488cc3e82c0f3aacfc60d/java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV2HTTPEventRequestAdapter.java#L27

aperez-worklytics (approved):

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/request/APIGatewayV2HTTPEventRequestAdapter.java
eschultink: 👍

PR #269, opened 2023-03-01
S143 use refresh endpoint envvar

aperez-worklytics:
After https://github.com/Worklytics/psoxy/pull/268, I've realized that `REFRESH_ENDPOINT` from MSFT connectors were coming from output and not from env var in specs. 

### Change implications

### DIFF
--- a/infra/modules/worklytics-connector-specs/main.tf
+++ b/infra/modules/worklytics-connector-specs/main.tf
@@ -133,7 +133,9 @@ locals {
         "User.Read.All",
         "Group.Read.All"
       ]
-      environment_variables : {}
+      environment_variables : {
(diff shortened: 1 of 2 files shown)


infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -133,7 +133,9 @@ locals {
         "User.Read.All",
         "Group.Read.All"
       ]
-      environment_variables : {}
+      environment_variables : {
+        REFRESH_ENDPOINT = "https://login.microsoftonline.com/${var.msft_tenant_id}/oauth2/v2.0/token"
```
eschultink, line 137: yeah, i think it's clearer this way since it's known here

eschultink (approved):

PR #273, opened 2023-03-07
Fix API Gateway example

davidfq:
.

### Fixes
- Amazon API Gateway example: 
  - Typo in route key
  - Map accessor syntax
  - Use `aws_iam_role_policy_attachment` instead of `aws_iam_policy_attachment` to avoid [exclusive attachments issue](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy_attachment)

### Features
.

### Change implications


eschultink (approved):
thanks!

PR #282, opened 2023-03-13
S144 Salesforce Psoxy support

aperez-worklytics:
- Added pure *client credentials* workflow, without certificate. See comments in the PR
- Added rules for Salesforce connector (only the ones we currently use)
- Updated AWS/GCP dev examples for supporting Salesfoce. Tested in both.

**NOTES**
- Most content of the PR are the json files used for tests
- NOT tested from app -connectors are not ready yet for proxy-.

### Features
- [Proxy: check how connector works](https://app.asana.com/0/0/1204138691954828)
- [Proxy: rules](https://app.asana.com/0/0/1204088124040621)

### Change implications


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsTokenRequestBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * implementation of an Access Token Request (per RFC6749 4.4.2) authenticated by an assertion (RFC 7521) based
+ * on client id + client secret
+ *
+ * see
+ *   - https://datatracker.ietf.org/doc/html/rfc7521
+ *   - https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2
+ */
+@NoArgsConstructor(onConstructor_ = @Inject)
+public class ClientCredentialsTokenRequestBuilder
```
aperez-worklytics, line 24: Added this and renamed `ClientCredentialsGrantTokenRequestBuilder` to `ClientCredentialsWithJWTAssertionGrantTokenRequestBuilder`; the idea is to support and differenciate two scenarios of same client credentials grant type: with client id + client secret to obtain a token and the one used for building a JWT assertion based on a public/private key.

ClientCredentialsWithJWTAssertionGrantTokenRequestBuilder is not currently used; it was the one used by MSFT connectors but since workload federation no one uses currently in current specs, but I think is interesting to keep it in anycase.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsWithJWTAssertionGrantTokenRequestBuilder.java
```diff
@@ -103,7 +104,7 @@ public HttpContent buildPayload() {
 
         Map<String, String> data = new TreeMap<>();
         //https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2
-        data.put(PARAM_GRANT_TYPE, getGrantType());
+        data.put(PARAM_GRANT_TYPE, "client_credentials");
```
aperez-worklytics, line 107: A bit dirty, at some point we need make a difference from specs for this. *grantType* will appear in specs to use this builder.

java/core/src/main/java/co/worklytics/psoxy/rules/salesforce/PrebuiltSanitizerRules.java
```diff
@@ -0,0 +1,150 @@
...
+                    .build())
+            .endpoint(Endpoint.builder()
+                    .pathRegex("^/services/data/v51.0/query[?]q=SELECT\\+Id\\+FROM\\+User.*$")
+                    .transforms(QUERY_ID_USER_TRANSFORMATION)
+                    .build())
+            .build();
+
+    public static final RuleSet SALESFORCE = Rules2.builder()
+            .endpoint(DESCRIBE)
+            .endpoint(UPDATED_ACCOUNTS_AND_ACTIVITY_HISTORY)
```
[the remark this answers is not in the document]
eschultink: as long as they don't have any rate limiting/throttling issues

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsWithJWTAssertionGrantTokenRequestBuilder.java
```diff
@@ -50,7 +51,7 @@ enum ConfigProperty implements ConfigService.ConfigProperty {
     // 'client_credentials' is MSFT
     //for Google, this is "urn:ietf:params:oauth:grant-type:jwt-bearer"
     @Getter(onMethod_ = @Override)
-    private final String grantType = "client_credentials";
+    private final String grantType = "client_credentials_jwt_assertion";
```
eschultink, line 54: This will break compatibility earlier terraform modules of this `0.4.x` generation, so we can't do it.

I think it would be better to parameterize the type of credential used, if needed, as a different config variable. (or duck-type it based on what sources of credentials are configured)

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsTokenRequestBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+
+        return new UrlEncodedContent(data);
+
+    }
+
+    @Override
+    public Set<ConfigService.ConfigProperty> getRequiredConfigProperties() {
+        return Set.of(
+            OAuthRefreshTokenSourceAuthStrategy.ConfigProperty.CLIENT_ID,
+            ConfigProperty.CLIENT_SECRET
```
eschultink, line 54: afiak, Zoom also uses a client_id + secret, so why isn't this that same flow? I guess as they're passing a different value for `grant_type` to server?

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsWithJWTAssertionGrantTokenRequestBuilder.java
eschultink: TOKEN_REQUEST_STRATEGY, or something? (where it defaults based on `GRANT_TYPE`, but if you pass explicit value you can override it?)

java/impl/aws/aws.iml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <module org.jetbrains.idea.maven.project.MavenProjectsManager.isMavenModule="true" type="JAVA_MODULE" version="4">
   <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_11">
-    <output url="file://$MODULE_DIR$/target/classes" />
-    <output-test url="file://$MODULE_DIR$/target/test-classes" />
+    <output url="file://$MODULE_DIR$/../target/classes" />
+    <output-test url="file://$MODULE_DIR$/../target/test-classes" />
```
eschultink, line 5: not sure what's right here, but adding magic `..` here seems a bit odd

eschultink (changes requested):

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -252,16 +252,66 @@ locals {
...
+      source_kind : "salesforce",
+      worklytics_connector_id : "salesforce-psoxy"
+      display_name : "Salesforce"
+      identifier_scope_id : "salesforce"
+      worklytics_connector_name : "Salesforce via Psoxy"
+      target_host : var.salesforce_domain
+      source_auth_strategy : "oauth2_refresh_token"
+      environment_variables : {
+        GRANT_TYPE : "client_credentials"
+        CREDENTIALS_FLOW : "client_secret"
```
aperez-worklytics, line 274: This is to define how client credentials will work (based on client secret or based on JWT assertion)

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestBuilder.java
```diff
@@ -78,12 +91,19 @@ enum ConfigProperty implements ConfigService.ConfigProperty {
 
     @Override
     public Set<ConfigService.ConfigProperty> getRequiredConfigProperties() {
-        return Set.of(
-            OAuthRefreshTokenSourceAuthStrategy.ConfigProperty.REFRESH_ENDPOINT, //should be redundant with strategy, but nonetheless req'd here too
-            OAuthRefreshTokenSourceAuthStrategy.ConfigProperty.CLIENT_ID,
-            ConfigProperty.PRIVATE_KEY_ID,
-            ConfigProperty.PRIVATE_KEY
-        );
+        if (getCredentialsType() == CredentialFlowType.CLIENT_SECRET) {
```
aperez-worklytics, line 94: Merged both client credentials in old class to ensure compatibility with old versions @eschultink

java/core/src/main/java/co/worklytics/psoxy/rules/salesforce/PrebuiltSanitizerRules.java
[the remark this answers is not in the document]
aperez-worklytics: We'll need to do more requests for that (in worst case, we could get 600k at once with update endpoint meanwhile with SOQL we'll need to page each 2k), always than the interval is not older than 30 days.

Another point is that this response will not be pseudonymized due limitation at 
https://github.com/Worklytics/psoxy/blob/main/java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerImpl.java#L382, explained in code comment. I can include support for that -in another PR to split the stuff- if interesting, but not sure if that could generate more issues than we try to solve here.

aperez-worklytics (commented):
Thanks -feedback applied

PR #283, opened 2023-03-13
build JAR from module package in examples, instead of repo clone

eschultink:
### Features
  - terraform clones entire GitHub repo; this provides ability to just use Java source from that, instead of depend separately on user having git clone with expected file structure

This potentially paves the way for using GitHub template repos for stand alone examples - rather than lumping everything together.

### Change implications

### DIFF
--- a/infra/examples/aws-google-workspace/build
+++ b/infra/examples/aws-google-workspace/build
@@ -3,4 +3,14 @@
 # NOTE: Terraform modules *should* build deployment bundle themselves, but in case that fails, this
 # script will do the equivalent with its output directly in your terminal
 
-../../../tools/build.sh aws ../../../java
+
+TF_CONFIG_ROOT=`pwd`
+PSOXY_BASE_DIR=${TF_CONFIG_ROOT}/.terraform/modules/psoxy
+
+# NOTE: the lines above point `PSOXY_BASE_DIR` at the clone Terraform makes of the psoxy git repo
+# if you're building from a local copy, uncomment the line below and set it to your local clone
+# PSOXY_BASE_DIR=../../../ # if this example configuration is within your clone
+# PSOXY_BASE_DIR=/path/to/your/local/psoxy/clone/ # if this example configuration is outside your clone
+
+
+${PSOXY_BASE_DIR}/tools/build.sh aws ${PSOXY_BASE_DIR}/java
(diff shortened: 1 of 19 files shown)


infra/examples/aws-google-workspace/build
```diff
@@ -3,4 +3,14 @@
 # NOTE: Terraform modules *should* build deployment bundle themselves, but in case that fails, this
 # script will do the equivalent with its output directly in your terminal
 
-../../../tools/build.sh aws ../../../java
+
+TF_CONFIG_ROOT=`pwd`
+PSOXY_BASE_DIR=${TF_CONFIG_ROOT}/.terraform/modules/psoxy
```
aperez-worklytics, line 8: Uhm -not for this PR-, in such case -where the code comes from module, so in theory it should not be locally modified- I think best option could be to just download the JAR from GH too and not compiling it in local

infra/modular-examples/aws-google-workspace/variables.tf
```diff
@@ -81,12 +81,17 @@ variable "connector_display_name_suffix" {
...
-  description = "the path where your psoxy repo resides. Preferably a full path, /home/user/repos/, avoid tilde (~) shortcut to $HOME"
+  description = "the path where your psoxy repo resides"
+  default     = null # will use `"${path.root}/.terraform/"` if not provided
 
   validation {
-    condition     = can(regex(".*\\/$", var.psoxy_base_dir))
+    condition     = var.psoxy_base_dir == null || can(regex(".*\\/$", var.psoxy_base_dir))
     error_message = "The psoxy_base_dir value should end with a slash."
   }
+  validation {
```
aperez-worklytics, line 91: Empty line between two validations? Same for other files:

```suggestion

  validation {
```

aperez-worklytics (approved):

infra/examples/aws-google-workspace/build
eschultink: distributing the JAR has some complexity:
  - analyze how all the dependencies are licensed.  If GPL, we'd need to, at minimum, list those with links to the source of each. (if customer builds JAR, we aren't distributing dependences - we're just listing them in the pom.xml, and customer is using their own `Maven` to obtain them)
  - trust factor that JAR is what we say it is (compilation of JAR from minimal Java code which customers can review plus linking to a list dependencies obtained from whatever Maven repository your trust, is more trustworthy than "here, take this JAR")
  - extra steps in our release process to build + publish the artifacts for each release

eschultink:
accidentally merged to `main`; reverted and put back on `rc-v0.4.14`

PR #285, opened 2023-03-14
expose deployment pkg references from modular examples

eschultink:
### Features
 - expose deployment pkg references from modular examples; facilitates customer configs that add some more lambdas on top of examples, without rebuilding JARs

### Change implications

### DIFF
--- a/infra/modular-examples/gcp-google-workspace/main.tf
+++ b/infra/modular-examples/gcp-google-workspace/main.tf
@@ -340,3 +340,15 @@ output "todos_3" {
     values(module.worklytics-psoxy-connection)[*].todo
   )
 }
+
+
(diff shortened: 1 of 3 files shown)


infra/modular-examples/gcp-google-workspace/main.tf
```diff
@@ -340,3 +340,15 @@ output "todos_3" {
     values(module.worklytics-psoxy-connection)[*].todo
   )
 }
+
+
+# use case: let someone consume this deploy another psoxy instance, reusing artifacts
```
aperez-worklytics, line 345: :+1: 
Probably something we could add by default: compile once and upload to the bucket in both platforms (AWS, GCP)

aperez-worklytics (approved):

PR #288, opened 2023-03-14
fix lookup tables

eschultink:
### Fixes
  - lookup table builders don't work, bc S3 doesn't allow multiple notifications configured on a bucket

--> well, turns out they ALSO don't allow a single notification to be configured to multiple lambdas with same event type, wtf

### Change implications

--> current blocker is:

psoxy_–_TODO_2_-_test_hris_md__infra_

I was attempted what Terraform docs show: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_notification#trigger-multiple-lambda-functions

--> but apparently the `filter_prefix`/`filter_suffix` part is needed, and they don't allow the thing to be the same???

---
  - https://app.asana.com/0/0/1204091733599946

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,27 @@ Working tracking of changes, updated as work done prior to release.  Please revi
         then wildcard policy to read shared also grants read of secrets across all connectors)
   - keys/salts per value kind (PII, item id, etc)
 
-## v0.4.11
+## [v0.4.14](https://github.com/Worklytics/psoxy/releases/tag/v0.4.14)
+
+Changes:
+  - you may see deletion and recreation of `aws_s3_bucket_notification` resources; this is expected
+    and will not alter behavior of deployment
+
+Fixes:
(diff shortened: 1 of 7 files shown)


eschultink:
ugh, ideas welcome. 

only workaround is what @jlorper was discussing, pushing this logic in the Java application layer where it writes same file to TWO outputs. This kinda works for the lookup table case, but isn't ideal for a subtly different case (JOIN with fields not sent to worklytics)

jlorper:
Another idea @eschultink - use [bucket replication](https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-how-setup.html)
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_replication_configuration

All input in HRIS bucket gets copied to the LOOKUP INPUT bucket (if you just want to avoid them configuring two exports from their HRIS source), then each bucket rules will be applied

PR #291, opened 2023-03-16
fix lookup tables

eschultink:
apologies for the massive refactor 😬 ; but generally we were using `Sanitizer` in bulk cases, only to get its "pseudonymize()" methods. So split out a `Pseudonymizer` interface/implementation, and made it much clear that the `Sanitizer` is really for REST APIs.

NOTE: take your time, don't think need to get this into `v0.4.14`, which I want out by EoW.

### Fixes
  - fixes the lookup table problem, but supporting additional transforms each outputting to their own bucket

Other ideas:
  - [x] `StorageHandler` is really `StorageObjectSanitizer`, to be clearer - as it it implements redaction, pseudonymization, etc. but arch differs from `RESTApiSanitizer`; `rules` are a method parameter to `StorageHandler` atm, but are a property of `RESTApiSanitizer` instances. Which is better approach? 
  - [x] ~~refactor `aws-psoxy-bulk` to use `aws-psoxy-output-bucket` module, as currently repeats a lot of that (at risk of having hierarchical TF modules)~~ _changed my mind this, bc will necessitate a lot of `moved` stuff in the modules for little value_

### Change implications


infra/modular-examples/aws-google-workspace/main.tf
```diff
@@ -325,40 +323,38 @@ module "psoxy-bulk-to-worklytics" {
   }, try(each.value.settings_to_provide, {}))
 }
 
-module "psoxy_lookup_tables_builders" {
+# BEGIN lookup builders
+module "lookup_output" {
```
eschultink, line 327: instead of a lambda for each lookup table, just an output bucket (with associated IAM policies, settings, etc).

infra/modular-examples/aws-google-workspace/main.tf
```diff
@@ -325,40 +323,38 @@ module "psoxy-bulk-to-worklytics" {
...
-      IS_DEVELOPMENT_MODE = contains(var.non_production_connectors, each.key)
-    }
-  )
 }
 
+locals {
+  inputs_to_build_lookups_for = toset(distinct([ for k, v in var.lookup_table_builders : v.input_connector_id ]))
+}
+
+resource "aws_ssm_parameter" "additional_transforms" {
```
eschultink, line 341: setting this SSM parameter for each "bulk" lambda to which we want to build an extra lookup table for - eg, but default lambda doesn't have any `ADDITIONAL_TRANSFORMS`; but if you provide this SSM param, then you change the behavior of existing lambda

java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
```diff
@@ -21,7 +21,7 @@
 @EqualsAndHashCode
 @JsonPropertyOrder({"allowAllEndpoints", "endpoints", "defaultScopeIdForSource"})
 @JsonInclude(JsonInclude.Include.NON_NULL) //NOTE: despite name, also affects YAML encoding
-public class Rules2 implements RuleSet, Serializable {
+public class Rules2 implements RESTRules {
```
eschultink, line 24: this is clearer; lets people bind to `RESTRules` interface instead of `Rules2` implementation

infra/modular-examples/aws-msft-365/main.tf
```diff
@@ -354,44 +354,37 @@ module "psoxy-bulk-to-worklytics" {
...
 }
 
-module "psoxy_lookup_tables_builders" {
+# BEGIN lookup builders
+module "lookup_output" {
   for_each = var.lookup_table_builders
 
-  source = "../../modules/aws-psoxy-bulk-existing"
-  # source = "git::https://github.com/worklytics/psoxy//infra/modules/aws-psoxy-bulk-existing?ref=v0.4.13"
+  source = "../../modules/aws-psoxy-output-bucket"
```
jlorper, line 361: shouldn't this point to source when deployed?

infra/modules/aws-psoxy-output-bucket/main.tf
```diff
@@ -0,0 +1,109 @@
+
+resource "aws_s3_bucket" "output" {
+  # note: this ends up with a long UTC time-stamp + random number appended to it to form the bucket name
+  bucket_prefix = "psoxy-${var.instance_id}-"
```
jlorper, line 4: could we use the customer id too (if available)? from out point of view all these buckets look similar.
In any case, with the new export connectors to come should not be an issue anymore

java/core/src/main/java/co/worklytics/psoxy/PseudonymizerImpl.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Override
+    public PseudonymizedIdentity pseudonymize(@NonNull String value) {
+        return pseudonymize((Object)  value);
+    }
+
+    @Override
+    public PseudonymizedIdentity pseudonymize(@NonNull Number value) {
+        return pseudonymize((Object) value);
+    }
```
jlorper, line 54: really needed in the interface? they just call the Object one. That could deal internally with specialized versions per object class.

java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
jlorper: Rules2 --> PsoxyRules

java/core/src/main/java/co/worklytics/psoxy/rules/RulesUtils.java
```diff
@@ -72,4 +77,19 @@ RuleSet parse(@NonNull String yamlString) {
...
+        CollectionType type = yamlMapper.getTypeFactory().constructCollectionType(ArrayList.class, StorageHandler.ObjectTransform.class);
+
+        if (additionalTransforms.isPresent()) {
+            try {
+                return yamlMapper.readValue(additionalTransforms.get(), type);
+            } catch (JsonProcessingException e) {
+                throw new RuntimeException("Failed to parse ADDITIONAL_TRANSFORMS from config", e);
+            }
+        } else {
+            return new ArrayList<>();
```
jlorper, line 91: Collections.emptyList() - immutable, guess you don't need it mutable

java/core/src/main/java/co/worklytics/psoxy/storage/BulkDataSanitizer.java
```diff
@@ -0,0 +1,29 @@
...
+     * @param reader The stream reader with the source content, including header unless schema
+     *               implicit in records
+     * @param bulkDataRules The rules to apply
+     * @param pseudonymizer The pseudonymizer to use
+     * @return A byte array with the processed content
+     * @throws IOException
+     */
+    byte[] sanitize(InputStreamReader reader,
+                    BulkDataRules bulkDataRules,
+                    Pseudonymizer pseudonymizer) throws IOException;
```
jlorper, line 28: I kind of liked the `Sanitizer` interface being (rules + pseudonymizer), but ok

java/impl/gcp/src/main/java/co/worklytics/psoxy/GCSFileEvent.java
```diff
@@ -1,75 +1,105 @@
...
-                            + sourceName + " and uploaded to " + storageEventResponse.getDestinationBucketName() + "/" + storageEventResponse.getDestinationObjectPath());
-                }
+                System.out.println("Writing to: " + storageEventResponse.getDestinationBucketName() + "/" + storageEventResponse.getDestinationObjectPath());
+
+                storage.createFrom(BlobInfo.newBuilder(BlobId.of(storageEventResponse.getDestinationBucketName(), storageEventResponse.getDestinationObjectPath()))
+                    .setContentType(blobInfo.getContentType())
+                    .build(), processedStream);
+
+                System.out.println("Successfully pseudonymized " + importBucket + "/"
+                    + sourceName + " and uploaded to " + storageEventResponse.getDestinationBucketName() + "/" + storageEventResponse.getDestinationObjectPath());
```
jlorper, line 97: `@Log` calls instead of sys out?

java/impl/aws/src/main/java/co/worklytics/psoxy/S3Handler.java
```diff
@@ -50,21 +62,37 @@ public String handleRequest(S3Event s3Event, Context context) {
...
+            StorageHandler.ObjectTransform.of(destinationBucket,  (CsvRules) defaultRules));
+
+        rulesUtils.parseAdditionalTransforms(configService)
+            .forEach(transforms::add);
+
+        for (StorageHandler.ObjectTransform transform : transforms) {
+            process(importBucket, sourceKey, transform.getDestinationBucketName(), transform.getRules());
+        }
+
+        return "Processed!";
```
jlorper, line 76: "200-OK", not sure where this is read back (probably nowhere) but if so, easier to process

jlorper (approved):
Few suggestions, no blockers. Liked the split of rules + pseudonymization :+1:

infra/modules/aws-psoxy-output-bucket/main.tf
eschultink: So this is the S3 bucket that we import customer's sanitized data from. we shouldn't be touching it directly; they should copy-paste the value into the "Add Connection" flow.

Security on this is that within our platform, only the specific customer's tenant SA can access it. So even if we did for some reason configure the connection on behalf of the customer, and mixed up buckets across customer, should be 403 from AWS IAM when the tenant SA tries to assume a role which it can't assume (based on the role's config in customer AWS)

java/core/src/main/java/co/worklytics/psoxy/PseudonymizerImpl.java
eschultink: yeah, maybe can be eliminated. The `Object` one was previously private

PR #292, opened 2023-03-16
shuffle bulk data

eschultink:
branches from #291 

### Features
  - shuffle records in bulk data cases, so original file is not a key to sanitized file

### Change implications

---
  - https://app.asana.com/0/0/1202303515554605


java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -151,37 +155,51 @@ public byte[] sanitize(@NonNull InputStreamReader reader,
...
+            Writer printWriter = new PrintWriter(baos);
             CSVPrinter printer = new CSVPrinter(printWriter, CSVFormat.DEFAULT
                 .withHeader(columnNamesForOutputFile.toArray(new String[0])))
             ) {
 
-            records.forEach(row -> {
-                Stream<Object> sanitized =
+            UnmodifiableIterator<List<CSVRecord>> chunks =
+                Iterators.partition(records.iterator(), rules.getRecordShuffleChunkSize());
+
```
jlorper, line 165: this previous step of splitting the rows into sublists is not really needed... in the end the whole process is held in memory (output = baos.toByteArray()) so could just do the shuffle on the whole file I think

java/gateway-core/src/main/java/com/avaulta/gateway/rules/ColumnarRules.java
```diff
@@ -35,13 +34,29 @@ public class ColumnarRules implements BulkDataRules {
...
+     * file acting as a lookup table for the output file
+     *
+     * (protip: setting to 1 should yield deterministic output for tests)
+     *
+     * TODO: fix serialization of default value
+     */
+    @JsonInclude(JsonInclude.Include.NON_EMPTY) // this doesn't work
+    @NonNull
+    @Builder.Default
+    protected Integer recordShuffleChunkSize = 500;
```
jlorper, line 54: yeah, not sure if this is really needed, but ok

jlorper (approved):
Given that everything is held in memory we could avoid the first step of splitting into sublists, but OK. Doesn't cause any harm either, it could be a little simpler

java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
eschultink: yeah, but I want to change that because it's stupid. So I don't want to *depend* on that behavior, as it blocks us from processing large files. (already broke asana import or something until they inc'd mem size in the lambda)

java/gateway-core/src/main/java/com/avaulta/gateway/rules/ColumnarRules.java
eschultink: i agree the number could have a default that comes from a config value somewhere, instead of being passed in via rules. and yeah, we could pick it dynamically or something based on row size to avoid mem problems

java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
jlorper: could just read line by line and write in buffers of random size.
// new list size of random(50, 500)
// do { process and add to the list } while (list is not full && more to process)
// shuffle and write

eschultink: yeah, i don't know that random buffer size really helps anonymity in theory; but I guess it does in practice, just because if I was trying to attack this, it's one more issue I'd have to consider when trying to re-identify the data (eg, that the corpus of potential matches gets larger the deeper I go into the data set).  but I think randomly shuffle sets of 500 is enough for me to attack something *other* than order first when trying to re-identify.

PR #293, opened 2023-03-17
AWS for all sources in single example

eschultink:
### Features
  - AWS example that includes ALL sources, just to reduce friction of having customers choose between 2 (and reduce maintaining 2 variants of very similar things)

### Change implications


infra/examples-dev/aws/main.tf
```diff
@@ -0,0 +1,94 @@
+terraform {
+  required_providers {
+    # for the infra that will host Psoxy instances
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 4.12"
+    }
+
+    # for the API connections to Google Workspace
```
jlorper, line 9: missing azure provider?

infra/examples-dev/aws/az-auth
```diff
@@ -0,0 +1 @@
+../../../tools/az-auth.sh
```
jlorper, line 1: is this a symlink??? looks like, the icon is different. Then use it in `build` too

infra/modular-examples/aws/main.tf
```diff
@@ -0,0 +1,553 @@
...
+
+  enabled_connectors             = var.enabled_connectors
+  google_workspace_example_user  = var.google_workspace_example_user
+  google_workspace_example_admin = try(coalesce(var.google_workspace_example_admin, var.google_workspace_example_user), null)
+  salesforce_domain              = var.salesforce_domain
+}
+
+locals {
+  msft_365_enabled = length(module.worklytics_connector_specs.enabled_msft_365_connectors) > 0
+}
```
jlorper, line 41: move this local to the MSFT section where it is used

tools/init-tfvars.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
+
+# fills a terraform vars file with values that are required for the Terraform configuration to work
```
jlorper, line 3: idea:
https://developer.hashicorp.com/terraform/language/values/variables#variable-definitions-tfvars-files

maybe use a node script that outputs the terraform vars in json
probably simpler to manage and don't really need to parse the terraform state/providers.

so it can read the `enabled_connectors` variable and generate the missing values if non-existent
so adding a new connector will just be enabling it and run this on top, it may complement existing config

infra/modular-examples/aws/main.tf
```diff
@@ -0,0 +1,553 @@
+terraform {
```
jlorper, line 1: I understand that this should be the only `main.tf` to exist for `aws` from now on

infra/examples-dev/aws-google-workspace/build
jlorper, line 1: aren't these folders - aws-google-workspace, aws-msft-365 deprecated?? now everything should be just `aws`, right?

jlorper (approved):
OK, some suggestions. Found the `init-tfvars.sh` script a bit hard to follow, probably easier to manage in JSON format.

infra/examples-dev/aws-google-workspace/build
eschultink: yeah, but people who set-up using current modular-examples will still `main.tf`s that are equivalent to these. so I prefer not to fully drop them at this point (we can't drop the `modular-example` versions, and I like have dev examples that use those)

PR #299, opened 2023-03-22
terraform improvements

eschultink:
### Features
 - expose variable to control JAR version at top level
 - bulk file test script for GCP

### Change implications

---
  - https://app.asana.com/0/0/1204191088895030
  - https://app.asana.com/0/0/1204210840874631
  - https://app.asana.com/0/0/1204227472646368


infra/modular-examples/aws-msft-365/variables.tf
```diff
@@ -76,6 +82,12 @@ variable "psoxy_base_dir" {
   }
 }
 
+variable "bundle_version" {
```
jlorper, line 85: couldn't this just be read from a file in the root?

eschultink: yeah, probably

infra/modules/psoxy-package/main.tf
```diff
@@ -39,5 +31,9 @@ output "path_to_deployment_jar" {
 }
 
 output "filename" {
-  value = local.filename
+  value = data.external.deployment_package.result.filename
+}
+
+output "version" {
```
jlorper, line 37: I think `version` can be removed, as far as I can tell is not used and value comes embedded in filename.

jlorper (approved):
OK, I think `version` is redundant as output and could be removed, the less outputs to propagate the better. But no blocker obviously.

jlorper:
Good improvement by the way!

infra/modules/psoxy-package/main.tf
eschultink: i think it's good to have the structured data there if we ever need it,

PR #301, opened 2023-03-23
support re-writing bulk object path prefixes when processing

eschultink:
### Features
 -  support `INPUT_BASE_PATH`/`OUTPUT_BASE_PATH` config params. by default, key of output object is the same as input object. `INPUT_BASE_PATH`, if provided, will be stripped from the key of the output object; and then `OUTPUT_BASE_PATH`, if provided, will be prepended to the key. 

This allows something like follows:

- input: `s3://my-bucket/in/foo.csv`
- output: `s3://my-bucket/out/foo.csv`

by configuring `INPUT_BASE_PATH='in/'` and `OUTPUT_BASE_PATH='out/'` 

### Change implications

---
  - https://app.asana.com/0/0/1204251780882977

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/BulkModeConfigProperty.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/BulkModeConfigProperty.java
@@ -7,7 +7,26 @@ package co.worklytics.psoxy.gateway;
 public enum BulkModeConfigProperty implements ConfigService.ConfigProperty {
 
     OUTPUT_BUCKET,
+
+    /**
+     * if provided, this path segment will be prepended to keys of output objects
+     */
     ADDITIONAL_TRANSFORMS,
+
+    /**
+     * if provided, this path segment will be removed from keys of input object to produce
+     * keys of corresponding output objects.
+     *
+     * NOTE: objects with keys that do not start with this segment will still be processed; this is
+     * not any kind of filter.
+     */
+    INPUT_BASE_PATH,
+
+    /**
+     * if provided, this path segment will be prepended to keys of output objects, after removing
+     * any {@link #INPUT_BASE_PATH} segment
+     */
+    OUTPUT_BASE_PATH,
     ;
 
 }
--- a/infra/modules/psoxy-package/build.sh
+++ b/infra/modules/psoxy-package/build.sh
@@ -23,6 +23,8 @@ if [ ! -f $PATH_TO_DEPLOYMENT_JAR ] || [ ! -z "$FORCE_BUILD" ] ; then
 
   ln -sf ${LOG_FILE} ${TERRAFORM_CONFIG_PATH}/last-build.log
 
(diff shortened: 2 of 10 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/BulkModeConfigProperty.java
```diff
@@ -7,7 +7,26 @@
...
     ADDITIONAL_TRANSFORMS,
+
+    /**
+     * if provided, this path segment will be removed from keys of input object to produce
+     * keys of corresponding output objects.
+     *
+     * NOTE: objects with keys that do not start with this segment will still be processed; this is
+     * not any kind of filter.
+     */
+    INPUT_BASE_PATH,
```
aperez-worklytics, line 23: Maybe as `INPUT_PREFIX` could be a better name? Using path it seems like the complete path, but both input and output are going to be the prefix to use between bucket name and object relative path

java/core/src/main/java/co/worklytics/psoxy/gateway/BulkModeConfigProperty.java
```diff
@@ -7,7 +7,26 @@
...
+
+    /**
+     * if provided, this path segment will be prepended to keys of output objects, after removing
+     * any {@link #INPUT_BASE_PATH} segment
```
aperez-worklytics, line 27: Bit confused with this comment; following the example in the test:
input_base_path = "in"
output_base_path = "out"
input_bucket = "bucket-in", 
output_bucket = "bucket-out"
object_path = "directory/file.csv"

Path finally used will be *bucket-in/in/directory/file.csv* due `input_base_path` variable; then output will be *bucket-out/out/directory/file.csv*; so object path is  `output_base_path` + `object_path`; no matter what's the value for `INPUT_BASE_PATH` here right? Maybe I'm wrong but I don't see that output transformation is using `INPUT_BASE_PATH` to replace it from path, it is just concatenating  `OUTPUT_BASE_PATH` to the output. 

If that's the case I'd drop that line in the comment.

aperez-worklytics (approved):
Approved; but please take a look on my comments

java/core/src/main/java/co/worklytics/psoxy/gateway/BulkModeConfigProperty.java
eschultink: No, `input_base_path` isn't used to *construct* a path from which to read the input object; we already have that object path (from the event that triggered the lambda)

`input_base_path` is only used to *construct* the path to which to write the output object.  

The point is to make possible: `/in/file.csv` --> `/out/file.csv`

If we only supported `PUT


With the values you put

> input_base_path = "in"
output_base_path = "out"
input_bucket = "bucket-in",
output_bucket = "bucket-out"
object_path = "directory/file.csv"

the output will be written to `s3://bucket-out/out/directory/file.csv`, having been read from `s3://bucket-in/directory/file.csv`

PR #303, opened 2023-03-24
use object metadata to avoid loop

eschultink:
### Features
  - write metadata on object, to help us detect that it was previously processed
  - consume object metadata and, if object was written by proxy, skip it so customers can't create loops as easily (currently, if they used a default notification rule to trigger lambdas, but put an ADDITIONAL_TRANSFORM back to the original bucket, it would loop infinitely)

### Change implications

---
  - https://app.asana.com/0/0/1204260893379996
  - https://app.asana.com/0/0/1204260893379998

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
@@ -110,13 +110,19 @@ public class StorageHandler {
     }
 
     /**
+     * @param sourceBucket the bucket from which the original object was read
+     * @param sourceKey the key of the original object within the source bucket
+     * @param transform the transform that was applied to the object
+     *
      * @return metadata to be written to the transformed object
      */
-    public Map<String, String> getObjectMeta(String importBucket, String sourceKey, ObjectTransform transform) {
+    public Map<String, String> buildObjectMetadata(String sourceBucket, String sourceKey, ObjectTransform transform) {
+        //transform currently unused; in future we probably want to record what transform was
+        // applied, to aid traceability of pipelines
         return Map.of(
             BulkMetaData.INSTANCE_ID.getMetaDataKey(), hostEnvironment.getInstanceId(),
             BulkMetaData.VERSION.getMetaDataKey(), config.getConfigPropertyAsOptional(ProxyConfigProperty.BUNDLE_FILENAME).orElse("unknown"),
-            BulkMetaData.ORIGINAL_OBJECT_KEY.getMetaDataKey(), importBucket + "/" + sourceKey
+            BulkMetaData.ORIGINAL_OBJECT_KEY.getMetaDataKey(), sourceBucket + "/" + sourceKey
         );
     }
 
--- a/java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
@@ -14,7 +14,6 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 
 /**
  * solves a DaggerMissingBinding exception in tests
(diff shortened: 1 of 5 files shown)


java/core/src/test/java/co/worklytics/psoxy/storage/StorageHandlerTest.java
```diff
@@ -98,4 +99,10 @@ void buildRequest(String inputBasePath, String outputBasePath, String expectedOu
         assertEquals("bucket", request.getDestinationBucketName());
         assertEquals(expectedOutputPath, request.getDestinationObjectPath());
     }
+
+    @Test
+    public void metadata() {
+        handler.getObjectMeta("bucket", "directory/file.csv", handler.buildDefaultTransform());
```
jlorper, line 105: what's the test? missing assertions

java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
```diff
@@ -85,6 +111,35 @@ ObjectTransform buildDefaultTransform() {
...
+            BulkMetaData.INSTANCE_ID.getMetaDataKey(), hostEnvironment.getInstanceId(),
+            BulkMetaData.VERSION.getMetaDataKey(), config.getConfigPropertyAsOptional(ProxyConfigProperty.BUNDLE_FILENAME).orElse("unknown"),
+            BulkMetaData.ORIGINAL_OBJECT_KEY.getMetaDataKey(), importBucket + "/" + sourceKey
+        );
+    }
+
+    /**
+     * @return true if the object has already been sanitized by this proxy instance
+     */
+    public boolean hasBeenSanitized(Map<String, String> objectMeta) {
```
jlorper, line 128: probably I'd name this as `isCreatedByProxy` rather than coupling with function... could not be sanitized if the column config is wrong.

jlorper (commented):
A couple of things, but looks ok

java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
eschultink: it also shouldn't have metadata if column config was wrong ...

java/core/src/test/java/co/worklytics/psoxy/storage/StorageHandlerTest.java
eschultink: yeah, added something. but kinda lame.

what I really want to test is that metadata works round trip - eg, we can set it, then get it back through AWS as expected so `hasBeenSanitized` call is correct.

java/core/src/test/java/co/worklytics/psoxy/storage/StorageHandlerTest.java
```diff
@@ -98,4 +100,21 @@ void buildRequest(String inputBasePath, String outputBasePath, String expectedOu
         assertEquals("bucket", request.getDestinationBucketName());
         assertEquals(expectedOutputPath, request.getDestinationObjectPath());
     }
+
+    @Test
+    public void metadata() {
```
aperez-worklytics, line 105:
```suggestion
    public void getObjectMeta() {
```

or

```suggestion
    public void getObjectMeta_should_match_metadataKey() {
```

java/impl/aws/src/main/java/co/worklytics/psoxy/S3Handler.java
```diff
@@ -60,6 +60,13 @@ StorageEventResponse process(String importBucket, String sourceKey, StorageHandl
         StorageEventResponse storageEventResponse;
         S3Object s3Object = s3Client.getObject(new GetObjectRequest(importBucket, sourceKey));
 
+        if (storageHandler.hasBeenSanitized(s3Object.getObjectMetadata().getUserMetadata())) {
```
aperez-worklytics, line 63: Can s3Object.getObjectMetadata() returns null ?

aperez-worklytics (approved):

java/impl/aws/src/main/java/co/worklytics/psoxy/S3Handler.java
eschultink: reasonable to be defensive and check for it; nothing in AWS's doc of the interface specifies that it can't be

PR #305, opened 2023-03-27
fix java8 build

eschultink:
### Fixes
 - java8 build

### Features
 - ci to cover java8 build

### Change implications

 - dependencies added/changed? **o**

PR #309, opened 2023-03-29
S145 java8 unified workflow

aperez-worklytics:
**NOTE:** Parent branch is https://github.com/Worklytics/psoxy/pull/305

From https://github.com/Worklytics/psoxy/pull/305#discussion_r1150322155

Adding a unified workflow to compile java with parameters we might require in both java 8 and java 17 CI

### DIFF
--- a/.github/workflows/ci-java8.yaml
+++ b/.github/workflows/ci-java8.yaml
@@ -9,29 +9,7 @@ on:
       - '**' # should match all branches
 
 jobs:
-  ci_java8:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Check out code
-        uses: actions/checkout@v2
-      - name: Setup Java
-        uses: actions/setup-java@v1
-        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-java8.yaml
```diff
@@ -9,29 +9,7 @@ on:
...
-        run: |
-          mvn clean compile -P dev -T 2C -Dversions.logOutput=false -DprocessDependencies=false -DprocessDependencyManagement=false
-      - name: Test
-        working-directory: java/
-        run: |
-          mvn post-clean test -T 2C -Dversions.logOutput=false -DprocessDependencies=false -DprocessDependencyManagement=false -Dsurefire.forkCount=2.5C -Dsurefire.reuseForks=true
+  ci_java_8:
+    uses: ./.github/workflows/build-java.yaml
+    with:
+      compile-profile: dev
```
aperez-worklytics, line 15: Failing because this profile doesn't exist, but once it is it should work

PR #312, opened 2023-03-30
Store shared token only if newer than what's in config

jlorper:
### Fixes
- Tries to minimize race conditions when storing the shared token.
NOTE: Test pending

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
@@ -11,11 +11,13 @@ import com.google.auth.Credentials;
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.OAuth2CredentialsWithRefresh;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.util.concurrent.Uninterruptibles;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.java.Log;
+import org.apache.commons.lang3.RandomUtils;
 import org.apache.commons.lang3.StringUtils;
 
 import javax.inject.Inject;
@@ -25,6 +27,7 @@ import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.*;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -67,6 +70,8 @@ public class OAuthRefreshTokenSourceAuthStrategy implements SourceAuthStrategy {
      */
     public static final Duration MIN_DURATION_TO_KEEP_REFRESH_TOKEN = Duration.ofDays(7);
 
+    // lock to use to synchronize on token refreshes within same JVM
+    private static final Object $refreshLock = new Object();
 
     //q: should we put these as config properties? creates potential for inconsistent configs
     // eg, orphaned config properties for SourceAuthStrategy not in use; missing config properties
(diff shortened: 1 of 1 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -205,9 +210,11 @@ public AccessToken refreshAccessToken() throws IOException {
             }
 
             if (shouldRefresh(this.currentToken, clock.instant())) {
+                // some jitter to minimize multiple instances refreshing at same time
+                Uninterruptibles.sleepUninterruptibly(50 + RandomUtils.nextLong(50, 200), TimeUnit.MILLISECONDS);
```
eschultink, line 214: there's already a `RandomNumberGenerator` being injected above - L182

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -348,17 +355,41 @@ Optional<AccessToken> getSharedAccessTokenIfSupported() {
...
-                        objectMapper.writerFor(AccessTokenDto.class)
-                            .writeValueAsString(AccessTokenDto.toAccessTokenDto(accessToken)));
-                    log.log(Level.INFO, "New token stored in config");
-                } catch (JsonProcessingException e) {
-                    log.log(Level.SEVERE, "Could not serialize token into JSON", e);
+                // at least synchronize in the same JVM
+                synchronized ($refreshLock) {
+                    try {
+                        // wait slightly before checking for most recent token in config
+                        Uninterruptibles.sleepUninterruptibly(RandomUtils.nextLong(50, 200), TimeUnit.MILLISECONDS);
```
eschultink, line 365: again, already a `RandomNumberGenerator` being injected above; use that, can fake it to 1 for tests or something

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -205,9 +210,11 @@ public AccessToken refreshAccessToken() throws IOException {
             }
 
             if (shouldRefresh(this.currentToken, clock.instant())) {
+                // some jitter to minimize multiple instances refreshing at same time
+                Uninterruptibles.sleepUninterruptibly(50 + RandomUtils.nextLong(50, 200), TimeUnit.MILLISECONDS);
                 tokenResponse = exchangeRefreshTokenForAccessToken();
-                this.currentToken = asAccessToken(tokenResponse);
-                storeSharedAccessTokenIfSupported(this.currentToken);
+                AccessToken newToken = asAccessToken(tokenResponse);
+                this.currentToken = storeSharedAccessTokenIfSupported(newToken);
```
eschultink, line 217: more readable to just have `this.currentToken = newToken`, then `storedSharedAccessTokenIfSupported(newToken)` as separate line

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -205,9 +210,11 @@ public AccessToken refreshAccessToken() throws IOException {
             }
```
eschultink, line 210: possibly everything above here should be moved, as it's not actually refreshing the token.

eschultink: and actually, this is `refreshAccessTokenIfNeeded` or something - as it won't *necessarily* refresh.

eschultink (commented):
nowhere quick to do it, but I think `currentToken` on this class is in effect acting to "cache" the token until it expires. in shared case, we might want to be periodically reloading the token from parameter store in case updated - regardless of TTL of token itself

jlorper:
Gonna discard, after a thorough review the jitter in the refresh token threshold is more than enough. We might do the lock to minimize writes on initial burst, but pretty negligible. (no 401 errors seen at all, besides initial ones)

PR #314, opened 2023-03-30
fix zoom token update

eschultink:
### Fixes
 - zoom token update fails bc no type (in practice, no one *should* be impacted if using our default TF examples, as they all create initial value with a type; and updating existing parameter doesn't seem to fail if no type)

### Change implications

### DIFF
--- a/java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
+++ b/java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
@@ -56,6 +56,9 @@ public class ParameterStoreConfigService implements ConfigService {
         String key = parameterName(property);
         try {
             PutParameterRequest parameterRequest = PutParameterRequest.builder()
+                //TODO: determine proper 'type' here based on ConfigProperty attribute?
+                // (as of 2023-03, we don't update any params that *aren't* secrets, so not really issue)
+                .type(ParameterType.SECURE_STRING) // in case parameter doesn't exist yet, SSM rejects PUT w/o type
                 .name(key)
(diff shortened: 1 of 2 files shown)


java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
```diff
@@ -56,6 +56,9 @@ public void putConfigProperty(ConfigProperty property, String value) {
         String key = parameterName(property);
         try {
             PutParameterRequest parameterRequest = PutParameterRequest.builder()
+                //TODO: determine proper 'type' here based on ConfigProperty attribute?
+                // (as of 2023-03, we don't update any params that *aren't* secrets, so not really issue)
+                .type(ParameterType.SECURE_STRING) // in case parameter doesn't exist yet, SSM rejects PUT w/o type
```
aperez-worklytics, line 61: So options are STRING, STRING_LIST, SECURE_STRING and UNKNOWN. I'd say that always SECURE_STRING, otherwise parameter could be as env variable

aperez-worklytics (approved):

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
eschultink: yeah, it's a reasonable assumption, although @jlorper and I were discussing yesterday that we think this can be used to implement locking (as a `PutParameterRequest` with `overwrite=false` should success if your request is first, fail otherwise - so we could implement `acquireLock` over SSM with that); in which case payload needn't be secure

PR #315, opened 2023-03-31
improve json schema filters

eschultink:
### Features
 - clarify the our `JsonSchema` is really a filter
 - ignore 'required' which doesn't make sense for filter semantics
 - fix schema evaluation more clearly, moving it up to root

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
@@ -58,6 +58,8 @@ public class RESTApiSanitizerImpl implements RESTApiSanitizer {
     List<Pair<Pattern, Endpoint>> compiledEndpointRules;
     Map<Transform, List<JsonPath>> compiledTransforms = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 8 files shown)


aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -89,6 +92,15 @@ Map<Endpoint, Pattern> getCompiledAllowedEndpoints() {
         return compiledAllowedEndpoints;
     }
 
+    SchemaRuleUtils.JsonSchemaFilter getRootDefinitions() {
+        if (rootDefinitions == null) {
+            synchronized ($writeLock) {
+                rootDefinitions = SchemaRuleUtils.JsonSchemaFilter.builder().definitions(rules.getDefinitions()).build();
+            }
```
jlorper, line 99: for doing the double locking singleton properly you need to check again in the lock

```java
synchronized ($writeLock) {
  if (rootDefinitions == null) {
     rootDefinitions = SchemaRuleUtils.JsonSchemaFilter.builder().definitions(rules.getDefinitions()).build();
  }
}
```

PR #317, opened 2023-03-31
JSON filter 2

eschultink:
### Features
 - omit `type` field from schema filters
 - omit `type: array`/`type: object` where implicit
 - omit `additionalProperties`; ignored anyways

(general goal here is to improve human readability while also shrinking size to keep RULES under SSM size limits)

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
@@ -95,7 +95,9 @@ public class RESTApiSanitizerImpl implements RESTApiSanitizer {
     JsonSchemaFilterUtils.JsonSchemaFilter getRootDefinitions() {
         if (rootDefinitions == null) {
             synchronized ($writeLock) {
-                rootDefinitions = JsonSchemaFilterUtils.JsonSchemaFilter.builder().definitions(rules.getDefinitions()).build();
+  if (rootDefinitions == null) {
+          rootDefinitions = JsonSchemaFilterUtils.JsonSchemaFilter.builder().definitions(rules.getDefinitions()).build();
+  }
             }
         }
         return rootDefinitions;
--- a/java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
@@ -3,11 +3,11 @@ package co.worklytics.psoxy.impl;
 import co.worklytics.psoxy.*;
 import co.worklytics.psoxy.rules.RESTRules;
 import com.avaulta.gateway.rules.Endpoint;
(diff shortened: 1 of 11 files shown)


java/gateway-core/src/test/java/com/avaulta/gateway/rules/JsonSchemaFilterUtilsTest.java
```diff
@@ -193,21 +189,41 @@ void filterBySchema_refs() {
...
+            objectMapper.writerWithDefaultPrettyPrinter().writeValueAsString(compactCopy));
+
+        // 30% reduction in yaml size
+        assertEquals(404, yamlMapper.writeValueAsString(schemaWithRefs).length());
+        assertEquals(286, yamlMapper.writeValueAsString(compactCopy).length());
+
+        // non-pretty JSON is actually more compact than the yaml (pretty JSON is a lot less compact)
+        // (so if going to give up on human readability by base64gzip'ing the schema, might as well
+        //  do that with the JSON)
+        // eg, https://developer.hashicorp.com/terraform/language/functions/base64gzip
```
jlorper, line 216: well, if we go this way, probably there's no point in doing all this compacting code. Just upload the full schema compressed, which make sense.
We might have special lambda calls to output rules, env variables back so can easy check config despite being compressed

jlorper (approved):
If we do the base64gzip then I feel all of this is irrelevant

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -92,10 +92,10 @@ Map<Endpoint, Pattern> getCompiledAllowedEndpoints() {
         return compiledAllowedEndpoints;
     }
 
-    SchemaRuleUtils.JsonSchemaFilter getRootDefinitions() {
+    JsonSchemaFilterUtils.JsonSchemaFilter getRootDefinitions() {
         if (rootDefinitions == null) {
             synchronized ($writeLock) {
-                rootDefinitions = SchemaRuleUtils.JsonSchemaFilter.builder().definitions(rules.getDefinitions()).build();
+                rootDefinitions = JsonSchemaFilterUtils.JsonSchemaFilter.builder().definitions(rules.getDefinitions()).build();
```
jlorper, line 98: for doing the double locking singleton properly you need to check again in the lock

```suggestion
  if (rootDefinitions == null) {
          rootDefinitions = JsonSchemaFilterUtils.JsonSchemaFilter.builder().definitions(rules.getDefinitions()).build();
  }
```

java/gateway-core/src/test/java/com/avaulta/gateway/rules/JsonSchemaFilterUtilsTest.java
eschultink: yeah, i prefer to keep human readable if we can, so people can just see the value in their SSM parameters to have confidence in it.  It's possible base64gzip would be conditional just for the extreme cases (eg, if > 8KB compacted it then gzip)

PR #326, opened 2023-04-05
build cleanup

eschultink:
### Fixes
 - disable java20 build, as doesn't seem to work
 - nicer name on the main java build

### Change implications

### DIFF
--- a/.github/workflows/ci-java-all.yaml
+++ b/.github/workflows/ci-java-all.yaml
@@ -24,8 +24,9 @@ jobs:
 
   # NOTE: java 19 support ended 21 Mar 2023; it's known to be incompatible with our code
 
-  # Java 20 - released 21 Mar 2023, supported until 19 Sept 2023
-  ci_java20:
-    uses: ./.github/workflows/build-java.yaml
-    with:
-      java-version: '20'
+  # Java 20 - released 21 Mar 2023, supported until 19 Sept 2023;
+  #  as of 5 Apr 2023, current version of java 20 provided by GitHub action is ALSO incompatible with our code
+#  ci_java20:
+#    uses: ./.github/workflows/build-java.yaml
+#    with:
+#      java-version: '20'
--- a/.github/workflows/ci-java.yaml
+++ b/.github/workflows/ci-java.yaml
@@ -1,3 +1,5 @@
+name: "CI - Java"
+
 # CI for our project as a GitHub action
 # see https://help.github.com/en/actions/language-and-framework-guides/building-and-testing-java-with-maven
 on:


.github/workflows/ci-java8.yaml
```diff
@@ -12,7 +12,8 @@ on:
       - '**' # should match all branches
 
 jobs:
-  ci_java_8:
+  ci_java8:
+    id: ci_java8
     env:
       compile-profile: '-P java8 ' # NOTE: trailing space is important
```
aperez-worklytics, line 18: I'd drop the trailing space at the end and change the command with a better format:
```suggestion
      compile-profile: '-P java8'
```

and then (same for rest of commands)

```
mvn package install ${{ env.compile-profile }} -T
```

.github/workflows/ci-java8.yaml
```diff
@@ -12,7 +12,8 @@ on:
       - '**' # should match all branches
 
 jobs:
-  ci_java_8:
+  ci_java8:
+    id: ci_java8
```
aperez-worklytics, line 16: I'd change the name to `ci_core_java8`, as this is to generate the package used in the other side and not a a *regular* CI process

.github/workflows/ci-java8.yaml
aperez-worklytics: Also, not sure if cache makes effect here. In https://github.com/Worklytics/psoxy/pull/326/files#diff-70c5462ec1ddc37f61c43b9bc892afc9fb4a8a96de2f1a559dcc3a6a0e09b624L29 we cache all packages and then we remove them in https://github.com/Worklytics/psoxy/pull/326/files#diff-70c5462ec1ddc37f61c43b9bc892afc9fb4a8a96de2f1a559dcc3a6a0e09b624L35

aperez-worklytics (approved):

.github/workflows/ci-java8.yaml
eschultink: eg, when compile-profile is blank, then `mvn package install ${{ env.compile-profile }} -T` evaluates to `mvn package install  -T` and I get that error

jlorper: hmm, -T is for thread count, if you are not specifying anything why use it at all?
Also order AFAIK is not important, you could rewrite this to
`mvn -T 2C package install ${{ env.compile-profile }}` - at least us some parallelization 2C = 2 times number of cores

.github/workflows/ci-terraform-example.yaml
```diff
@@ -0,0 +1,31 @@
...
+        type: string
+        description: Java version to use # see https://github.com/actions/setup-java#supported-version-syntax
+        default: 17
+
+jobs:
+  validate_tf_example:
+    name: 'validate tf examples'
+    runs-on: ubuntu-latest
+    env:
+      TERRAFORM_VERSION: 1.3.3
```
jlorper, line 17: update to latest in minor 1.3.9 ?

jlorper (approved):

PR #327, opened 2023-04-07
dev/alpha example of GCP for all data sources

eschultink:
### Features
  - example of using GCP host connecting to all data sources
  - improvements to init stuff

### Change implications


infra/examples-dev/gcp/README.md
```diff
@@ -0,0 +1,73 @@
+# example-google-workspace
+** alpha quality example atm **
+
+A Terraform root module to provision GCP project for Psoxy, configure it, and create necessary infra
+to for connections to all supported Google Workspace sources, with state stored to local filesystem.
```
jlorper, line 5:
```suggestion
for connections to all supported Google Workspace sources, with state stored to local filesystem.
```

infra/examples-dev/gcp/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Usage
+
+### Prereqs
+
+You'll need:
+- a Bash-like shell environment on Linux, MacOS, or [WSL on Windows](https://learn.microsoft.com/en-us/windows/wsl/install).
+- [`git` installed](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git), although it
+  is usually included on those platforms (check with `git --version`).
+- an AWS account and credentials, as described in [Psoxy's AWS - Getting Started docs](https://github.com/Worklytics/psoxy/blob/v0.4.17/docs/aws/getting-started.md)
```
jlorper, line 19: point links to `main` so they go to latest version always?
```suggestion
- an AWS account and credentials, as described in [Psoxy's AWS - Getting Started docs](https://github.com/Worklytics/psoxy/blob/main/docs/aws/getting-started.md)
```

jlorper: why AWS is needed at all? Isn't this the GCP example?

jlorper: Also, the GCP project is not created, it should exist, right? Shouldn't it be listed as a pre-req too?

infra/examples-dev/gcp/variables.tf
```diff
@@ -0,0 +1,162 @@
...
+    "us-central1",
+    "us-west1",
+  ]
+}
+
+variable "enabled_connectors" {
+  type        = list(string)
+  description = "list of ids of connectors to enabled; see modules/worklytics-connector-specs"
+
+  default = [
```
jlorper, line 74: should this be empty by default, to force the customer to fill it in?

jlorper (approved):
A couple of suggestions

infra/examples-dev/gcp/README.md
eschultink: well, imho it's like an interface - if we move stuff around the link will break

re: AWS, yes should be changed to gcp

infra/examples-dev/gcp/variables.tf
eschultink: yeah, that's fair; we should force explicit choice at least.

PR #329, opened 2023-04-11
support for custom_rest_rules, path prefixes in GCP

eschultink:
### Features
 - `custom_rest_rules` support in GCP
 - option to impersonate a GCP SA when deploying, rather than be auth'd directly
 - support to pass `environment_id` / `config_parameter_path`, to aid in deploying multiple Psoxy configurations within same GCP project (mainly dev use-case)

### Change implications

---
  - https://app.asana.com/0/0/1204332867116146
  - https://app.asana.com/0/0/1204364771334903
  - https://app.asana.com/0/0/1203715901381860

### DIFF
--- a/infra/examples-dev/gcp-google-workspace/main.tf
+++ b/infra/examples-dev/gcp-google-workspace/main.tf
@@ -22,6 +22,10 @@ terraform {
   #  }
 }
 
+provider "google" {
+  impersonate_service_account = var.gcp_terraform_sa_account_email
+}
+
 module "psoxy" {
   source = "../../modular-examples/gcp-google-workspace"
   # source = "git::https://github.com/worklytics/psoxy//infra/modular-examples/aws-google-workspace?ref=v0.4.18"
--- a/infra/modules/gcp-secret-fill-md/main.tf
+++ b/infra/modules/gcp-secret-fill-md/main.tf
@@ -7,26 +7,35 @@ variable "secret_id" {
   type        = string
   description = "id of the secret"
 }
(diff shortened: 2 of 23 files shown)


infra/examples-dev/gcp-google-workspace/main.tf
```diff
@@ -22,6 +22,10 @@ terraform {
   #  }
 }
 
+provider "google" {
+  impersonate_service_account = var.gcp_terraform_sa_account_email
```
aperez-worklytics, line 26: Is this working with a variable? If so cool, because other times it was doing nothing here

infra/modules/gcp-secret-fill-md/main.tf
```diff
@@ -7,26 +7,35 @@ variable "secret_id" {
   type        = string
   description = "id of the secret"
 }
+variable "path_prefix" {
```
aperez-worklytics, line 10:
```suggestion

variable "path_prefix" {
```

aperez-worklytics (approved):

infra/examples-dev/gcp-google-workspace/main.tf
eschultink: by default, it's `null` and does nothing; if you pass the variable, it does the impersonation. (AWS case is similar). So yeah useful to have this flexibility in the example. While not a lot of people use this type of approach in practice, imho it's better from security perspective (as users not themselves privileged; just SAs are which can be assumed)

PR #332, opened 2023-04-14
gcp lookup table

eschultink:
### Features
 - GCP lookup table

### Change implications

 - dependencies added/changed? **y no**


java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -82,7 +97,7 @@ private String parameterName(ConfigProperty property) {
         if (StringUtils.isBlank(this.namespace)) {
             return property.name();
         } else {
-            return String.join("_", this.namespace, property.name());
+            return this.namespace + property.name();
```
aperez-worklytics, line 100: I'd prefer that the namespace doesn't contains `_` as it is used as separator

aperez-worklytics (approved):

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
eschultink: that's the default for now for backwards compatibility; this gives flexibility, as you can override with whatever you want. If you pass explicit value via ENV VARS, that value is used directly as `namespace`; if you don't pass anything, by default will now be `PSOXY_` (instead of `PSOXY` and `_` getting appended). This allows people to use a path-like structure `PSOXY/` if they want.

PR #342, opened 2023-04-25
cleaner endpoint rules

eschultink:
### Features
 - use `pathTemplate` instead of `pathRegex` to make cleaner rules where possible

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
@@ -345,9 +345,11 @@ public class RESTApiSanitizerImpl implements RESTApiSanitizer {
     Map<Endpoint, Pattern> getCompiledAllowedEndpoints() {
         if (compiledAllowedEndpoints == null) {
             synchronized ($writeLock) {
+            if (compiledAllowedEndpoints == null) {
                 compiledAllowedEndpoints = rules.getEndpoints().stream()
                     .collect(Collectors.toMap(Function.identity(),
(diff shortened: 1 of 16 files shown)


java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -75,57 +76,28 @@ public RESTApiSanitizerImpl(@Assisted RESTRules rules,
...
-                .filter(entry -> entry.getValue().matcher(relativeUrl).matches())
+                .filter(entry -> entry.getKey().getPathRegex() != null && entry.getValue().matcher(relativeUrl).matches()
+                    || (entry.getKey().getPathTemplate() != null && entry.getValue().matcher(url.getPath()).matches() && allowedQueryParams(entry.getKey(), URLUtils.queryParamNames(url))))
                 .filter(entry -> entry.getKey().getAllowedMethods()
                     .map(methods -> methods.stream().map(String::toUpperCase).collect(Collectors.toList())
                             .contains(httpMethod.toUpperCase()))
                     .orElse(true))
-                .filter(entry -> entry.getKey().getAllowedQueryParamsOptional()
-                    .map(allowedParams -> allowedParams.containsAll(URLUtils.queryParamNames(url))).orElse(true))
+                .filter(entry -> allowedQueryParams(entry.getKey(), URLUtils.queryParamNames(url))) // redundant in the pathTemplate case
```
jlorper, line 100: why don't you move it out from the L95 if it's going to be done here either way?

jlorper (approved):
OK, but be aware than path templates are less restrictive with regard of the query string.
Before, methods that didn't accept one blocked the rule, the path template allows it  now.

PR #357, opened 2023-05-09
more AWS caller types, generalize testing support

eschultink:
generally, a bunch of test examples were coupled to our IAM approach, eg AWS CLI auth'd as AWS user, which would assume a *super* admin role in the target AWS account when deploying/testing.

### Fixes
 - our test commands for bulk cases presumed Terraform running as a 'super admin' role that could read/write to all buckets
 - our test commands for REST cases defaulted to assumption of Terraform role, which again needed to be the 'super admin' to be able to 'invoke' the lambdas; while this worked for many customers in practice, these examples won't work in more rigorous IAM policy setups. (eg, there's no reason why the role that's applying the terraform configuration should be able to invoke lambdas)
 - initialization of `terraform.tfvars` worked in way that's only appropriate for the 'super admin' role approach

### Features
  - option to provisioning TestingPolicy and attach it to a testing role to better support bulk case (as `PsoxyCaller` role will never be able to test uploads)

### Change implications

### DIFF
--- a/infra/examples-dev/aws-google-workspace/variables.tf
+++ b/infra/examples-dev/aws-google-workspace/variables.tf
@@ -69,9 +69,9 @@ variable "caller_aws_arns" {
 
   validation {
     condition = alltrue([
-      for i in var.caller_aws_arns : (length(regexall("^arn:aws:iam::\\d{12}:\\w+$", i)) > 0)
+      for arn in var.caller_aws_arns : (length(regexall("^arn:aws:iam::\\d{12}:((role|user)\\/)?\\w+$", arn)) > 0)
     ])
-    error_message = "The values of caller_aws_arns should be AWS Resource Names, something like 'arn:aws:iam::914358739851:root'."
+    error_message = "The values of caller_aws_arns should be AWS Resource Names for principals; something like 'arn:aws:iam::914358739851:root'."
   }
 }
 
--- a/infra/examples-dev/aws-google-workspace/variables.tf
+++ b/infra/examples-dev/aws-google-workspace/variables.tf
@@ -49,7 +49,7 @@ variable "force_bundle" {
   default     = false
 }
 
-variable "enable_deployment_testing" {
+variable "provision_testing_infra" {
   type        = bool
   description = "whether to provision infra needed to support testing of deployment"
   default     = false
--- a/infra/examples-dev/aws-google-workspace/variables.tf
+++ b/infra/examples-dev/aws-google-workspace/variables.tf
@@ -49,7 +49,7 @@ variable "force_bundle" {
   default     = false
 }
 
-variable "enable_testing" {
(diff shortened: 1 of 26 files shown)


infra/examples-dev/aws-google-workspace/variables.tf
```diff
@@ -49,6 +49,12 @@ variable "force_bundle" {
   default     = false
 }
 
+variable "enable_testing" {
```
aperez-worklytics, line 52: Maybe I'd rename it as `support_deployment_testing` or something like that, `enable_testing` seems too generic

aperez-worklytics (approved):

infra/examples-dev/aws-msft-365/variables.tf
```diff
@@ -51,7 +51,7 @@ variable "caller_aws_arns" {
 
   validation {
     condition = alltrue([
-      for i in var.caller_aws_arns : (length(regexall("^arn:aws:iam::\\d{12}:\\w+$", i)) > 0)
+      for i in var.caller_aws_arns : (length(regexall("^arn:aws:iam::\\d{12}:((role|user)\\/)?\\w+$", i)) > 0)
```
jlorper, line 54:
```suggestion
      for arn in var.caller_aws_arns : (length(regexall("^arn:aws:iam::\\d{12}:((role|user)\\/)?\\w+$", arn)) > 0)
```
just to be consistent with previous variable declaration (appears in more places)

infra/examples-dev/aws/variables.tf
```diff
@@ -121,6 +121,12 @@ variable "force_bundle" {
   default     = false
 }
 
+variable "enable_deployment_testing" {
```
jlorper, line 124:
```suggestion
variable "provision_testing_infrastructure" {
```
I find the name a bit confusing:
An `enable_deployment_testing = false` looks like you don't want to do testing, rather than infra is not needed.

jlorper (approved):
A couple of suggestions, mainly the variable name that I find confusing

infra/modules/aws-psoxy-rest/main.tf
```diff
@@ -87,13 +84,19 @@ Review the deployed function in AWS console:
...
+   - Node.js (>=16) and npm (v >=8) installed.
+   - install the tool itself (in the location from which you plan to run the test commands, if it's
+     not the same location where you originally ran the Terraform apply)
 
 ```shell
 ${local.command_npm_install}
 ```
+   - ensure the location you're running from is authenticated as an AWS principal which can assume
+     the role `${var.api_caller_role_arn}` ( `aws sts get-caller-identity` to determine who you're
+     authenticated as; if necessary, as this ARN to the `caller_aws_arns` list in the
```
davidfq, line 98: "...add this ARN.."

PR #362, opened 2023-05-23
fix GCP test

eschultink:
### Fixes
 - args (added in 0.4.19) that don't seem to be needed (commit - https://github.com/Worklytics/psoxy/pull/332/commits/4898a235dce6cc6b50568c1e0d4ab23844bff970 - no idea why)

### Change implications


eschultink:
and we didn't see this, bc the command in bash scripts that are generated `eg, `./test-hris.sh` don't include the `-p`/`-r` args

davidfq (approved):

PR #365, opened 2023-05-25
S148 jira support

aperez-worklytics:
Adding support for Jira Server and Jira Cloud for Psoxy:
- New rules for proxy
- Supporting sharing token with `refresh_token` kind
- Fixing GCP secrets for sharing tokens
- Updated TF examples-dev and modules to pass jira spec parameters

NOTES: 
- Most part of this PR are due tests and JSON files used there
- A 2nd PR will be created to populate `jira.md` documentation -once this is approved-

See comments on the review for further info

### Features
- [Proxy: 3LO support](https://app.asana.com/0/0/1204504995636807)
- [Proxy: rules](https://app.asana.com/0/0/1204504995636808)

### Change implications


infra/modules/gcp-oauth-secrets-role/main.tf
```diff
@@ -0,0 +1,14 @@
+resource "google_project_iam_custom_role" "secret_manager_updater" {
```
aperez-worklytics, line 1: For Jira cloud it is required to have a shared access token. Refresh token can be used once, if a new refresh token is served the prev one will be invalidated. For this, it is needed to build a shared access token with a lock thing.
See the comment on `SecretManager` for GCP for more details, but basically this is the custom role with custom permissions to update/list secrets; otherwise only an admin could do it.

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -504,6 +504,151 @@ all the operations for the connector:
...
+    jira-server = {
+      source_kind : "jira-server"
+      worklytics_connector_id : "jira-server-psoxy"
+      target_host : var.jira_server_url
+      source_auth_strategy : "oauth2_access_token"
+      display_name : "Jira REST API"
+      identifier_scope_id : "jira"
+      worklytics_connector_name : "Jira Server REST API via Psoxy"
+      secured_variables : [
+        { name : "ACCESS_TOKEN", writable : false },
```
aperez-worklytics, line 518: Jira Server works with a personal access token; it could work with [certificate](https://developer.atlassian.com/server/jira/platform/oauth/#step-3--authorize) or [user credentials](https://developer.atlassian.com/server/jira/platform/basic-authentication/) but it is not an OAuth2.0 compliant.
For this, I've followed the access token, as it seems less hard to setup and better than providing user/name to the requests.

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -504,6 +504,151 @@ all the operations for the connector:
...
+      source_auth_strategy : "oauth2_refresh_token"
+      display_name : "Jira REST API"
+      identifier_scope_id : "jira"
+      worklytics_connector_name : "Jira REST API via Psoxy"
+      secured_variables : [
+        { name : "ACCESS_TOKEN", writable : true },
+        { name : "REFRESH_TOKEN", writable : true },
+        { name : "CLIENT_ID", writable : false },
+        { name : "CLIENT_SECRET", writable : false },
+        { name : "OAUTH_REFRESH_TOKEN", writable : true }, # Lock for sharing access token, handled by Psoxy
```
aperez-worklytics, line 550: And Jira Cloud uses the access token + refresh + token; as refresh_token is invalidated once a new one is requested, access_token needs to be shared between instances. For that, `OAUTH_REFRESH_TOKEN` is a variable used as a lock stuff

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/RefreshTokenTokenRequestBuilder.java
```diff
@@ -35,8 +35,21 @@
     public enum ConfigProperty implements ConfigService.ConfigProperty {
         REFRESH_TOKEN, //NOTE: you should configure this as a secret in Secret Manager
         CLIENT_SECRET, //NOTE: you should configure this as a secret in Secret Manager
+        USE_SHARED_TOKEN,
     }
 
+    @Override
+    public boolean useSharedToken() {
+        Optional<String> useSharedTokenConfig = config.getConfigPropertyAsOptional(ConfigProperty.USE_SHARED_TOKEN);
```
aperez-worklytics, line 43: Adding support for shared token with `refresh_token` grant_type; used in jira cloud.

java/core/src/main/java/co/worklytics/psoxy/rules/jira/PrebuiltSanitizerRules.java
```diff
@@ -0,0 +1,589 @@
...
+            .endpoint(ISSUE_COMMENT_V2)
+            .endpoint(ISSUE_COMMENT_V3)
+            .endpoint(ISSUE_WORKLOG_V2)
+            .endpoint(ISSUE_WORKLOG_V3)
+            .endpoint(USERS)
+            .endpoint(PROJECTS)
+            .build();
+
+    @VisibleForTesting
+    static final RESTRules JIRA_SERVER = Rules2.builder()
```
aperez-worklytics, line 307: Two group of rules: one for cloud and another one for server. Server doesn't have same version of api than cloud (cloud is 3 in beta, server in 2) so:
- Adding rules for cloud for version 2 and 3 when required; main difference is with [comments](https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#version) where in 3 we could extract information about mentions, etc. But in 2, comment appears *as-is*, so it should be redacted. In 3, is pseudonymized.
- Groups, Users, Projects and Changelogs are only present for Cloud, not for Server.

For that, there are test files for both versions (2,3 and server/cloud) and two test classes to cover them.

java/core/src/test/java/co/worklytics/psoxy/rules/jira/JiraCloudTests.java
```diff
@@ -0,0 +1,251 @@
...
+    void sha() {
+        this.assertSha("7869e465607b7a00b4bd75a832a9ed1f811ce7f2");
+    }
+
+    @SneakyThrows
+    @Test
+    @Override
+    @Disabled
+    public void yamlLength() {
+        // Do nothing, as response schema is bigger than we allow for advanced parameters
```
aperez-worklytics, line 41: So yeah, rules are quite long and exceed the basic yaml length. I've put the test to do nothing, because rules are working once deployed and I don't have more ideas to reduce the size of the rules. The main part is the schema response metadata for searching issues, which is quite huge...

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -110,54 +108,39 @@ public Optional<String> getConfigPropertyAsOptional(ConfigProperty property) {
...
-        long secretTtl = Math.max(expires.getSeconds(), LOCK_SECRETS_TTL.getSeconds());
+        final SecretName lockSecretName = getLockSecret(lockId);
 
         try (SecretManagerServiceClient client = SecretManagerServiceClient.create()) {
             Secret lockSecret;
             try {
                 lockSecret = client.getSecret(lockSecretName);
             } catch (com.google.api.gax.rpc.NotFoundException e) {
-                //create it
-                Secret initial = Secret.newBuilder()
```
aperez-worklytics, line 135: That wasn't working due:
1) Lack of perms, as to create a new secret we need to have admin permission on the whole project
2) Creating/dropping a secret is a cost operation for GCP; labels seems its free

infra/modules/worklytics-connector-specs/main.tf
eschultink: certainly better than username + password, but is it tied to a single user?

infra/examples-dev/aws-google-workspace/variables.tf
```diff
@@ -267,6 +267,24 @@ variable "salesforce_domain" {
   default     = ""
 }
 
+variable "jira_server_url" {
+  type        = string
+  default     = ""
```
eschultink, line 272: I think it's clearer to be `null` to imply unset, rather than `""`

infra/examples-dev/aws-google-workspace/variables.tf
```diff
@@ -267,6 +267,24 @@ variable "salesforce_domain" {
...
+  description = "URL of the Jira server (only required if using Jira Server connector)."
+}
+
+variable "jira_cloud_id" {
+  type        = string
+  default     = ""
+  description = "Cloud id of the Jira Cloud to connect to (only required if using Jira connector)."
```
eschultink, line 279: validation or example of what value should be here?

infra/modules/gcp-oauth-secrets-role/main.tf
eschultink: also, this isn't oauth in particular; so should name the module differently. It's maybe `gcp-instance-managed-secret` or something?

infra/modules/gcp-oauth-secrets-role/main.tf
```diff
@@ -0,0 +1,14 @@
...
+  project     = var.project_id
+  role_id     = "SecretManagerAccess"
```
eschultink, line 3: role id probably needs env/namespace component, to avoid collisions across deployments. (some customers mix dev+prod stuff in same host account/project, distinguishing with env prefixes)

infra/modules/gcp-oauth-secrets-role/main.tf
```diff
@@ -0,0 +1,14 @@
...
+  description = "Access for updating and reading secrets because updating secrets are only available as roles/secretmanager.admin"
+
+  permissions = [
+    "resourcemanager.projects.get",
+    "secretmanager.secrets.get",
+    "secretmanager.secrets.getIamPolicy",
+    "secretmanager.secrets.list",
+    "secretmanager.secrets.update"
+  ]
+}
```
eschultink, line 14: possibly this role should just move to the `gcp` module. maybe clearer.

infra/modules/gcp-oauth-secrets-role/main.tf
eschultink: maybe call it `PsoxyInstanceSecretManager` or something?

infra/modules/gcp-oauth-secrets-role/main.tf
```diff
@@ -0,0 +1,14 @@
+resource "google_project_iam_custom_role" "secret_manager_updater" {
+  project     = var.project_id
+  role_id     = "SecretManagerAccess"
+  title       = "Access for updating and reading secrets"
```
eschultink, line 4: `Psoxy Instance Secret Manager`

infra/modules/gcp-oauth-secrets-role/main.tf
```diff
@@ -0,0 +1,14 @@
...
+  description = "Access for updating and reading secrets because updating secrets are only available as roles/secretmanager.admin"
```
eschultink, line 5:
```suggestion
  description = "Role to grant on secret that is to be managed by a Psoxy instance (cloud function); subset of roles/secretmanager.admin, to support reading/updating the secret"
```

infra/modules/gcp-oauth-secrets-role/variables.tf
```diff
@@ -0,0 +1,4 @@
+variable "project_id" {
+  type        = string
+  description = "name of the gcp project"
```
eschultink, line 3: ID
```suggestion
  description = "alphanumeric ID of gcp project"
```

infra/modules/worklytics-connector-specs/main.tf
eschultink: I'm very confused. Why isn't this the same as `REFRESH_TOKEN`? imho, `REFRESH_TOKEN` should be the canonical oauth one. If JIRA  cloud has some additional token that doesn't work like canonical oauth,  we should give it some more specific name

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -504,6 +504,151 @@ all the operations for the connector:
...
+]
+```
+
+Use that id as `jira_cloud_id` parameter to include as part of Terraform deployment. That will target your instance for REST API requests.
+8. Finally set following variables in AWS System Manager parameters store / GCP Cloud Secrets (if default implementation):
+  - `PSOXY_JIRA_CLOUD_ACCESS_TOKEN` secret variable with value of `access_token` received in previous response
+  - `PSOXY_JIRA_CLOUD_REFRESH_TOKEN` secret variable with value of `refresh_token` received in previous response
+  - `PSOXY_JIRA_CLOUD_CLIENT_ID` with `Client Id` value.
+  - `PSOXY_JIRA_CLOUD_CLIENT_SECRET` with `Client Secret` value.
+
```
eschultink, line 651: please also put version of this in `docs/sources/atlassian/jira.md` or something

docs/sources/example-rules/jira/jira-cloud.yaml
eschultink, line 1: should we have a directory level for `atlassian` instead of `jira`? Confluence is the other data source we'd be likely to support here, although could argue that it's really also part of JIRA

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -504,6 +504,151 @@ all the operations for the connector:
...
+        { name : "ACCESS_TOKEN", writable : true },
+        { name : "REFRESH_TOKEN", writable : true },
+        { name : "CLIENT_ID", writable : false },
+        { name : "CLIENT_SECRET", writable : false },
+        { name : "OAUTH_REFRESH_TOKEN", writable : true }, # Lock for sharing access token, handled by Psoxy
+      ],
+      environment_variables : {
+        GRANT_TYPE : "refresh_token"
+        REFRESH_ENDPOINT : "https://auth.atlassian.com/oauth/token"
+        USE_SHARED_TOKEN : "TRUE"
```
eschultink, line 555: why isn't this being set for `zoom` as well?

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
eschultink: OK, sure - where's it creating the secret in Terraform? I would except it in `gcp-oauth-secrets` or something, but I don't see it there.

infra/modules/gcp-oauth-secrets-role/main.tf
eschultink: or `PsoxyInstanceSecretLocker`? 

labels just being implementation detail of how we lock/unlock a secret.

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -110,54 +108,39 @@ public Optional<String> getConfigPropertyAsOptional(ConfigProperty property) {
...
-                    //ttl here, bc this Secret not managed by Terraform; so want it to clean itself
-                    // up after some time
-                    .setTtl(Duration.newBuilder().setSeconds(secretTtl).build())
-                    .build();
-                lockSecret = client.createSecret(lockSecretName.getProject(), lockSecretName.getSecret(),
-                    initial);
+                // It should not happen, as the secret has been created by Terraform
+                throw new RuntimeException(e);
```
eschultink, line 123: error message here should provide more detail on the problem; eg, secret named X must exist, should have been created by terraform.

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -110,54 +108,39 @@ public Optional<String> getConfigPropertyAsOptional(ConfigProperty property) {
...
-    }
-
     @Override
     public boolean acquire(@NonNull String lockId, @NonNull java.time.Duration expires) {
         Preconditions.checkArgument(StringUtils.isNotBlank(lockId), "lockId cannot be blank");
 
-        final SecretName lockSecretName = SecretName.of(projectId, this.namespace + lockId);
-
-        long secretTtl = Math.max(expires.getSeconds(), LOCK_SECRETS_TTL.getSeconds());
+        final SecretName lockSecretName = getLockSecret(lockId);
```
eschultink, line 115: effect of this is that it's not a generic locking service; but it's actually acquiring a lock on an existing secret. Not a problem atm, but that doesn't extend to case where customer might want there secrets actually in Vault or something. (even if customer wants secrets in vault, it's perfectly reasonable to still implement locking on top of GCP Secret Manager - as lock doesn't contain sensitive info)

eschultink (changes requested):

infra/modular-examples/gcp-google-workspace/main.tf
```diff
@@ -159,6 +162,15 @@ module "worklytics-psoxy-connection" {
...
   long_access_parameters = { for entry in module.worklytics_connector_specs.enabled_oauth_secrets_to_create : "${entry.connector_name}.${entry.secret_name}" => entry }
+  env_vars_for_locker = distinct(flatten([
+  for k, v in module.worklytics_connector_specs.enabled_oauth_long_access_connectors : [
+  for env_var in v.environment_variables : {
+    connector_name = k
+    env_var_name    = "OAUTH_REFRESH_TOKEN"
+  } if try(v.environment_variables.USE_SHARED_TOKEN, null) != null
+  ] if try(v.environment_variables, null) != null
+  ]))
+  env_vars_for_locker_for_locker_parameters = { for entry in local.env_vars_for_locker : "${entry.connector_name}.${entry.env_var_name}" => entry }
```
aperez-worklytics, line 173: Instead of creating a *fake* secret through specs, I've move it to the combination of new *gcp-instance-secret-locker* module and to the variable created here. This is something specific for GCP due the way we work with labels for sharing thing.

infra/modules/gcp-oauth-secrets-role/main.tf
aperez-worklytics: Ok, moved to GCP module

infra/modules/gcp/main.tf
```diff
@@ -183,6 +183,21 @@ moved {
   to   = google_project_iam_custom_role.bucket_write
 }
 
+resource "google_project_iam_custom_role" "psoxy_instance_secret_locker_role" {
```
aperez-worklytics, line 186: The new role, but in GCP module.

infra/modules/worklytics-connector-specs/main.tf
aperez-worklytics: Now required as `HSE_SHARED_TOKEN`  property is used for GCP to create the secret, but in `account_credentials` grant type (the one Zoom uses) that value is always hardcoded to true.

infra/modules/worklytics-connector-specs/main.tf
aperez-worklytics: Yeah I was to do it in next PR, but included here.

infra/modular-examples/gcp-google-workspace/main.tf
eschultink: why `for_locker` twice?

infra/modular-examples/gcp-google-workspace/main.tf
```diff
@@ -191,7 +193,18 @@ module "connector-oauth" {
   secret_name           = "PSOXY_${upper(replace(each.value.connector_name, "-", "_"))}_${upper(each.value.secret_name)}"
   project_id            = var.gcp_project_id
   service_account_email = google_service_account.long_auth_connector_sa[each.value.connector_name].email
-  updater_role_id       = module.connector-oauth-secret-role.role_id
+}
+
+module "psoxy-instance-secrer-locker" {
```
eschultink, line 198:
```suggestion
module "psoxy-instance-secret-locker" {
```

eschultink (changes requested):
👍 just one typo to change

eschultink (approved):

PR #369, opened 2023-06-06
module and example arch

eschultink:
### TODO
 - [x] infra/modules/worklytics-connectors/main.tf deal with locker stuff
 - [x] aws IAM role naming : `PsoxyExec_all-dev-outlook-mail` --> `all-dev_Exec_outlook-mail`
 - [x] aws IAM role naming : `BucketRead_psoxy-aws-all-dev-hris` --> 
 - [x] bulk naming : `psoxy-all-dev-hris`
 - [x] identity pool error  : `software.amazon.awssdk.services.cognitoidentity.model.CognitoIdentityException - User: arn:aws:sts::705189459437:assumed-role/PsoxyExec_all-dev-azure-ad/all-dev-azure-ad is not authorized to perform: cognito-identity:GetOpenIdTokenForDeveloperIdentity on resource: arn:aws:cognito-identity:us-east-1:705189459437:identitypool/us-east-1:0f13da79-8313-44bf-996a-4c9fe5589a1f because no identity-based policy allows the cognito-identity:GetOpenIdTokenForDeveloperIdentity action (Service: CognitoIdentity, Status Code: 400, Request ID: 5ebb0017-59bf-45bb-bfb3-ba6ea1c213f1, Extended Request ID: null)`

### Features
  - gcp/aws host examples, decoupled from data source; current aws/gcp examples that support ALL sources depend on az CLI + gcloud CLI, even if those sources aren't used

### Change implications

---
  - https://app.asana.com/0/0/1204451002484765
  - https://app.asana.com/0/0/1204472640908681


infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -557,6 +558,7 @@ EOT
       secured_variables : [
         { name : "ACCESS_TOKEN", writable : true },
         { name : "REFRESH_TOKEN", writable : true },
+        { name : "OAUTH_REFRESH_TOKEN", writable : true, lockable : true },
```
eschultink, line 561: @aperez-worklytics note this is the solution Jira/Zoom locking; please take a good look, and we should try to test

infra/examples-dev/aws-all/variables.tf
```diff
@@ -0,0 +1,251 @@
...
+  description = "friendly qualifier to distinguish resources created by this terraform configuration other Terraform deployments, (eg, 'prod', 'dev', etc)"
+
+  validation {
+    condition     = can(regex("^[a-zA-Z][a-zA-Z0-9-_ ]*[a-zA-Z0-9]$", var.environment_name))
+    error_message = "The `environment_name` must start with a letter, can contain alphanumeric characters, hyphens, underscores, and spaces, and must end with a letter or number."
+  }
+
+  validation {
+    condition     = !can(regex("^(?i)(aws|ssm)", var.environment_name))
+    error_message = "The `environment_name` cannot start with 'aws' or 'ssm', as this will name your AWS resources with prefixes that displease the AMZN overlords."
```
jlorper, line 12: :rofl:

infra/examples-dev/aws-all/variables.tf
```diff
@@ -0,0 +1,251 @@
...
+variable "aws_assume_role_arn" {
+  type        = string
+  description = "ARN of role Terraform should assume when provisioning your infra. (can be `null` if your CLI is auth'd as the right user/role)"
+  default     = null
+}
+
+variable "aws_region" {
+  type        = string
+  description = "default region in which to provision your AWS infra"
+  default     = "us-east-1"
```
jlorper, line 35: remove default to ensure is set always by customer?

infra/examples-dev/aws-all/variables.tf
```diff
@@ -0,0 +1,251 @@
...
+variable "salesforce_domain" {
+  type        = string
+  description = "Domain of the Salesforce to connect to (only required if using Salesforce connector). To find your My Domain URL, from Setup, in the Quick Find box, enter My Domain, and then select My Domain"
+  default     = ""
+}
+
+variable "jira_server_url" {
+  type        = string
+  default     = null
+  description = "(Only required if using Jira Server connector) URL of the Jira server (ex: myjiraserver.mycompany.com)"
```
jlorper, line 87: rename to `jira_server_host`? looks more appropriate

infra/examples-dev/gcp/main.tf
```diff
@@ -27,9 +27,50 @@ provider "google" {
...
+# be provisioned via Terraform, so doesn't add any dependencies
+# call this 'generic_source_connectors'?
+module "worklytics_connectors" {
+  source = "../../modules/worklytics-connectors"
+
+  enabled_connectors    = var.enabled_connectors
+  example_jira_issue_id = var.example_jira_issue_id
+  jira_cloud_id         = var.jira_cloud_id
+  jira_server_url       = var.jira_server_url
+  salesforce_domain     = var.salesforce_domain
```
jlorper, line 48: we should probably combine this under an object `connector_settings`, so it can grow just changing the description of the variable rather than passing every single one here.

jlorper (commented):
Still half-way through. A couple of suggestions so far.

infra/examples-dev/aws-all/google-workspace-variables.tf
```diff
@@ -0,0 +1,17 @@
+variable "gcp_project_id" {
+  type        = string
+  description = "string ID of GCP project that will host psoxy instance; must exist. Can leave null if not using GCP/Google Workspace."
+  default     = null
+}
+
+variable "google_workspace_example_user" {
```
jlorper, line 7: this doesn't seem to be used if google_workspace_example_admin is set.

Wouldn't be just easier to have one and say that some calls might not be available if the user hasn't certain admin rights?

infra/examples-dev/aws-all/main.tf
```diff
@@ -0,0 +1,161 @@
+terraform {
```
jlorper, line 1: what about splitting L1-L26 into `providers.tf` and `state.tf`, that way `main.tf` could be reused too

infra/examples-dev/aws-all/main.tf
```diff
@@ -0,0 +1,161 @@
...
+}
+
+
+## Host platform (AWS) configuration
+
+# NOTE: you need to provide credentials. usual way to do this is to set env vars:
+#        AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
+# see https://registry.terraform.io/providers/hashicorp/aws/latest/docs#authentication for more
+# information as well as alternative auth approaches
+provider "aws" {
```
jlorper, line 70: If following suggestion move to  `providers.tf`

infra/modules/aws-host/main.tf
```diff
@@ -0,0 +1,194 @@
+# NOTE: region used to be passed in as a variable; put it MUST match the region in which the lambda
+# is provisioned, and that's implicit in the provider - so we should just infer from the provider
```
jlorper, line 2: ok, then why passing it in variables? totally eliminate then?

infra/modules/env-id/README.md
```diff
@@ -0,0 +1,43 @@
...
+Expectation is that users of terraform modules/examples will want to 'namespace' infra resources
+created by a given Terraform configuration in a manner that groups them together via the interfaces
+of the host platform, making it easy to find all related infrastructure.
+
+To this end, the convention of an `environment_name` variable will be supported.
+
+As this value will be used to prefix infrastructure resource names/ids in many systems with different
+restrictions on allowed characters, we will enforce some universal restrictions on the value, as
+specify some transformations:
+- restrict characters to `[A-Za-z0-9-_ ]`, starting and ending with an alphabetical character
```
jlorper, line 12: could default to `w8s-dev` for all new customers when in testing phase, then `w8s-prod`

jlorper (approved):
Some more suggestions, but looks good.
Maybe we the new modular example like **aws-all** is promoted to be default it deserves a bump to 0.5

infra/examples-dev/aws-all/variables.tf
eschultink: fair idea, although our bugs are normally when they set this to something *other* than `us-east-1`, but other tooling defaults to `us-east-1` (such as the npm packages used in the testing tools)

infra/examples-dev/aws-all/variables.tf
eschultink: host would imply it doesn't include a path ... which it might, right?

@aperez-worklytics ?  we ideally should be consistent everywhere; jira via proxy not really released yet, so now's the time to make a change

infra/examples-dev/gcp/main.tf
eschultink: I don't agree.

It would make configuration in the `terraform.tfvars` file less straight-forward. To add sources, you couldn't just append to it; but would have to find the `connector_settings` variable within that and modify it to add what you want.

infra/examples-dev/aws-all/google-workspace-variables.tf
eschultink: 🤔 in theory, it should be the user impersonated EXCEPT for the gdirectory/google-chat/google-meet connectors, which required a privileged user (admin)

infra/examples-dev/aws-all/main.tf
eschultink: 🤔 yes, that's worth thinking about

infra/examples-dev/aws-all/main.tf
eschultink: imho, I think better to organize it as an `aws.tf` right? as the `aws` provider is coupled to the "feature" of hosting in AWS. 

But yeah, doesn't exactly extend well to GCP case, as we'd end up with `google` provider defined in both the `gcp.tf` for the host as well as potentially a `google-workspace.tf` for the data source.

infra/modules/aws-host/main.tf
eschultink: mmm, yeah; this is copy-paste from places where eliminating it would be a breaking change, but it can be removed from `aws-host` as this is a new module that no existing deployments should reference.

infra/modules/worklytics-connector-specs/main.tf
aperez-worklytics: Yeah that the thing I was to asking you; ok for that solution but we'll need to update this on java side too

PR #374, opened 2023-06-19
avoid duplicated example code

eschultink:
### Fixes
 - various edge cases around environment ids/names, mainly to be robust against short/empty ones in GCP case

### Features
  - use hardlinks to avoid having multiple couples of identical files for different examples (pros and cons of this approach; ideas welcome)

### Change implications


infra/examples-dev/gcp/google-workspace.tf
eschultink, line 1: this was created as a hardlink to `example-lib/`; `git`/`GitHub` don't know the difference however; and I don't think OS does either.  as I understand it, hardlinks are  just another reference in the file system to the physical page on the disk of the file metadata/content - so equivalent to the original file name. 

the only advantage is changing it one will change it from all 3 perspectives.

infra/examples-dev/gcp/google-workspace-variables.tf
eschultink, line 1: this was created as a hardlink to `example-lib/`; `git`/`GitHub` don't know the difference however; and I don't think OS does either.  as I understand it, hardlinks are  just another reference in the file system to the physical page on the disk of the file metadata/content - so equivalent to the original file name. 

the only advantage is changing it one will change it from all 3 perspectives.

aperez-worklytics (approved):
I don't find a better idea, so ok for me

PR #379, opened 2023-06-20
fix conflicting gcp SA ids by adding -fn qualifier

eschultink:
### Fixes
 - since split of GCP host from Google Workspace as data source, potential for SA names used by both to conflict; this fixes that

### Change implications

---
  - https://app.asana.com/0/0/1204771340174904

### DIFF
--- a/infra/modules/gcp-output-bucket/variables.tf
+++ b/infra/modules/gcp-output-bucket/variables.tf
@@ -55,8 +55,10 @@ variable "sanitizer_accessor_principals" {
   default     = []
 }
 
+# NOTE: while this value *can* be 0, that's the same as setting no expiration, which is OK except
+# that GCP will not persist the lifecycle rule for the expiration, so Terroform will always show change
 variable "expiration_days" {
   type        = number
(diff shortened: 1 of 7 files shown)


infra/modules/gcp-output-bucket/variables.tf
```diff
@@ -55,8 +55,10 @@ variable "sanitizer_accessor_principals" {
   default     = []
 }
 
+# NOTE: while this value *can* be 0, that's the same as setting no expiration, which is OK except
```
jlorper, line 58: add validation rule then?

infra/modules/gcp-output-bucket/variables.tf
```diff
@@ -55,8 +55,10 @@ variable "sanitizer_accessor_principals" {
...
+# that GCP will not persist the lifecycle rule for the expiration, so Terroform will always show change
```
jlorper, line 59:
```suggestion
# that GCP will not persist the lifecycle rule for the expiration, so Terraform will always show change
```

jlorper (approved):

PR #380, opened 2023-06-20
add test-all script to aws/gcp examples

eschultink:
### Features
 - a `test-all.sh` to facilitate quick testing of ALL connectors after deploy

### Change implications

---
  - https://app.asana.com/0/0/1204736719765095

### DIFF
--- a/infra/modules/aws-host/main.tf
+++ b/infra/modules/aws-host/main.tf
@@ -190,3 +190,23 @@ locals {
   all_instances = merge(local.api_instances, local.bulk_instances)
 }
 
+# script to test ALL connectors
+resource "local_file" "test_all_script" {
+  filename        = "test-all.sh"
+  file_permission = "0770"
+  content         = <<EOF
+#!/bin/bash
+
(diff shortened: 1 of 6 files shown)


infra/modules/aws-host/main.tf
```diff
@@ -190,3 +190,23 @@ locals {
   all_instances = merge(local.api_instances, local.bulk_instances)
 }
 
+# script to test ALL connectors
+resource "local_file" "test_all_script" {
```
jlorper, line 194: exactly the same as gcp-host. Extract to own file and do sym link?

jlorper (approved):

infra/modules/aws-host/main.tf
eschultink: Terraform doesn't seem to like to follow sym links. I tried it when splitting `main.tf`/`variables.tf` ... I have not tried within modules, only at top level (root) configurations.  But yeah, maybe could try more.

There is A LOT that's very, very similar between AWS and GCP cases.

PR #381, opened 2023-06-21
gcp default_labels support

eschultink:
### Features
  - default_labels for gcp modules

### Change implications

---
  - https://app.asana.com/0/0/1204867920858701


infra/modules/gcp-host/main.tf
```diff
@@ -204,6 +210,7 @@ resource "google_secret_manager_secret" "additional_transforms" {
 
   project   = var.gcp_project_id
   secret_id = "${local.config_parameter_prefix}${upper(replace(each.key, "-", "_"))}_ADDITIONAL_TRANSFORMS"
+  labels    = var.default_labels
```
jlorper, line 213: should we do ignore changes on labels always? if so add the ignore block
first time on creation it will get the labels (if any set), then will be ignored thereafter. 
This could be told to customer to really specify labels if needed before first deployment to production.

jlorper (approved):
just one thing on the secret regarding the ignore lifecycle

infra/modules/gcp-host/main.tf
eschultink: yeah, did that change in the few places it was missing. more consistent that way.

PR #384, opened 2023-06-21
gcp custom bucket names

eschultink:
### Features
  - add ability to customize bucket names for gcp hosted deployments

### Change implications


infra/modules/gcp-psoxy-bulk/main.tf
```diff
@@ -89,7 +89,8 @@ module "output_bucket" {
   project_id                     = var.project_id
   bucket_write_role_id           = var.bucket_write_role_id
   function_service_account_email = google_service_account.service_account.email
-  bucket_name_prefix             = local.bucket_prefix
+  bucket_name_prefix             = coalesce(var.sanitized_bucket_name, local.bucket_prefix)
+  bucket_name_suffix             = var.sanitized_bucket_name == null ? "-output" : ""
```
jlorper, line 93:
```suggestion
  bucket_name_prefix             = coalesce(var.sanitized_bucket_name, local.bucket_prefix)
  bucket_name_suffix             = var.sanitized_bucket_name == null ? "" : "-sanitized"
```
1. isn't it twisted? if a name is provided ignore the suffix
2. "sanitized" maybe works better than "output"?

jlorper (changes requested):
I think the condition in the output bucket is reversed

infra/modules/gcp-psoxy-bulk/main.tf
eschultink: it has to be left as `-output`, or will be breaking change (existing configs will have their buckets destroyed and rebuilt when upgrading versions)

if the `sanitized_bucket_name` is not set, then we want defaults (which for legacy suffix is `-output`; 
if it IS set, then we want `""` and will just pass a prefix that is actually the WHOLE bucket name.

eschultink: the actual bucket name is just "${prefix}${suffix}"

it's kinda confusing, but this bucket module is used to create `-input`, `-santized`, and `-lookup` buckets each with the same prefix - for example. Just to make it clear that all are related. It probably would have been better to have just passed a name, and do the concatenation repeatedly up a level - but not so straightforward to refactor

jlorper: yeah, you are right. My brain just twisted the condition when I read it.

jlorper (approved):

infra/modules/gcp-psoxy-bulk/main.tf
jlorper: but where was the suffix set before?

jlorper: nevermind, I saw it was defaulting to "-output" ok, that was confusing

jlorper (approved):

PR #386, opened 2023-06-22
tf cloud 0.2

eschultink:
### Fixes
 - missing ref in example

### Change implications


tools/build.sh
```diff
@@ -6,15 +6,23 @@
...
 
+while getopts ":q" opt; do
+  case $opt in
+    q)
+      QUIET_OPTIONS="-q -Dmaven.test.skip=true"
+      ;;
+  esac
+done
+
+QUIET_OPTIONS="-q -Dmaven.test.skip=true"
```
aperez-worklytics, line 17: :yellow_circle:  As now quiet is optional, this line it seems wrong -it will enforce it, no matter the setting you provide-. Maybe you want to just initialize it before the while loop to an empty value.

aperez-worklytics (approved):

tools/init-tfvars.sh
```diff
@@ -78,35 +88,51 @@ if test $GOOGLE_PROVIDER_COUNT -ne 0; then
...
-        # example admin for Google Workspace
-    printf "# Google Workspace example admin \n" >> $TFVARS_FILE
-    printf "#  - this is used to aid testing of Google Workspace connectors against a real account, in cases where an admin is explicitly required\n" >> $TFVARS_FILE
-    GOOGLE_WORKSPACE_EXAMPLE_USER=$(gcloud config get account)
-    printf "google_workspace_example_admin=\"${GOOGLE_WORKSPACE_EXAMPLE_USER}\"\n\n" >> $TFVARS_FILE
-    printf "\tgoogle_workspace_example_admin=${BLUE}\"${GOOGLE_WORKSPACE_EXAMPLE_USER}\"${NC}\n"
+    [[ -f variables.tf ]] && grep -q '^variable "worklytics_sa_emails"' variables.tf
+    if [[ $? -eq 0 ]]; then
+      printf "# GCP service account emails in the list below will be allowed to invoke your proxy instances\n" >> $TFVARS_FILE
+      printf "#  - if you're deploying to AWS rather than GCP, you can omit this value\n" >> $TFVARS_FILE
```
jlorper, line 102: skip if HOST_PLATFORM === "aws" ? or remove and do the whole block if host === "gcp"
```suggestion
      printf "#  - if you're deploying to AWS rather than GCP, you can omit this value\n" >> $TFVARS_FILE
```

jlorper (approved):
one suggestion

PR #388, opened 2023-06-27
support AWS cmeks for lambda env, ssm params, cloudwatch log groups; add s3 example

eschultink:
### Features
  - expose variable to encrypt ssm parameters with CMEK
  - support encrypting cloud watch log groups with CMEK, expose in example
  - support encrypting lambda env vars with CMEK, expose in example
  - add an example of encrypting S3 bucket with CMEK (currently supported, but requires some advanced tf config)

### Change implications

---
  - https://app.asana.com/0/0/1204910069771257

### DIFF
--- a/infra/examples-dev/aws-all/kms-cmek.tf
+++ b/infra/examples-dev/aws-all/kms-cmek.tf
@@ -1,0 +1,37 @@
+
+# uncomment to test this
+
+
+#resource "aws_kms_key_policy" "key_policy_including_cloudwatch" {
+#  key_id = var.project_aws_kms_key_arn
+#  policy = jsonencode(
+#    {
+#      "Version" : "2012-10-17",
+#      "Id" : "key-default-1",
+#      "Statement" : [
+#        {
+#          "Sid": "Allow IAM Users to Manage Key",
+#          "Effect": "Allow",
+#          "Principal": {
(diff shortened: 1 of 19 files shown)


infra/examples-dev/aws-all/kms-cmek.tf
```diff
@@ -0,0 +1,101 @@
+
+# uncomment to use encryption for S3 buckets and have it work properly for Cloud Watch Logs
+
+
+#locals {
```
jlorper, line 5: maybe less tedious using multi-line comments per block
https://developer.hashicorp.com/terraform/language/syntax/configuration#comments
like
```tf
/*
testing_policy_statemens = var.provision_testing_infra ? [
...
] : []
*/

```

infra/modules/aws-psoxy-lambda/main.tf
```diff
@@ -19,6 +19,11 @@ module "env_id" {
 locals {
   salt_parameter_name_suffix = "PSOXY_SALT"
   function_name              = "${module.env_id.id}-${var.instance_id}"
+
+  kms_keys_to_allow_ids = merge(
```
jlorper, line 23: To me sounds more natural like this
```suggestion
  kms_key_ids_to_allow = merge(
```

jlorper (approved):
A couple of suggestions, no blockers

PR #403, opened 2023-06-29
zoom configured with two environment_variable maps

eschultink:
### Fixes
 - zoom missing `REFRESH_ENDPOINT`

### Change implications


infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -405,9 +406,6 @@ EOT
         { name : "ACCESS_TOKEN", writable : true },
         { name : "OAUTH_REFRESH_TOKEN", writable : true, lockable : true }, # q: needed? per logic as of 9 June 2023, would be created
       ],
-      environment_variables : {
-        USE_SHARED_TOKEN : "TRUE"
-      }
```
eschultink, line 410: why hashicorp would think allowing a property to be set twice in same object literal should be legal syntax is beyond me, especially given that values are otherwise immutable in hcl ...

jlorper (approved):

PR #404, opened 2023-06-30
fix JIRA cloud moves

eschultink:
### Fixes
  - migration missed a jira-cloud case (and also missed a perm that should move, although less critical)

### Change implications


aperez-worklytics (approved):
Sorry, I've missed them from prev review

jlorper (approved):

PR #405, opened 2023-06-30
fix writable secrets

eschultink:
### Fixes
 - writable secrets don't have correct perms in GCP

### Change implications

---
  - https://app.asana.com/0/0/1204940253452631

### DIFF
--- a/infra/modules/gcp-host/main.tf
+++ b/infra/modules/gcp-host/main.tf
@@ -30,6 +30,15 @@ locals {
 
 # BEGIN API CONNECTORS
 
+resource "google_service_account" "api_connectors" {
+  for_each = var.api_connectors
+
+  project      = var.gcp_project_id
+  account_id   = substr("${local.sa_prefix}${replace(each.key, "_", "-")}", 0, local.SA_NAME_MAX_LENGTH)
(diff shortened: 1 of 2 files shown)


jlorper (approved):

infra/modules/gcp-host/main.tf
```diff
@@ -71,10 +76,20 @@ module "secrets" {
...
   role      = module.psoxy.psoxy_instance_secret_locker_role_id
+
+}
+
+resource "google_secret_manager_secret_iam_member" "grant_sa_secretVersionAdder_on_writable_secret" {
+  for_each = { for secret in local.writable_secrets : secret.instance_secret_id => secret }
+
   project   = var.gcp_project_id
   secret_id = "${local.config_parameter_prefix}${each.value.instance_secret_id}"
+  member    = "serviceAccount:${google_service_account.api_connectors[each.value.instance_id].email}"
```
jlorper, line 91: Obviously it doesn't matter for TF but kind of confusing reading the code that the api connectors SA are defined below. I'd probably do:
1. define SA connectors
2. define Secrets (dependent on 1)
3. generate each api connector (dependent on 1 and 2)

right now code does 2,1,3

jlorper (approved):
Just one suggestion on how that file is organized

PR #419, opened 2023-07-11
split Terraform-managed secrets/parameters from externally-managed ones

eschultink:
### Features
 - split secrets/configuration values that are managed within Terraform, from those managed outside Terraform - this will result in clearer plans, and minimize creation of versions of with dummy values. 

### TODO:
 - [x] whole bunch of moves for the AWS case (GCP not needed, bc deletion of placeholder versions OK / expected)
 - [x] changelog NOTES on the moves in AWS case, deletions in GCP case
 - [x] bunch of testing of this

### Change implications

---
  - https://app.asana.com/0/0/1204973045518667
  - https://app.asana.com/0/0/1205043877019559


infra/modules/gcp-secrets/main.tf
```diff
@@ -5,6 +5,9 @@
 
 locals {
   replica_locations = coalesce(var.replica_regions, var.replica_locations)
+
+  terraform_managed_secrets  = { for k, v in var.secrets : k => v if v.value_managed_by_tf }
```
aperez-worklytics, line 9: `terraform_managed_secrets` as var name makes more sense in AWS, but here as we are going to create or not the versions, I'd rename it `terraform_managed_secrets_versions` or something like that

aperez-worklytics: Yeah, as https://github.com/Worklytics/psoxy/pull/419/files#diff-169b5a139f9d65376c18eb220ce34546a3fde51283399f09e28cc60c7ce224e9L11 it could be confusing why we are creating all secrets coming from `var.secrets` but only versions from `terraform_managed_secrets`. I think that the rename worth it...

aperez-worklytics (changes requested):
Some naming suggestions; and comments about file migration and missing github settings

infra/modules/gcp-secrets/main.tf
eschultink: ok, yeah - tried to clean it up.

aperez-worklytics (approved):

PR #420, opened 2023-07-11
S153 fix variable passing

aperez-worklytics:
- Fixed variable passing to worklytics spec module from worlytics connector
- Adding salesforce domain variable validation at top level, only if salesforce connector is enabled
- Fixed a NPE when not deploying from bundle
- Including TARGET_HOST as part of health check validation. If empty or missing, it will produce a 512 + being included in the response model:

```json
{
    "bundleFilename": "psoxy-gcp-0.4.29.jar",
    "configPropertiesLastModified": null,
    "configuredSource": "salesforce",
    "missingConfigProperties": [
        "TARGET_HOST"
    ],
    "nonDefaultSalt": true,
    "pseudonymImplementation": null,
    "sourceAuthGrantType": "client_credentials",
    "sourceAuthStrategy": "oauth2_refresh_token",
    "version": "0.4.29"
}
```

### Fixes
[salesforce_domain fixes](https://app.asana.com/0/1201039336231823/1204991668553251)

### Change implications


infra/examples-dev/aws-all/variables.tf
```diff
@@ -294,3 +294,14 @@ variable "todos_as_local_files" {
   description = "whether to render TODOs as flat files"
   default     = true
 }
+
+locals {
+  # tflint-ignore: terraform_unused_declarations
+  validate_salesforce_domain         = (var.salesforce_domain == null || var.salesforce_domain == "") && contains(var.enabled_connectors, "salesforce")
+  validate_salesforce_domain_message = "The salesforce_domain var should be populated if enabled."
+  validate_salesforce_domain_check = regex(
```
aperez-worklytics, line 302: Weird validation, open if there are other way. The problem is that `validation` can only work with same variable; in this case it involves two vars (domain + enabled connectors). 

It could be solved in the future if we have an object for all stuff related with connectors, so we don't need to add variables in all the places everytime we want to add/modify something and maybe it would be cleaner.

infra/examples-dev/gcp-google-workspace/main.tf
```diff
@@ -51,6 +51,7 @@ module "psoxy" {
   jira_server_url                = var.jira_server_url
   jira_cloud_id                  = var.jira_cloud_id
   example_jira_issue_id          = var.example_jira_issue_id
+  deployment_bundle              = var.deployment_bundle
```
aperez-worklytics, line 54: I know is deprecated, but missing.

infra/modules/aws-psoxy-lambda/main.tf
```diff
@@ -40,9 +40,9 @@ locals {
   salt_arn              = [for l in var.global_parameter_arns : l if endswith(l, local.salt_parameter_name_suffix)][0]
   path_to_shared_config = regex("arn.+parameter(/.*)${local.salt_parameter_name_suffix}", local.salt_arn)[0]
 
-  bundle_from_s3       = startswith(var.path_to_function_zip, "s3://")
-  s3_bucket            = local.bundle_from_s3 ? regex("s3://([^/]+)/.*", var.path_to_function_zip)[0] : null
-  s3_key               = local.bundle_from_s3 ? regex("s3://[^/]+/(.*)", var.path_to_function_zip)[0] : null
+  bundle_from_s3 = startswith(var.path_to_function_zip, "s3://")
```
aperez-worklytics, line 43: Just indent

infra/modules/gcp/main.tf
```diff
@@ -141,12 +141,12 @@ moved {
 }
 
 locals {
-  is_remote_bundle       = var.deployment_bundle != null && startswith(var.deployment_bundle, "gs://")
+  is_remote_bundle       = var.deployment_bundle != null && startswith(var.deployment_bundle == null ? "" : var.deployment_bundle, "gs://")
```
aperez-worklytics, line 144: Had to add this, if a variable with a null value is provided to `startswith` it will complain. Dind't check if happening in AWS side @eschultink

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandler.java
```diff
@@ -51,53 +52,62 @@ private boolean isHealthCheckRequest(HttpEventRequest request) {
...
-        HealthCheckResult.HealthCheckResultBuilder healthCheckResult = HealthCheckResult.builder()
-            .configuredSource(config.getConfigPropertyAsOptional(ProxyConfigProperty.SOURCE).orElse(null))
-            .nonDefaultSalt(config.getConfigPropertyAsOptional(ProxyConfigProperty.PSOXY_SALT).isPresent())
-            .pseudonymImplementation(config.getConfigPropertyAsOptional(ProxyConfigProperty.PSEUDONYM_IMPLEMENTATION).orElse(null))
-            .missingConfigProperties(missing);
+        try {
+            Optional<String> targetHost = config.getConfigPropertyAsOptional(ProxyConfigProperty.TARGET_HOST);
 
+            if (targetHost.isEmpty() || StringUtils.isBlank(targetHost.get())) {
+                missing.add(ProxyConfigProperty.TARGET_HOST.name());
```
aperez-worklytics, line 64: Including TARGET_HOST as missing property; not a config property itself, but is something that might be configured. Also, adding it as a boolean flag in `HealthCheckResult` will force us to update the lib and put code to handle that validation in app side.

infra/modules/worklytics-connectors/variables.tf
```diff
@@ -27,8 +27,32 @@ variable "example_jira_issue_id" {
...
+  type        = string
+  description = "ID of Microsoft tenant to connect to (req'd only if config includes MSFT connectors)"
+  default     = ""
+}
+
+variable "example_msft_user_guid" {
+  type        = string
+  description = "example MSFT user guid (uuid) for test API calls (OPTIONAL)"
+  default     = "{EXAMPLE_MSFT_USER_GUID}"
+}
```
eschultink, line 52: we can rename the module, but these cases are handled through
https://github.com/Worklytics/psoxy/tree/main/infra/modules/worklytics-connectors-msft-365

and https://github.com/Worklytics/psoxy/tree/main/infra/modules/worklytics-connectors-google-workspace

so the msft/google cases are intentionally not exposed here

infra/examples-dev/aws-all/main.tf
```diff
@@ -34,11 +34,15 @@ module "worklytics_connectors" {
...
-  salesforce_domain     = var.salesforce_domain
+  enabled_connectors             = var.enabled_connectors
+  example_jira_issue_id          = var.example_jira_issue_id
+  jira_cloud_id                  = var.jira_cloud_id
+  jira_server_url                = var.jira_server_url
+  salesforce_domain              = var.salesforce_domain
+  msft_tenant_id                 = var.msft_tenant_id
+  example_msft_user_guid         = var.example_msft_user_guid
+  google_workspace_example_admin = var.google_workspace_example_admin
+  google_workspace_example_user  = var.google_workspace_example_user
```
eschultink, line 45: the google + msft are split out in their own files, to facilitate easy adaption of the example

infra/examples-dev/aws-all/variables.tf
eschultink: yeah due to this limitation in Terraform I think best approach is to validate that `var.salesforce_domain`, if provided, is domain without path or protocol.

Then validate `TARGET_HOST` in java as you have. 

additionally, we could also use https://developer.hashicorp.com/terraform/language/checks on `TARGET_HOST`; but as only from 1.5+ may need to wait on that, as many people still on 1.3x or 1.4x

infra/examples-dev/gcp-google-workspace/main.tf
eschultink: yeah, fine - the feature just isn't implemented for that case - but OK if it works ...

eschultink (changes requested):

infra/modules/gcp/main.tf
eschultink: 👍 thanks, yeah, I had this on different branch. I think using `try()` more canonical than nested ternary, but reasonable

jlorper: weird, doesn't tf have lazy evaluation? if null shouldn't be evaluating the `startswith`

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandler.java
jlorper: I don't see why this try/catch is needed at all, how is adding a String to a Set going to fail??

jlorper (approved):
Some comments, but fine by me after you deal with Erik's changes.

infra/modules/gcp/main.tf
eschultink: yeah, it doesn't; have seen this before

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandler.java
aperez-worklytics: Not for that, just for reading the setting from config. It should not fail as it is an optional get, but just to be defensive here.

aperez-worklytics:
@eschultink when possible take a look on the PR

infra/examples-dev/aws-all/variables.tf
```diff
@@ -286,3 +286,14 @@ variable "todos_as_local_files" {
   description = "whether to render TODOs as flat files"
   default     = true
 }
+
+locals {
+  # tflint-ignore: terraform_unused_declarations
+  validate_salesforce_domain         = (var.salesforce_domain == null || var.salesforce_domain == "" || can(regex(":|\\/", try(var.salesforce_domain, "")))) && contains(var.enabled_connectors, "salesforce")
+  validate_salesforce_domain_message = "The salesforce_domain var should be populated and to be with only the domain (without protocol or query paths) itself if enabled."
```
jlorper, line 293: I find this message a bit weird
```suggestion
  validate_salesforce_domain_message = "salesforce_domain var must be populated with just the domain (no protocol or query paths)"
```
if appears if because the connector is enabled

jlorper (approved):

eschultink (approved):

PR #417, opened 2023-07-10
S151 Github: Rules for proxy

aperez-worklytics:
- Added documentation about how to build the connector through Github Apps
- Added bunch of rules and tests for GH REST API endpoints (repos, logs, events, timeline, pull request, commits, comments, etc)
- Added a rule for querying the GraphQL only for users; as the endpoint is open to retrieve everything is done in a request, a response has been created to include only the stuff that we are going to query.

NOTE: Most of the lines included in the PR are related to the files and sanitized files that it contains.

### Features
[Psoxy rules support](https://app.asana.com/0/0/1204953309467291)

### Change implications

 - dependencies added/changed? **no**

PR #426, opened 2023-07-13
S153 gh jwt support

aperez-worklytics:
**NOTE** Parent PR is #417 

For Github, adding support for using a [JWT](https://docs.github.com/en/enterprise-server@3.8/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app) to [retrieve an access token](https://docs.github.com/en/enterprise-server@3.8/apps/creating-github-apps/authenticating-with-a-github-app/generating-an-installation-access-token-for-a-github-app)  that can be used for making regular requests to the API. This is for fully supporting server-to-server scenario, instead of using user tokens.

See review comments for further detail, questions. Open for suggestions for token naming.

### Features
[Psoxy: GH with JWT support](https://app.asana.com/0/0/1205004120364723)

### Change implications


infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -274,15 +275,15 @@ EOT
       target_host : "api.github.com"
       source_auth_strategy : "oauth2_refresh_token"
       secured_variables : [
-        { name : "ACCESS_TOKEN", writable : true },
+        { name : "PRIVATE_KEY", writable : true },
```
aperez-worklytics, line 278: Very similar with client_credentials for MSFT certificates; here should be the value of the converted PKSC8 certificate downloaded from the application created.

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -274,15 +275,15 @@ EOT
...
         { name : "REFRESH_TOKEN", writable : true },
         { name : "OAUTH_REFRESH_TOKEN", writable : true, lockable : true },
-        { name : "CLIENT_ID", writable : false },
-        { name : "CLIENT_SECRET", writable : false }
+        { name : "CLIENT_ID", writable : false }
       ],
       environment_variables : {
-        GRANT_TYPE : "refresh_token"
-        REFRESH_ENDPOINT : "https://github.com/login/oauth/access_token"
+        GRANT_TYPE : "certificate_credentials"
```
aperez-worklytics, line 284: As a difference from `client_credentials`, this is just to use the JWT obtained without key and directly as authentication header, not as a POST request with a body.

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -274,15 +275,15 @@ EOT
...
+        TOKEN_RESPONSE_TYPE : "GITHUB_ACCESS_TOKEN"
```
aperez-worklytics, line 285: By default all response are handled in the same way for building `CanonicalOAuthAccessTokenResponseDto`, but Github response is different. This is for adding support for that in `GithubAccessTokenResponseParserImpl`

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -265,8 +272,7 @@ private CanonicalOAuthAccessTokenResponseDto exchangeRefreshTokenForAccessToken(
 
             HttpResponse response = tokenRequest.execute();
             CanonicalOAuthAccessTokenResponseDto tokenResponse =
-                objectMapper.readerFor(CanonicalOAuthAccessTokenResponseDto.class)
-                    .readValue(response.getContent());
+                    tokenResponseParser.parseTokenResponse(response);
```
aperez-worklytics, line 275: Now with the interface, leaving in `TokenResponseParser` instance the responsibility of converting the response into `CanonicalOAuthAccessTokenResponseDto`

eschultink (approved):

infra/modules/worklytics-connector-specs/main.tf
eschultink: we have to be super clear; may be worth adding 'description' field here as well, so it even ends up being specified on the parameter/secret what this value is and where it's coming from, particularly where it depends on a specific format.  

with description, person seeing it in AWS / GCP console when they fill it might notice if they're about to fill the wrong format (prob not super relevant here, as most people won't know whether they have PKSC8 or not)

docs/sources/github.md
```diff
@@ -11,34 +11,43 @@ Follow the following steps:
...
 - "Homepage URL" can be anything, not required in this flow but required by Github.
-- Callback URL is required for next step, it can be any URL (http://localhost, for example);
 - Webhooks check can be disabled as this connector is not using them
 - Keep `Expire user authorization tokens` enabled, as GitHub documentation recommends
-2. Once is created please generate a new client secret. We will both (clientId and clientSecret) in next steps.
-3. Install the application in your organization.
+2. Once is created please generate a new `Private Key`.
+3. It is required to convert the format of the certificate downloaded from PKCS#1 in previous step to PKCS#8. Please run following command:
+```shell
+openssl pkcs8 -topk8 -inform PEM -outform PEM -in {YOUR DOWNLOADED CERTIFICATE FILE} -out priv8.pem -nocrypt
```
jlorper, line 34: I'd rename `priv8.pem` to something more meaningful, like `gh_pk_pkcs8.pem`. (here and in L50) 

And also I'd suggest a step 7 to remove the downloaded files pkcs1 and pkcs8 and store the PK somewhere else if they want to (not sure if the PK can be downloaded at will or only once)

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/GithubAccessTokenResponseParserImpl.java
```diff
@@ -0,0 +1,60 @@
...
+    public String getName() {
+        return "GITHUB_ACCESS_TOKEN";
+    }
+
+    @Override
+    public CanonicalOAuthAccessTokenResponseDto parseTokenResponse(HttpResponse response) throws IOException {
+        AccessTokenResponseDto accessTokenResponseDto = objectMapper.readerFor(AccessTokenResponseDto.class)
+                .readValue(response.getContent());
+
+        CanonicalOAuthAccessTokenResponseDto dto = new CanonicalOAuthAccessTokenResponseDto();
```
jlorper, line 36: Could add `@Builder` to the Dto to have a nicer new

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -274,15 +275,15 @@ EOT
       target_host : "api.github.com"
       source_auth_strategy : "oauth2_refresh_token"
       secured_variables : [
-        { name : "ACCESS_TOKEN", writable : true },
```
jlorper, line 277: :red_circle: I think this is not going to work. ACCESS_TOKEN needs to exist if using `USE_SHARED_TOKEN : "TRUE"` as it will try to persist it.
If not, will be checking for the token (might fail) or generating a new token every single call I believe.

jlorper (changes requested):

jlorper:
I'm going to the changes to get this released today

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -275,9 +275,9 @@ EOT
       target_host : "api.github.com"
       source_auth_strategy : "oauth2_refresh_token"
       secured_variables : [
-        { name : "PRIVATE_KEY", writable : true },
-        { name : "REFRESH_TOKEN", writable : true },
-        { name : "OAUTH_REFRESH_TOKEN", writable : true, lockable : true },
+        { name : "ACCESS_TOKEN", writable : true }, # writable, as needs to be shared
```
jlorper, line 278: - add ACCESS_TOKEN back
- remove REFRESH_TOKEN, GH doesn't have refresh tokens, the JWT built with private key + client id acts as that and returns back the access token (that doesn't have a refresh token)

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -275,9 +275,9 @@ EOT
...
+        { name : "PRIVATE_KEY", writable : false },
+        { name : "OAUTH_REFRESH_TOKEN", writable : true, lockable : true }, # I believe this is not needed anymore
```
jlorper, line 280: Also, locks are independent now, right? No need to declare this here. Leaving as can't really test right now.

jlorper (approved):

PR #428, opened 2023-07-14
S153 misc fixes

aperez-worklytics:
Two issues:
- If google connectors are not enabled, TF will fail while generating TODOs due a NPE. Adding a try to prevent that.
- `google_workspace_gcp_project_id` is always required, no matter if it needs to be used. Added a validation rule similar as https://github.com/Worklytics/psoxy/pull/420 to ensure is only required if there are some g-workspace connector enabled.

### Fixes
[problem where it expects at least 1 non google connector, or max failes in worklytics-connector-specs](https://app.asana.com/0/1201039336231823/1205040992501143)

### Change implications


infra/examples-dev/gcp/google-workspace-variables.tf
```diff
@@ -25,3 +26,14 @@ variable "google_workspace_example_admin" {
   description = "user to impersonate for Google Workspace API calls (null for value of `google_workspace_example_user`)"
   default     = null # will failover to user
 }
+
+locals {
+  # tflint-ignore: terraform_unused_declarations
+  validate_google_workspace_gcp_project_id         = (var.google_workspace_gcp_project_id == null || var.google_workspace_gcp_project_id == "") && (contains(var.enabled_connectors, "gdirectory") || contains(var.enabled_connectors, "gcal") || contains(var.enabled_connectors, "gmail") || contains(var.enabled_connectors, "google-chat") || contains(var.enabled_connectors, "google-meet") || contains(var.enabled_connectors, "gdrive"))
+  validate_google_workspace_gcp_project_id_message = "The google_workspace_gcp_project_id var should be populated if a Google Workspace connector is enabled."
+  validate_google_workspace_gcp_project_id_check = regex(
```
jlorper, line 34: I don't really follow this kind of checks... matching regexp on message, I guess is to make it fail.
Probably this could work and is easier (to follow and understand, imho)

`validate_google_workspace_gcp_project_id_check = tobool(local.validate_google_workspace_gcp_project_id ? "true" : "The google_workspace_gcp_project_id var should be populated if a Google Workspace connector is enabled.")`

> Only boolean values, null, and the exact strings "true" and "false" can be converted to boolean. All other values will produce an error.

jlorper (approved):
A couple of suggestions

infra/examples-dev/gcp/google-workspace-variables.tf
```diff
@@ -1,6 +1,7 @@
 variable "google_workspace_gcp_project_id" {
   type        = string
   description = "string ID of GCP project that will host oauth clients for Google Workspace API connectors; must exist"
+  default = null
```
eschultink, line 4: in the case that it *isn't* used, then they should be deleting this whole file

infra/examples-dev/gcp/google-workspace-variables.tf
eschultink: yeah, this seems too hacky for me. It's hard to read the conditions, and doesn't really stop TF from working.

infra/modules/worklytics-connectors/outputs.tf
```diff
@@ -21,4 +21,4 @@ output "todos" {
 
 output "next_todo_step" {
   value = max(values(module.source_token_external_todo)[*].next_todo_step...)
```
eschultink, line 23: this needed `try()` to, so I just ported both to https://github.com/Worklytics/psoxy/pull/432

infra/examples-dev/aws-all/google-workspace-variables.tf
```diff
@@ -1,6 +1,7 @@
 variable "google_workspace_gcp_project_id" {
   type        = string
   description = "string ID of GCP project that will host oauth clients for Google Workspace API connectors; must exist"
+  default = null
```
eschultink, line 4: I don't think having the default is great here, bc leaving it empty isn't the proper solution. for real use, you should just delete this file from your configuration if not using Google Workspace sources - that's why split into separate file.

infra/examples-dev/gcp/google-workspace-variables.tf
aperez-worklytics: Yes, the purpose is to make TF to fail if the condition is not meet:
image

PR #441, opened 2023-07-19
pass-through `Link` HTTP Header

aperez-worklytics:
GH uses [Link header](https://docs.github.com/en/enterprise-cloud@latest/rest/guides/using-pagination-in-the-rest-api?apiVersion=2022-11-28#using-link-headers) for audit log. It seems something standard, so adding it as part of default headers to include.

### Features
[Psoxy support for link header](https://app.asana.com/0/0/1205095768432261)

### Change implications

 - dependencies added/changed? **yes (explain) / **yes** `org.apache.httpcomponents.core` 4.4.16 to `org.apache.httpcomponents.core5` to 5.2.2
   - anything that will show up in `terraform plan`/`apply` that isn't obviously a no-op? **no**

### DIFF
--- a/java/core/psoxy-core.iml
+++ b/java/core/psoxy-core.iml
@@ -42,6 +42,7 @@
     <orderEntry type="library" name="Maven: com.google.http-client:google-http-client:1.43.3" level="project" />
     <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpclient:4.5.14" level="project" />
     <orderEntry type="library" name="Maven: commons-logging:commons-logging:1.2" level="project" />
+    <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpcore:4.4.16" level="project" />
     <orderEntry type="library" name="Maven: io.opencensus:opencensus-api:0.31.1" level="project" />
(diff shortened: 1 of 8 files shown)


eschultink (approved):
OK, but I think there's a constant for `Link`; (maybe not in version of `HttpHeaders` lib that we have linked? could upgrade if easy, if not just comment that we should fix later)

java/core/psoxy-core.iml
```diff
@@ -42,6 +42,7 @@
     <orderEntry type="library" name="Maven: com.google.http-client:google-http-client:1.43.3" level="project" />
     <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpclient:4.5.14" level="project" />
     <orderEntry type="library" name="Maven: commons-logging:commons-logging:1.2" level="project" />
+    <orderEntry type="library" name="Maven: org.apache.httpcomponents:httpcore:4.4.16" level="project" />
```
aperez-worklytics, line 45: No idea why intelliJ does that, is not used in any place.

eschultink: remove it?

eschultink (approved):

PR #444, opened 2023-07-23
S153 salesforce query rules

aperez-worklytics:
- Adding support for Salesforce SOQL pagination.

The original request can be a SOQL to a single entity (user, account) and for retrieving activity histories based on the account id. There are rules to ensure no other query is invoked and the response is modeled to avoid including more properties than the expected ones for each case.

In case of pagination, SOQL returns an URL like *.../query/PAGE_TOKEN*. There are no way to determine if token belongs to an user, account or activity history part. For that, it has been added support for apply transformation rules conditionally and JSON Schema condition and group validation. See comments in the PR.

It has been also deleted some rules that are not being used with the app implementation.

### Fixes
[Salesforce: Missing rule for pagination](https://app.asana.com/0/1204983368671064/1205101737839878)

### Change implications


docs/sources/api-response-examples/salesforce/account_id_query.json
aperez-worklytics, line 1: Deleted endpoint and related files, as is not being used from app.

docs/sources/api-response-examples/salesforce/sanitized/accounts_by_query.json
```diff
@@ -3,6 +3,9 @@
   "done":true,
   "records":[
     {
+      "attributes":{
+        "type":"Account"
```
aperez-worklytics, line 7: Sanitization produces same results than we had before, with the difference that the `attributes` object is present -required for doing the json validation for transformation-. They were redacted in the past not for security stuff, just because they are not being used in the caller.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -191,6 +192,15 @@ Object applyTransform(Transform transform, Object document ) {
                 .map(JsonPath::compile)
                 .collect(Collectors.toList()));
 
+        if (transform.getApplyOnlyWhen() != null) {
+            Object filterResult = JsonPath.compile(transform.getApplyOnlyWhen()).read(document);
```
aperez-worklytics, line 196: Support apply this transformation only when the condition is satisfied. The scenario is the following:
- For Users, the id needs to be pseudonymized
- For accounts, id is not pseudonymized.

From the query pagination response, it means that:
- If `attributes.type == User` the user rules needs to be applied for transformation
- If `attributes.type == Account` the account rules needs to be applied for transformation

And this is resolved checking the json path expression set no `applyOnlyWhen`, for example:
```
$..records[?(@.attributes.type == \"User\")]
```

So the transformation will be applied if the json path conditional expression is matched with something (aka: there are NO empty results after applying that rule)

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilterUtils.java
```diff
@@ -65,9 +65,79 @@ private Object filterBySchema(String path, JsonNode provisionalOutput, JsonSchem
...
+                    filtered.put(key, filteredValue);
+                }
+            });
+
+            return filtered;
         } else if (schema.hasType()) {
-            //must have explicit type
 
+            if (schema.hasIf()) {
+                Object conditionResult = filterBySchema(path, provisionalOutput,
```
aperez-worklytics, line 116: Adding conditional rules following the [JSON standard spec](https://json-schema.org/understanding-json-schema/reference/conditionals.html#id6). 
If a schema contains an if, it will check the expression. At the moment, `constant` has been added to included in the check as it is the only currently required. Condition can return anything or an instance of `NotMatchedConstant` which in this case, the parent should handle about what to do

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilterUtils.java
```diff
@@ -311,6 +402,48 @@ public boolean hasType() {
...
+    @NoArgsConstructor
+    public static class ConditionJsonSchema extends JsonSchemaFilter {
     }
 
+    @SuperBuilder(toBuilder = true)
+    @NoArgsConstructor
+    public static class ThenJsonSchema extends JsonSchemaFilter {
+    }
+
+    private static class NotMatchedConstant {
```
aperez-worklytics, line 442: Not clean, but I need something *defined* to propagate in the object structure to differentiate between an object with json content or the condition not matched.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilterUtils.java
```diff
@@ -65,9 +65,79 @@ private Object filterBySchema(String path, JsonNode provisionalOutput, JsonSchem
...
+                    conditionResult = filterBySchema(path, provisionalOutput, schema.get_then(),
+                            root,
+                            redactionsMade);
+                }
+
+                return conditionResult;
+            }
+
+            if (schema.hasConstant()) {
+                return schema.getConstant().equals(provisionalOutput.asText()) ? "" : NotMatchedConstant.getInstance();
```
aperez-worklytics, line 137: Just the constant expression to match

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilterUtils.java
```diff
@@ -65,9 +65,79 @@ private Object filterBySchema(String path, JsonNode provisionalOutput, JsonSchem
                 //cases like URLs relative to schema URI are not supported
                 throw new RuntimeException("unsupported ref: " + schema.getRef());
             }
+        } else if (schema.hashOneOf()) {
+            // Get first schema with matches its inner condition.
```
aperez-worklytics, line 69: Following [oneOf](https://json-schema.org/understanding-json-schema/reference/combining.html#oneof), just a set of children with conditions. The first one which match it, the first one used to get the schema

java/core/src/main/java/co/worklytics/psoxy/rules/salesforce/PrebuiltSanitizerRules.java
```diff
@@ -245,10 +310,105 @@ private static JsonSchemaFilterUtils.JsonSchemaFilter jsonSchemaForQueryResult(J
...
+                    put("done", JsonSchemaFilterUtils.JsonSchemaFilter.builder()
+                            .type("boolean")
+                            .build());
+                    put("nextRecordsUrl", JsonSchemaFilterUtils.JsonSchemaFilter.builder()
+                            .type("string")
+                            .build());
+                    put("records", JsonSchemaFilterUtils.JsonSchemaFilter.builder()
+                            .type("array")
+                            .items(JsonSchemaFilterUtils.JsonSchemaFilter.builder()
+                                    .oneOf(Arrays.asList(
```
aperez-worklytics, line 345: This is the model for query pagination response, something like:

```
oneoff : [
{
if: is user
then: get the user response model
},
{
if: is account/activityhistory
then: get the account/activityhistory response model
}]
```
 
In this case, account/activityhistory have merged rules as the queries are:
- accounts: get all the accounts
- activity history: get the accounts with activity history, as an inner query

java/core/src/main/java/co/worklytics/psoxy/rules/salesforce/PrebuiltSanitizerRules.java
```diff
@@ -27,122 +24,168 @@ public class PrebuiltSanitizerRules {
                     .jsonPath("$..Email")
                     .jsonPath("$..Name")
                     .jsonPath("$..Username")
+                    .applyOnlyWhen("$..records[?(@.attributes.type == \"User\")]")
```
eschultink, line 27: why `..` not just `.`?  Haven't profiled it, I just assume `..` is pretty inefficient as it needs to recursively check the whole hierarchy for matching nodes, not just one level.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
eschultink: OK for now, but you're assuming here that compilation of the JSON path transform is cheap.  I'm ok with it, bc should avoid premature optimization.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -191,6 +192,15 @@ Object applyTransform(Transform transform, Object document ) {
...
+
+            ArrayList<?> results = (ArrayList<?>) filterResult;
+            if (results != null && results.size() == 0) {
+                return document;
```
eschultink, line 200: style-wise, I prefer avoiding multiple paths that `return` within a method.

i think more readable to more explicitly separate what you're doing by either having a `boolean transformApplies` that you compute the value of, or by splitting into two methods: 1) the method that actually applies the transformation, and 2) a method that evaluates the condition, and either returns directly or returns result of calling method (1).

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -191,6 +192,15 @@ Object applyTransform(Transform transform, Object document ) {
                 .map(JsonPath::compile)
                 .collect(Collectors.toList()));
 
+        if (transform.getApplyOnlyWhen() != null) {
+            Object filterResult = JsonPath.compile(transform.getApplyOnlyWhen()).read(document);
+
+            ArrayList<?> results = (ArrayList<?>) filterResult;
+            if (results != null && results.size() == 0) {
```
eschultink, line 199: some form of `CollectionUtils.isEmpty()` - null-safe size check - would make this condition clearer/more readable, if we have such a library method linked here.

eschultink (approved):
no blockers, some thoughts on style/structure

aperez-worklytics (commented):
Thanks!

PR #447, opened 2023-07-24
S154 disabling old gcp secrets

aperez-worklytics:
Only for GCP: everytime a secret is added to the config, a new version is added in the secret manager. The problem is every new version added has a cost. Disabling the older versions will avoid for paying versions are not used.

### Fixes
- (bug: zoom excessive token rotation)[bug: zoom excessive token rotation](https://app.asana.com/0/1204736724839176/1205115368713792)

### Change implications


infra/modules/gcp-host/main.tf
```diff
@@ -95,13 +95,13 @@ resource "google_secret_manager_secret_iam_member" "grant_sa_updater_on_lockable
...
 
-resource "google_secret_manager_secret_iam_member" "grant_sa_secretVersionAdder_on_writable_secret" {
+resource "google_secret_manager_secret_iam_member" "grant_sa_secretVersionManager_on_writable_secret" {
   for_each = { for secret in local.writable_secrets : secret.instance_secret_id => secret }
 
   project   = var.gcp_project_id
   secret_id = "${local.config_parameter_prefix}${each.value.instance_secret_id}"
   member    = "serviceAccount:${google_service_account.api_connectors[each.value.instance_id].email}"
-  role      = "roles/secretmanager.secretVersionAdder"
+  role      = "roles/secretmanager.secretVersionManager"
```
aperez-worklytics, line 104: That role supports version handling, required for disabling/destroying versions -> https://cloud.google.com/secret-manager/docs/access-control

```
resourcemanager.projects.get
resourcemanager.projects.list
secretmanager.versions.add
secretmanager.versions.destroy
secretmanager.versions.disable
secretmanager.versions.enable
secretmanager.versions.get
secretmanager.versions.list
```

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -71,6 +72,25 @@ public void putConfigProperty(ConfigProperty property, String value) {
                 SecretVersion version = client.addSecretVersion(secretName, payload);
 
                 log.info(String.format("Property: %s, stored version %s", secretName, version.getName()));
+
+                SecretManagerServiceClient.ListSecretVersionsPage enabledVersions = client.listSecretVersions(ListSecretVersionsRequest.newBuilder()
+                                .setFilter("state:ENABLED")
+                                .setParent(secretName.toString())
+                                // Reduce the page, as each version will be disabled one by one
+                                .setPageSize(20)
```
aperez-worklytics, line 80: Just the first page to avoid doing too much in a single request

infra/modules/gcp-host/main.tf
eschultink: 👍

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
eschultink: although affected customer won't require this, I think worth happening in case a DISABLE request fails or something in future; this gives some robustness against versions being left behind and gradually increasing costs

eschultink (approved):

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -71,6 +72,25 @@ public void putConfigProperty(ConfigProperty property, String value) {
...
+                                .build())
+                        .getPage();
+
+                // Disable secret version, just the first page
+                enabledVersions.getValues().forEach(i -> {
+                    if (!i.getName().equals(version.getName())) {
+                        client.disableSecretVersion(DisableSecretVersionRequest.newBuilder()
+                                .setName(i.getName())
+                                .build());
```
jlorper, line 89: I'd wrap this in a try-catch or better yet extract a method `disableSecretVersions(SecretVersion versionToKeep)` with proper error handling just to make sure that a failure disabling won't affect the outcome of the method (specially if the new version was successfully added)

PR #448, opened 2023-07-25
fix URL_SAFE_TOKEN format for bulk case

eschultink:
### Fixes
 - `URL_SAFE_TOKEN` format for bulk case doesn't behave as expected

### Change implications

    - use of `URL_SAFE_TOKEN` format for `pseudonymize` transforms in REST rules w/o `includeReversible`, will result in backwards-incompatible format. (this is not a case we believe anyone using in practice; the default rules don't do it, so even those people who have customized it would have had to explicitly change format for some reason)


jlorper (approved):
OK, remember to update the library dependency in Worklytics

PR #449, opened 2023-07-26
token streamline

eschultink:
### Features
 - cleanup some token refresh logic, to be more direct

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
@@ -198,7 +198,27 @@ public class OAuthRefreshTokenSourceAuthStrategy implements SourceAuthStrategy {
         // open a lambda waiting for a lock
         private static final String TOKEN_REFRESH_LOCK_ID = "oauth_refresh_token";
         private static final int MAX_TOKEN_REFRESH_ATTEMPTS = 3;
-        private static final long WAIT_AFTER_FAILED_LOCK_ATTEMPTS = 2000L;
+
+        /**
+         * how long to wait after a failed lock attempt before trying again; multiplier on the
+         * attempt.
+         *
+         * goal is that this value should be big enough to allow the process that holds the lock to
+         *  1) refresh the token it
+         *  2) write it
+         *  3) release the lock
+         *  4) have that write be visible to other processes, given eventual consistency in GCP
+         *     Secret Manager case
+         *
+         * this includes the allowance for eventual consistency, so wait should be >= that value
+         * in practice.
+         */
+        private static final long WAIT_AFTER_FAILED_LOCK_ATTEMPTS = 4000L;
+
+        /**
+         * how long to allow for eventual consistency after write to config
+         */
+        private static final long ALLOWANCE_FOR_EVENTUAL_CONSISTENCY_SECONDS = 2L;
 
         /**
          * implements canonical oauth flow to exchange refreshToken for accessToken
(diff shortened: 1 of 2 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -199,8 +200,6 @@ public static class TokenRefreshHandlerImpl implements OAuth2CredentialsWithRefr
         private static final int MAX_TOKEN_REFRESH_ATTEMPTS = 3;
         private static final long WAIT_AFTER_FAILED_LOCK_ATTEMPTS = 2000L;
 
-        private AccessToken currentToken = null;
```
eschultink, line 202: new instance of `TokenRefreshHandlerImpl` created for each request; so what's the point of storing `currentToken` here? 

risk is that in 401 case we still have the `currentToken` that gave the 401, so we'll enter a loop with it.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -220,19 +219,10 @@ private AccessToken refreshAccessToken(int attempt) throws IOException {
...
+            AccessToken token = getSharedAccessTokenIfSupported().orElse(null);
 
-            if (sharedAccessToken.isPresent()) {
-                // we have a token, but shared token is newer. Other instance refreshed, so use it
-                if (sharedAccessToken.get().getExpirationTime().after(this.currentToken.getExpirationTime())) {
-                    this.currentToken = sharedAccessToken.get();
-                }
-            }
 
-            if (shouldRefresh(this.currentToken, clock.instant())) {
```
eschultink, line 235: we have a risk here, bc of proactive refresh, that another process has done this, but we still see stale `sharedAccessToken`

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -248,15 +238,17 @@ private AccessToken refreshAccessToken(int attempt) throws IOException {
...
 
                 tokenResponse = exchangeRefreshTokenForAccessToken();
-                this.currentToken = asAccessToken(tokenResponse);
-                storeSharedAccessTokenIfSupported(this.currentToken);
+                token = asAccessToken(tokenResponse);
+
+                storeSharedAccessTokenIfSupported(token);
 
                 if (lockNeeded) {
+                    //q: hold lock extra long, to wait for consistency??
```
eschultink, line 246: to limit risk of stale read, should we hold lock extra long here?

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -220,19 +219,10 @@ private AccessToken refreshAccessToken(int attempt) throws IOException {
...
 
-            if (sharedAccessToken.isPresent()) {
-                // we have a token, but shared token is newer. Other instance refreshed, so use it
-                if (sharedAccessToken.get().getExpirationTime().after(this.currentToken.getExpirationTime())) {
-                    this.currentToken = sharedAccessToken.get();
-                }
-            }
 
-            if (shouldRefresh(this.currentToken, clock.instant())) {
+            if (token == null || shouldRefresh(token, clock.instant())) {
```
jlorper, line 225:
```suggestion
            if (shouldRefresh(token, clock.instant())) {
```
revert, `shouldRefresh` handles null case

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
jlorper: I just found another issue here, in the previous lines:
```
                if (!acquired) {
                    //NOTE: check shared access token again, in case the instance which held the
                    // lock refreshed the token
                    Uninterruptibles.sleepUninterruptibly(attempt * WAIT_AFTER_FAILED_LOCK_ATTEMPTS, TimeUnit.MILLISECONDS);
 --------->       return refreshAccessToken(attempt + 1); 
                }
```
missing `return` in the recursive call, so might exchange again after a successful one or releasing a lock that might be in progress. Potential inconsistencies.

Said this, adding some random wait before releasing the lock is not a bad idea.

jlorper (approved):
OK, but found one issue that needs to be addressed @eschultink

PR #450, opened 2023-07-26
sha256 encoder

eschultink:
### Features
 - a Sha256 encoder, implementing defacto encoding done by proxy in bulk mode when format set to `URL_SAFE_TOKEN`

### Change implications

### DIFF
--- a/java/gateway-core/src/main/java/com/avaulta/gateway/pseudonyms/impl/Sha256PseudonymEncoder.java
+++ b/java/gateway-core/src/main/java/com/avaulta/gateway/pseudonyms/impl/Sha256PseudonymEncoder.java
@@ -1,0 +1,43 @@
+package com.avaulta.gateway.pseudonyms.impl;
+
+import com.avaulta.gateway.pseudonyms.Pseudonym;
+import com.avaulta.gateway.pseudonyms.PseudonymEncoder;
+import org.apache.commons.lang3.StringUtils;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+
+/**
+ * implementation of defacto encoding used by BulkDataSanitizerImpl as of v0.4.30
+ */
+public class Sha256PseudonymEncoder implements PseudonymEncoder {
+
+
+    @Override
+    public String encode(Pseudonym pseudonym) {
+        return base64Encode(pseudonym.getHash());
+    }
+
+    @Override
+    public Pseudonym decode(String input) {
+        return Pseudonym.builder()
+            .hash(input.getBytes(StandardCharsets.UTF_8))
+            .build();
+    }
+
+    @Override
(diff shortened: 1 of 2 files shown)


java/gateway-core/src/main/java/com/avaulta/gateway/pseudonyms/impl/Sha256PseudonymEncoder.java
```diff
@@ -0,0 +1,43 @@
...
+    public Pseudonym decode(String input) {
+        return Pseudonym.builder()
+            .hash(input.getBytes(StandardCharsets.UTF_8))
+            .build();
+    }
+
+    @Override
+    public boolean canBeDecoded(String possiblePseudonym) {
+        return possiblePseudonym != null &&
+            possiblePseudonym.getBytes(StandardCharsets.UTF_8).length == 43; //43 rather than 32, bc of base64 encoding
```
jlorper, line 31:
```suggestion
            possiblePseudonym.getBytes(StandardCharsets.UTF_8).length == 43; //43 rather than 32, bc of base64 encoding without padding
```

jlorper: could also go a bit further and try to check that input contains valid base64 characters, but I think there's no easy way of doing it with the jre library (rather than try to decode and catch)

jlorper (changes requested):

eschultink:
@jlorper  it's not used in `ColumnarBulkDataSanitizerImpl` anymore; it represents the legacy bad behavior (directly encoding JUST the hash) when set to `URL_SAFE_TOKEN`; I created this class to bind into worklytics codebase to be able to parse this legacy case. So we'll use it from there.

jlorper (approved):

java/gateway-core/src/main/java/com/avaulta/gateway/pseudonyms/impl/Sha256PseudonymEncoder.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    public String encode(Pseudonym pseudonym) {
+        return base64Encode(pseudonym.getHash());
+    }
+
+    @Override
+    public Pseudonym decode(String input) {
+        return Pseudonym.builder()
+            .hash(input.getBytes(StandardCharsets.UTF_8))
```
jlorper, line 24: doesn't this need to be decoded??
```suggestion
            .hash(base64decoder.decode(input.getBytes(StandardCharsets.UTF_8)))
```

jlorper: just made up the `base64decoder` part

PR #451, opened 2023-07-26
S154 secrets updated too much

aperez-worklytics:
**NOTE** #449 has been merged to my branch.

Only for GCP. 

- Secrets were exposed as env vars in the function, so when they were read they go through EnvVarConfigService and not from SecretManagerConfigService. As env vars, we have set that the version to use is always `latest`
- implemented the solution to put the label with the version of the secret to use in order to avoid issues with eventual consistency. That includes:
  - Using a specific roles for secrets (instead of 2 (locker and regular ones), as now for any secret we could read it, update it (labels), reading version content, etc. 
  - On `SecretManagerConfigService`, adding support of update the label everytime a new version is added.

### Fixes
[GCP: More than one new token version when refreshing](https://app.asana.com/0/1205115400675822/1205136895082281)

### Change implications

### DIFF
--- a/infra/modules/gcp-host/main.tf
+++ b/infra/modules/gcp-host/main.tf
@@ -55,15 +55,6 @@ locals {
       var_def)
     }
   }
-  /*lockable_secrets = flatten([
-    for instance_id, secrets in local.secrets_to_provision :
-    [for secret_id, secret in values(secrets) : secret if secret.lockable]
-  ])
-
-  writable_secrets = flatten([
-    for instance_id, secrets in local.secrets_to_provision :
-    [for secret_id, secret in values(secrets) : secret if secret.writable]
-  ])*/
 
   updateable_secrets = flatten([
   for instance_id, secrets in local.secrets_to_provision :
@@ -86,19 +77,6 @@ module "secrets" {
   default_labels    = var.default_labels
   replica_locations = local.secret_replica_locations
 }
-/*
-resource "google_secret_manager_secret_iam_member" "grant_sa_updater_on_lockable_secrets" {
-  for_each = { for secret in local.lockable_secrets : secret.instance_secret_id => secret }
-
-  project   = var.gcp_project_id
-  secret_id = "${local.config_parameter_prefix}${each.value.instance_secret_id}"
-  member    = "serviceAccount:${google_service_account.api_connectors[each.value.instance_id].email}"
-  role      = module.psoxy.psoxy_instance_secret_role_id
-
-  depends_on = [
-    module.secrets
-  ]
-}*/
 
 resource "google_secret_manager_secret_iam_member" "grant_sa_secretVersionManager_on_writable_secret" {
   for_each = { for secret in local.updateable_secrets : secret.instance_secret_id => secret }
@@ -165,7 +143,7 @@ module "api_connector" {
 
   secret_bindings = merge(
     # bc some of these are later filled directly, bind to 'latest'
-    { for k, v in module.secrets[each.key].secret_bindings : k => merge(v, { version_number : "latest" }) },
+   {for k, secrets in local.secrets_to_provision[each.key] : k => merge({ secret_id : module.secrets[each.key].secret_ids_within_project[k], version_number : "latest" }) if (try(!secrets.lockable, true) && try(!secrets.writable, true))},
     module.psoxy.secrets
   )
 }
--- a/infra/modules/gcp-host/main.tf
+++ b/infra/modules/gcp-host/main.tf
@@ -95,13 +95,13 @@ resource "google_secret_manager_secret_iam_member" "grant_sa_updater_on_lockable
   ]
 }
 
-resource "google_secret_manager_secret_iam_member" "grant_sa_secretVersionAdder_on_writable_secret" {
+resource "google_secret_manager_secret_iam_member" "grant_sa_secretVersionManager_on_writable_secret" {
   for_each = { for secret in local.writable_secrets : secret.instance_secret_id => secret }
 
   project   = var.gcp_project_id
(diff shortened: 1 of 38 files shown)


infra/modules/gcp-host/main.tf
```diff
@@ -160,7 +143,7 @@ module "api_connector" {
 
   secret_bindings = merge(
     # bc some of these are later filled directly, bind to 'latest'
-    { for k, v in module.secrets[each.key].secret_bindings : k => merge(v, { version_number : "latest" }) },
```
aperez-worklytics, line 163: Here all secrets created were added as part of security bindings, being available as it were env vars

infra/modules/gcp-host/main.tf
```diff
@@ -160,7 +143,7 @@ module "api_connector" {
...
+   {for k, secrets in local.secrets_to_provision[each.key] : k => merge({ secret_id : module.secrets[each.key].secret_ids_within_project[k], version_number : "latest" }) if (try(!secrets.lockable, true) && try(!secrets.writable, true))},
```
eschultink, line 146: yeah, I agree with this - good change, if not very readable to do it this way 😓  - it's correct via the interface we've defined.

infra/modules/gcp-host/main.tf
```diff
@@ -55,19 +55,15 @@ locals {
       var_def)
     }
   }
-  lockable_secrets = flatten([
-    for instance_id, secrets in local.secrets_to_provision :
-    [for secret_id, secret in values(secrets) : secret if secret.lockable]
-  ])
 
-  writable_secrets = flatten([
```
eschultink, line 63: I think writable is fine; in practice, lockable should be subset of writable.

eschultink (changes requested):

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -108,7 +94,19 @@ public String getConfigPropertyOrError(ConfigProperty property) {
...
+
+            Secret secret = client.getSecret(secretName);
+
+            String versionName = secret.getLabelsMap() != null ? secret.getLabelsMap().get(VERSION_LABEL) : null;
+
+            if (StringUtils.isBlank(versionName)) {
+                versionName = "latest";
+            }
+
+            log.severe("Reading secret version: " + versionName);
```
eschultink, line 107: don't make this `severe`

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -194,6 +182,48 @@ public void release(String lockId) {
...
+                            .setName(secretName.toString())
+                            // Label format is https://cloud.google.com/compute/docs/labeling-resources#requirements
+                            .putLabels(label, labelValue)
+                            .build())
+                    .setUpdateMask(FieldMask.newBuilder()
+                            .addPaths("labels")
+                            .build())
+                    .build());
+        } catch (Exception e) {
+            log.severe(String.format("Cannot put the label of the version on the secret %s due exception: %s", secretName.toString(), e));
```
eschultink, line 198: don't serialize exception like this; leverage `log.log(Level.SEVERE, ..., e)` to get it via the native support for that.

eschultink: also, imho you shouldn't catch this here, or should at least re-throw.  you're supressing it, so caller won't detect failure and will just continue. That's probably OK in the version labeling case, but not in the lock case - it will believe lock was acquired, right?

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -194,6 +182,48 @@ public void release(String lockId) {
...
+                if (!i.getName().equals(exceptVersion.getName())) {
+                    client.disableSecretVersion(DisableSecretVersionRequest.newBuilder()
+                            .setName(i.getName())
+                            .build());
+
+                    log.info(String.format("Property: %s, disabled version %s", secretName, i.getName()));
+                }
+            });
+        } catch (Exception e) {
+            log.severe(String.format("Cannot disable old versions from secret %s due exception: %s", secretName.toString(), e));
```
eschultink, line 223: i think leave suppressing the exception to the caller

eschultink (changes requested):

eschultink (approved):

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -194,6 +180,50 @@ public void release(String lockId) {
...
+                            .setName(i.getName())
+                            .build());
+
+                    log.info(String.format("Property: %s, disabled version %s", secretName, i.getName()));
+                }
+            });
+        } catch (Exception e) {
+            log.log(Level.SEVERE, String.format("Cannot disable old versions from secret %s", secretName.toString()), e);
+            throw e;
+        }
```
jlorper, line 224: again, disabling the secrets is a secondary action that in my option shouldn't jeopardize the main one `putConfigProperty`
access errors like not enough permissions should fail as it clearly states misconfiguration, but transient errors should be ignored. Next run will attempt to disable the secrets again.

infra/modules/gcp-host/main.tf
jlorper, line 146: I think the `if` could be
`if !contains(keys(local.writable_secrets), secret_id)`, more succinct

PR #455, opened 2023-07-28
fixes to pseudonym encoding for URL_SAFE_TOKEN in bulk mode

eschultink:
### Fixes
  - double-encoding introduced in v0.4.31 of URL_SAFE_TOKEN formatted pseudonyms in 'bulk' case
  - `Base64UrlSha256HashPseudonymEncoder` - original implementation of this expected Base64- encoded SHA-256 with some re-mapping of a few characters; but what we actually need it to deal with are standard Base64-UrlEncoded SHA-256

### Change implications

 - something to note in `CHANGELOG.md`? **v0.4.31 double-encodes bulk pseudonyms in format URL_SAFE_TOKEN; this fixes that behavior, but is breaking change**

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/HashUtils.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/HashUtils.java
@@ -12,12 +12,16 @@ import java.util.Base64;
 public class HashUtils {
 
     public String hash(String... fragments) {
+        return encode(DigestUtils.sha256(String.join("", fragments))));
+    }
+
+    String encode(byte[] bytes) {
         // No padding saves us the '=' character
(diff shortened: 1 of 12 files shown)


java/core/src/main/java/co/worklytics/psoxy/HashUtils.java
```diff
@@ -8,16 +9,26 @@
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 
+/**
+ * NOTE: this class is used for LEGACY pseudonymization implementation, not DEFAULT since v0.4
```
aperez-worklytics, line 13: Then I've annotated with `@Deprecated`

aperez-worklytics (approved):

PR #460, opened 2023-08-02
cache refreshed access tokens

eschultink:
### Features
 - make shared token setting config parsed by the RefreshTokenHandler, rather than hardcoded on TokenRefreshPayload
 - make TokenRefreshHandlerImpl singleton, synchronize refresh (GCP cloud function gen2 support concurrency - https://cloud.google.com/functions/docs/concepts/version-comparison#comparison-table) 
 - support a ACCESS_TOKEN_CACHEABLE config variable, that determines whether access token can be cached or not. 

### Change implications


infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -383,6 +383,8 @@ EOT
         GRANT_TYPE : "client_credentials"
         CREDENTIALS_FLOW : "client_secret"
         REFRESH_ENDPOINT : "https://${var.salesforce_domain}/services/oauth2/token"
+        ACCESS_TOKEN_CACHEABLE : "true",
```
jlorper, line 386: so this doesn't make a difference in any case, right?
Tokens can be locally cacheable **always**, if the container allows it, it will benefit from it. Otherwise, use the shared token is the way to go.
There is no way to have one token "locally" per function instance... So probably I'll just revert the ACCESS_TOKEN_CACHEABLE, I think in the end it just add noise and another config that won't matter much

jlorper (approved):
In my opinion, I'd revert the cacheable config and do it always.

infra/modules/worklytics-connector-specs/main.tf
eschultink: in theory, there should be. caching doesn't seem to work in GCP as the instance of `RefreshHandlerImpl` it's cached on is getting lost between requests.

if we get the object graph fixed to persist that across requests, this should fully work as expected and caching tokens should work independently of sharing tokens.

caching tokens should really be the default, yes; Zoom is the exception that has some kind of strict token rotation policy (can't find a term for this).

sharing could be the default, at the risk of cycling load on Secret Manager / Parameter Store, and creating lots of versions of the token as we saw with Zoom.

PR #461, opened 2023-08-02
S154 cache token fixes

aperez-worklytics:
Parent PR is #460.

Couple of issues fixed after trying it; see comments

### Change implications

### DIFF
--- a/infra/modules/gcp-psoxy-bulk/main.tf
+++ b/infra/modules/gcp-psoxy-bulk/main.tf
@@ -177,7 +177,7 @@ resource "google_cloudfunctions_function" "function" {
     var.path_to_config == null ? {} : yamldecode(file(var.path_to_config)),
     var.environment_variables,
     var.config_parameter_prefix == null ? {} : { PATH_TO_SHARED_CONFIG = var.config_parameter_prefix },
-    var.config_parameter_prefix == null ? {} : { PATH_TO_INSTANCE_CONFIG = "${var.config_parameter_prefix}${upper(local.instance_id)}_" },
+    var.config_parameter_prefix == null ? {} : { PATH_TO_INSTANCE_CONFIG = "${var.config_parameter_prefix}${replace(upper(var.instance_id), "-", "_")}_" },
   )
 
   dynamic "secret_environment_variables" {
--- a/infra/modules/worklytics-connector-specs/main.tf
+++ b/infra/modules/worklytics-connector-specs/main.tf
@@ -399,6 +399,13 @@ EOT
           sensitive : false
           value_managed_by_tf : false
         },
+        {
+          name : "OAUTH_REFRESH_TOKEN"
+          writable : true
+          lockable : true
+          sensitive : true
+          value_managed_by_tf : false
+        },
       ]
       reserved_concurrent_executions : null
       example_api_calls_user_to_impersonate : null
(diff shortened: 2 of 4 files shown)


infra/modules/gcp-psoxy-bulk/main.tf
```diff
@@ -177,7 +177,7 @@ resource "google_cloudfunctions_function" "function" {
     var.path_to_config == null ? {} : yamldecode(file(var.path_to_config)),
     var.environment_variables,
     var.config_parameter_prefix == null ? {} : { PATH_TO_SHARED_CONFIG = var.config_parameter_prefix },
-    var.config_parameter_prefix == null ? {} : { PATH_TO_INSTANCE_CONFIG = "${var.config_parameter_prefix}${upper(local.instance_id)}_" },
+    var.config_parameter_prefix == null ? {} : { PATH_TO_INSTANCE_CONFIG = "${var.config_parameter_prefix}${replace(upper(var.instance_id), "-", "_")}_" },
```
aperez-worklytics, line 180: Detected this -same as rest function-.

For Jira it was failing, as the `PATH_TO_INSTANCE_CONFIG` was `PSOXY_JIRA-CLOUD_` instead of `PSOXY_JIRA_CLOUD_`, causing that secrets never been found as the prefix mismatch the name of the secret created. Example:
`psoxy_JIRA-CLOUD_OAUTH_REFRESH_TOKEN` -> `psoxy_JIRA_CLOUD_OAUTH_REFRESH_TOKEN`

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -284,9 +284,11 @@ private AccessToken refreshAccessToken(int attempt) throws IOException {
                     tokenResponse = exchangeRefreshTokenForAccessToken();
                     token = asAccessToken(tokenResponse);
 
-                    storeSharedAccessTokenIfSupported(token);
                     if (isAccessTokenCacheable()) {
                         this.cachedToken = token;
+                    } else {
+                        //q: assuming that if the token is cacheable, is not present as a SECRET to be stored
+                        storeSharedAccessTokenIfSupported(token);
```
aperez-worklytics, line 291: That will fail if the secret token is not defined in secret manager / parameter store. I think your idea is to get rid off that if token is cacheable? 

This is the main change done apart from fixes.

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -759,7 +766,6 @@ EOT
         GRANT_TYPE : "refresh_token"
         REFRESH_ENDPOINT : "https://auth.atlassian.com/oauth/token"
         USE_SHARED_TOKEN : "TRUE"
-        ACCESS_TOKEN_CACHEABLE : "TRUE"
```
aperez-worklytics, line 762: Jira cloud has the `ACCESS_TOKEN` secret defined, so in theory it will not need that unless the secret is dropped...

eschultink (approved):

infra/modules/worklytics-connector-specs/main.tf
eschultink: so I think we should do both, if possible; cache in the process even if storing in Secret Manager; otherwise will have roundtrip to secret manager on every request, which we shouldn't do unless we need to

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
eschultink: sry, yeah - if we have `USE_SHARED_TOKEN`, then we need to define `ACCESS_TOKEN` secret - yes, I agree.

PR #467, opened 2023-08-04
example of simple survey rules that this version is OK with

eschultink:
not to merge; PoC of issue seen w old version + straightforward configuration of survey rules


eschultink:
confirmed fix for customer, working for them

PR #476, opened 2023-08-09
fix gcp secret access

eschultink:
### Fixes
  - not binding GCP secrets as env vars (#472), that are read-only so IAM role is `SecretAccessor` fail - bc of attempt to read `Secret` itself, which that role doesn't permit, before actual read of the version

### Change implications

### DIFF
--- a/java/impl/gcp/src/main/java/co/worklytics/psoxy/GcpModule.java
+++ b/java/impl/gcp/src/main/java/co/worklytics/psoxy/GcpModule.java
@@ -5,10 +5,7 @@ import co.worklytics.psoxy.gateway.ConfigService;
 import co.worklytics.psoxy.gateway.HostEnvironment;
 import co.worklytics.psoxy.gateway.LockService;
 import co.worklytics.psoxy.gateway.ProxyConfigProperty;
-import co.worklytics.psoxy.gateway.impl.BlindlyOptimisticLockService;
-import co.worklytics.psoxy.gateway.impl.CompositeConfigService;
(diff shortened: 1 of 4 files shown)


java/impl/gcp/src/main/java/co/worklytics/psoxy/GcpModule.java
```diff
@@ -79,10 +83,15 @@ static ConfigService nativeConfigService(EnvVarsConfigService envVarsConfigServi
...
+        SecretManagerConfigService shared = secretManagerConfigServiceFactory.create(ServiceOptions.getDefaultProjectId(), pathToSharedConfig);
+
+        Duration proxyInstanceConfigCacheTtl = Duration.ofMinutes(5);
+        Duration sharedConfigCacheTtl = Duration.ofMinutes(20);
         return CompositeConfigService.builder()
-                .preferred(instanceConfigService)
-                .fallback(secretManagerConfigServiceFactory.create(ServiceOptions.getDefaultProjectId(), pathToSharedConfig))
-                .build();
+            .preferred(new CachingConfigServiceDecorator(instanceConfigService, proxyInstanceConfigCacheTtl))
+            .fallback(new CachingConfigServiceDecorator(shared, sharedConfigCacheTtl))
```
eschultink, line 93: GCP wasn't actually using a caching config service, probably bc I expected most stuff we wanted to cache to be bound as env vars (which prior to v0.4.33, they are)

aperez-worklytics (approved):

PR #484, opened 2023-08-11
S155 fix jira rules

aperez-worklytics:
A missing rule in response metadata without items defined could generate a NPE in older versions.

### Fixes
- [peterpan jira items failed](https://app.asana.com/0/1201332175930266/1205243271503608)

### Change implications


eschultink (approved):
ok, schema stuff pretty verbose; we probably need worklytics-side code to generate using `#ref` to make it a bit more abstract

aperez-worklytics:
> ok, schema stuff pretty verbose; we probably need worklytics-side code to generate using `#ref` to make it a bit more abstract

Yeah, we could use definitions and refs there

PR #490, opened 2023-08-15
compression test

eschultink:
### Fixes
  - passing `-z` option alone doesn't get gzipp'd response from GCP, bc `User-Agent` not sent that GCP recognizes as accepting gzip'd content.
  - Better auth error feedback: AWS (MFA) expired session

### Change implications

---
  - https://app.asana.com/0/0/1205319886373301
  - https://app.asana.com/0/0/1205274265612040
  - https://app.asana.com/0/1201039336231823/1205205544108360/f


davidfq (commented):
Added [2c26a77](https://github.com/Worklytics/psoxy/pull/490/commits/2c26a7732a14fcdb77a25920d187a301c49a1938); could you try, plz? 
I want to make more tests and implement the `-z` as default.

tools/psoxy-test/lib/utils.js
```diff
@@ -64,8 +65,9 @@ function getFileNameFromURL(url, timestamp = true, extension = '.json') {
  */
 function getCommonHTTPHeaders(options = {}) {
   const headers = {
-    'Accept-Encoding': options.gzip ? 'gzip' : '*',
+    'Accept-Encoding': options.gzip ? 'gzip,deflate' : '*',
     'X-Psoxy-Skip-Sanitizer': options.skip?.toString() || 'false',
+    'User-Agent': 'psoxy-test (gzip)', // w/o (gzip) here, GCP Cloud functions don't compress response
```
jlorper, line 70: weird, how did you discover this? and also, is this happening to regular requests from GAE??? are we sure we are compressing the responses in GCP?

eschultink: yes, i was testing everything as had a cost concern from a customer, and reducing network would be quick win - but it's OK.  

GCP handles compression auto-magically for cloud functions as long as the `Accepts` and `User-Agent` follow what they expect; on the Worklytics-side we already have something in the client factory that's adding an `HttpRequestInitializer` which adds these headers for proxy connections to GCP.

Separately, I've added code on the proxy-side to add warnings to both logs and response headers if compression is NOT requested by the client.

eschultink:
@davidfq - yeah, seems to work OK with AWS, requesting compressed content; see below

psoxy

davidfq:
@eschultink I've added "deflate" support (also sent as accepted) and changed the error display. Could you please re-test this? (both AWS/GCP, I don't currently have any Psoxy instance completely configured).

davidfq (approved):

PR #491, opened 2023-08-15
S155 headers passtrough

aperez-worklytics:
- Add support for defining headers as part of the spec

NOTES: Tested in GCP, working. Not yet in AWS as I'm facing with [this error](https://github.com/Worklytics/psoxy/pull/481#discussion_r1291335164)

### Features
- [Proxy: headers not invoked on request in proxy?](https://app.asana.com/0/1205115400675822/1205114067549132)

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
@@ -31,10 +31,7 @@ import org.apache.http.entity.ContentType;
 import javax.inject.Inject;
 import java.io.IOException;
 import java.net.URL;
-import java.util.Locale;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.regex.Pattern;
@@ -193,10 +190,7 @@ public class CommonRequestHandler {
         }
 
         //TODO: what headers to forward???
-        sourceApiRequest.setHeaders(sourceApiRequest.getHeaders()
-                //seems like Google API HTTP client has a default 'Accept' header with 'text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2' ??
-                .setAccept(ContentType.APPLICATION_JSON.toString())  //MSFT gives weird "{"error":{"code":"InternalServerError","message":"The MIME type 'text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2' requires a '/' character between type and subtype, such as 'text/plain'."}}
-        );
+        sourceApiRequest.setHeaders(getHeadersFromSource(request));
 
         //setup request
         sourceApiRequest
@@ -403,6 +397,21 @@ public class CommonRequestHandler {
         if (config.isDevelopment()) {
             log.info(messageSupplier.get());
         }
+    }
+
+    private com.google.api.client.http.HttpHeaders getHeadersFromSource(HttpEventRequest request) {
+        com.google.api.client.http.HttpHeaders headers = new com.google.api.client.http.HttpHeaders();
+
+        //seems like Google API HTTP client has a default 'Accept' header with 'text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2' ??
+        headers.setAccept(ContentType.APPLICATION_JSON.toString());  //MSFT gives weird "{"error":{"code":"InternalServerError","message":"The MIME type 'text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2' requires a '/' character between type and subtype, such as 'text/plain'."}}
+
+        config.getConfigPropertyAsOptional(ProxyConfigProperty.SUPPORTED_REQUEST_HEADERS)
+                .ifPresent(i -> Arrays.stream(i.split(" "))
+                        .forEach(h -> {
+                            request.getHeader(h).ifPresent(headerValue -> headers.set(h, headerValue));
+                        }));
+
+        return headers;
     }
 
 }
--- a/infra/examples-dev/aws-all/google-workspace.tf
+++ b/infra/examples-dev/aws-all/google-workspace.tf
@@ -8,7 +8,7 @@ provider "google" {
 
 module "worklytics_connectors_google_workspace" {
   source = "../../modules/worklytics-connectors-google-workspace"
-  # source = "git::https://github.com/worklytics/psoxy//infra/modules/worklytics-connectors-google-workspace?ref=v0.4.34"
+  # source = "git::https://github.com/worklytics/psoxy//infra/modules/worklytics-connectors-google-workspace?ref=rc-v0.4.35"
 
   providers = {
     google = google.google_workspace
--- a/infra/examples-dev/aws-all/main.tf
+++ b/infra/examples-dev/aws-all/main.tf
@@ -32,7 +32,7 @@ terraform {
 # general cases
 module "worklytics_connectors" {
   source = "../../modules/worklytics-connectors"
-  # source = "git::https://github.com/worklytics/psoxy//infra/modules/worklytics-connectors?ref=v0.4.34"
+  # source = "git::https://github.com/worklytics/psoxy//infra/modules/worklytics-connectors?ref=rc-v0.4.35"
 
   enabled_connectors            = var.enabled_connectors
   jira_cloud_id                 = var.jira_cloud_id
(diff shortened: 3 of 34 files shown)


eschultink (changes requested):
for consistency, should be part of `Rules` - similar to allowed HTTP methods, allowed query params, etc. 

headers are just another type of parameter. 

OpenAPI spec is too verbose, imho - we should just have an allow list - but has the idea that header analogous to any other type of parameter that might be accepted by the API.
https://swagger.io/docs/specification/describing-parameters/#header-parameters

configuring on Terraform is brittle.

docs/sources/example-rules/dropbox-business/dropbox-business.yaml
```diff
@@ -31,6 +31,8 @@ endpoints:
           - "$.events[*].assets[*]['display_name', 'doc_title', 'folder_title', 'showcase_title']"
           - "$.events[*].assets[*].path['contextual']"
           - "$.events[*].assets[*].path.namespace_relative['relative_path']"
+    supportedHeaders:
+      - "Dropbox-API-Select-User"
```
aperez-worklytics, line 35: While testing this on dropbox, I've discovered that the connector is not working; something related with tokens. I'll fix it in a different PR.

docs/sources/example-rules/salesforce/salesforce.yaml
```diff
@@ -131,6 +133,8 @@ endpoints:
           - "$.records[*].OwnerId"
         applyOnlyWhen: "$.records[?(@.attributes.type == \"Account\")]"
         encoding: "JSON"
+    supportedHeaders:
+      - "Sforce-Query-Options"
```
aperez-worklytics, line 137: End-to-end test with https://github.com/Worklytics/evalengin/pull/5999

docs/sources/example-rules/dropbox-business/dropbox-business.yaml
aperez-worklytics: https://app.asana.com/0/1205303754976321/1205322459838014

java/gateway-core/src/main/java/com/avaulta/gateway/rules/Endpoint.java
```diff
@@ -55,6 +55,16 @@ public Optional<Set<String>> getAllowedMethods() {
         return Optional.ofNullable(allowedMethods);
     }
 
+    //if provided, headers provided will be pass-through to the endpoint
+    // NOTE: Using List, as Set is not being serializable in YAML
```
aperez-worklytics, line 59: This is interesting; I have to put `List` instead of `Set` -which originally I've used- because I've discovered that is not being serialized in YAML. Then, I'll adapt https://app.asana.com/0/1205303754976321/1205322459838014 in a different PR, because even the change is simple, the changes are going to be huge (all YAMLs + rules structure files)

aperez-worklytics (commented):
Thanks @eschultink  - moved to rules

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -411,7 +411,9 @@ private com.google.api.client.http.HttpHeaders getHeadersFromSource(HttpEventReq
 
         sanitizer.getSupportedHeaders(request.getHttpMethod(), targetUrl)
                 .ifPresent(i -> i.forEach(h -> {
-                    request.getHeader(h).ifPresent(headerValue -> headers.set(h, headerValue));
+                    request.getHeader(h).ifPresent(headerValue -> {
+                        log.info(String.format("Header %s included", h));
```
eschultink, line 415: may be too verbose for prod use, although i understand that in practice we don't have very many cases where we're forwarding headers to source

java/gateway-core/src/main/java/com/avaulta/gateway/rules/Endpoint.java
```diff
@@ -57,10 +57,10 @@ public Optional<Set<String>> getAllowedMethods() {
 
     //if provided, headers provided will be pass-through to the endpoint
     @JsonInclude(JsonInclude.Include.NON_NULL)
-    Set<String> supportedHeaders;
+    List<String> supportedHeaders;
```
eschultink, line 61: comment for necessity of this?

java/gateway-core/src/main/java/com/avaulta/gateway/rules/Endpoint.java
```diff
@@ -57,10 +57,10 @@ public Optional<Set<String>> getAllowedMethods() {
...
 
     @JsonIgnore
-    public Optional<Set<String>> getSupportedHeaders() {
+    public Optional<List<String>> getSupportedHeadersAsOptional() {
```
eschultink, line 64: could be `Collection`

docs/sources/example-rules/dropbox-business/dropbox-business.yaml
eschultink: i don't really care; mark it `alpha` everywhere and we can return to it later

docs/sources/example-rules/dropbox-business/dropbox-business.yaml
```diff
@@ -47,6 +49,8 @@ endpoints:
           - "$.entries[*]['name', 'path_lower', 'path_display', 'preview_url', 'content_hash']"
           - "$.entries[*].export_info['export_as']"
           - "$.entries[*].file_lock_info['lockholder_name']"
+    supportedHeaders:
+      - "Dropbox-API-Select-User"
```
eschultink, line 53: if this is common case, I'm OK to have this as a top-level property as alternative? (eg, on `Rules`, rather than `Endpoint`; and for each endpoint we allow super-set of headers specified on `Rules` + `Endpoint`?)

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -85,11 +85,7 @@ public RESTApiSanitizerImpl(@Assisted RESTRules rules,
 
     @Override
     public boolean isAllowed(@NonNull String httpMethod, @NonNull URL url) {
-        if (rules.getAllowAllEndpoints()) {
-            return true;
-        } else {
-            return getEndpoint(httpMethod, url).isPresent();
-        }
+        return rules.getAllowAllEndpoints() || getEndpoint(httpMethod, url).isPresent();
```
eschultink, line 88: 👍

docs/sources/example-rules/dropbox-business/dropbox-business.yaml
aperez-worklytics: Done in https://github.com/Worklytics/evalengin/pull/6000

docs/sources/example-rules/dropbox-business/dropbox-business.yaml
aperez-worklytics: Good point, let's see when I revisit dropbox; ATM for Salesforce is for endpoints related with SOQL.

java/core/core.iml
aperez-worklytics, line 1: I have some issues with intelliJ (packages not found, compilation errors, etc). But from terminal or even maven from IDE work fine.
I think this file was out-of-date; valid file is `psoxy-core.iml` because is how the pom refers to the project, but time ago the project was `core` and the iml file was `core.iml`. That was causing some inconsistencies, at least in my side. 
After dropping that file, issue of "package not found" has been solved, but time to time I have issues with compilation resources which I expect to find the root and fix them at some point.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/Endpoint.java
```diff
@@ -55,6 +55,15 @@ public Optional<Set<String>> getAllowedMethods() {
         return Optional.ofNullable(allowedMethods);
     }
 
+    //if provided, headers provided will be pass-through to the endpoint
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    Set<String> supportedHeaders;
```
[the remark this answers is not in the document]
aperez-worklytics: > `allowed` may not be the best term, as unlike for `allowedQueryParams`, proxy will not return a 403 if request includes headers not in this list; it will just not forward them. But I think it's OK for consistency with `allowedQueryParams`, even if the logic isn't perfectly analogous.

Finally I've put `allowed` as part of the name, but I was between that and `supported`. Probably I'd have a better name in the future.

eschultink (approved):

PR #495, opened 2023-08-15
avoid json schema filter npe

eschultink:
### Fixes
  - NPE seen in some json schema filter case

### Features
   - less verbose logging of redacted fields from json schema filters

### Change implications

### DIFF
--- a/java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilterUtils.java
+++ b/java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilterUtils.java
@@ -8,7 +8,9 @@ import lombok.experimental.SuperBuilder;
 import lombok.extern.java.Log;
 import org.apache.commons.lang3.tuple.Pair;
 
+import java.io.Serializable;
 import java.util.*;
+import java.util.stream.Collectors;
 
 @Log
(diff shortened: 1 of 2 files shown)


java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilterUtils.java
```diff
@@ -185,10 +227,13 @@ private Object filterBySchema(String path, JsonNode provisionalOutput, JsonSchem
 
                         String key = entry.getKey();
                         JsonNode value = entry.getValue();
-                        JsonSchemaFilter propertySchema = schema.getProperties().get(key);
+                        JsonSchemaFilter propertySchema =
+                            schema.getProperties() == null ? null : schema.getProperties().get(key);
```
aperez-worklytics, line 231: that was the NPE, right?

aperez-worklytics (approved):

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilterUtils.java
eschultink: yeah, at least in the trace from yesterday

PR #500, opened 2023-08-18
S155 delete sanitized file on test completion

davidfq:
### Fixes
.

### Features
 - [add option to DELETE bulk file from bulk -output bucket after test](https://app.asana.com/0/1201039336231823/1205043877019512/f): it's `default = true` now, so no need to update `.sh` scripts

### Change implications

### DIFF
--- a/tools/psoxy-test/cli-file-upload.js
+++ b/tools/psoxy-test/cli-file-upload.js
@@ -34,6 +34,7 @@ const { version } = require('./package.json');
       'us-east-1')
     .option('-v, --verbose', 'Verbose output', false)
     .option('-s, --save-sanitized-file', 'Save sanitized file to disk', false)
+    .option('--delete-sanitized-file', 'Delete sanitized file from output bucket', false)
     .configureOutput({
       outputError: (str, write) => write(chalk.bold.red(str)),
     });
--- a/tools/psoxy-test/cli-file-upload.js
+++ b/tools/psoxy-test/cli-file-upload.js
@@ -34,7 +34,7 @@ const { version } = require('./package.json');
       'us-east-1')
     .option('-v, --verbose', 'Verbose output', false)
     .option('-s, --save-sanitized-file', 'Save sanitized file to disk', false)
-    .option('--delete-sanitized-file', 'Delete sanitized file from output bucket', false)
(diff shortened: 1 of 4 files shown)


tools/psoxy-test/cli-file-upload.js
```diff
@@ -48,6 +49,10 @@ const { version } = require('./package.json');
 
   program.parse(process.argv);
   const options = program.opts();
+  if (_.isString(options.deleteSanitizedFile)) {
+    options.deleteSanitizedFile = !(options.deleteSanitizedFile === 'false');
```
aperez-worklytics, line 53: I don't see this comparison, it will be always return same value as defined in `options.deleteSanitizedFile`

aperez-worklytics (approved):

tools/psoxy-test/cli-file-upload.js
davidfq: Yes; hard to see, a mix of JS and `commander.js` weirdness.

`--delete-sanitized-file` is true by default even if absent as CLI option. So, if user wants to override, the option has to be `--delete-sanitized-file false`, and commander will parse the `false` as string. Then, `false` strict equality acts as type casting, and will get us back to the default `true`, if the option is any other value than `"false"`.

`--delete-sanitized-file false` -> `options.deleteSanitizedFile === false`
`--delete-sanitized-file foo` -> `options.deleteSanitizedFile === true`

tools/psoxy-test/lib/gcp.js
```diff
@@ -178,14 +178,29 @@ async function upload(bucketName, filePath, client, filename) {
...
+ * @param {string} bucketName
+ * @param {string} filename - name of the file to delete (includes "path")
+ * @param {Storage} client
+ * @returns {Promise} - https://googleapis.dev/nodejs/storage/latest/global.html#DeleteFileResponse
+ */
+async function deleteFile(bucketName, filename, client) {
+  if (!client) {
+    client = createStorageClient();
+  }
+  return client.bucket(bucketName).file(filename).delete();
```
jlorper, line 193: will this fail if path doesn't exist? maybe check first? same for aws

jlorper (approved):

PR #505, opened 2023-08-24
v0.4.36

eschultink:
v0.4.36 back to main


aperez-worklytics (approved):
I'd integrate https://github.com/Worklytics/psoxy/pull/524 and I think release will be ready from my side

PR #506, opened 2023-08-25
support microsoft 365 connectors in GCP-hosted case

aperez-worklytics:
- Adding support for federation between MSFT and GCP for MSFT AAD Apps
- Fixed an issue with header parsing from https://github.com/Worklytics/psoxy/pull/491 where the auth header were remove from target request

### Features
- [support federated identity to MSFT 365 sources from GCP](https://app.asana.com/0/1201039336231823/1204771340174900)

### Change implications


infra/examples-dev/gcp/msft-365.tf
```diff
@@ -0,0 +1,59 @@
...
+locals {
+  msft_api_connectors_with_auth = {
+    for k, msft_connector in module.worklytics_connectors_msft_365.enabled_api_connectors :
+    k => merge(msft_connector, {
+      environment_variables = merge(try(msft_connector.environment_variables, {}),
+        {
+          # NOTE: hardcoded due a cycle (through api_connectors -> module.psoxy), ideally better if coming from
+          # module.msft-connection-auth-federation[k].audience output variable
+          # but for GCP is always "api://AzureADTokenExchange".
+          AUDIENCE          = "api://AzureADTokenExchange"
```
aperez-worklytics, line 48: Cleaner to get it from the output, but a big change needs to be done or the approach should be with a different structure than in AWS case. I've done this to keep same approach but having this hardcoded; in theory for GCP it should not change

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -402,8 +402,8 @@ private void logIfDevelopmentMode(Supplier<String> messageSupplier) {
         }
     }
 
-    private com.google.api.client.http.HttpHeaders getHeadersFromSource(HttpEventRequest request, URL targetUrl) {
-        com.google.api.client.http.HttpHeaders headers = new com.google.api.client.http.HttpHeaders();
+    private void populateHeadersFromSource(HttpRequest sourceApiRequest, HttpEventRequest request, URL targetUrl) {
+        com.google.api.client.http.HttpHeaders headers = sourceApiRequest.getHeaders();
```
aperez-worklytics, line 406: That was the issue with authentication. Some endpoints returns 401, other errors, other just empty results... hard to see. 
Any header (only authentication) added when the request was created on https://github.com/Worklytics/psoxy/pull/506/files#diff-13cad7c0c6986bc926b158c0c4f6eccfd88cfc3623a4f76b37b2e94d2099c6c0L178 was removed here as a new entity was created instead of reusing them.

aperez-worklytics: BTW I can create a PR only with this if you prefer not merging the whole PR for the release @eschultink

infra/modules/aws-psoxy-lambda/main.tf
```diff
@@ -40,9 +40,9 @@ locals {
   salt_arn              = [for l in var.global_parameter_arns : l if endswith(l, local.salt_parameter_name_suffix)][0]
   path_to_shared_config = regex("arn.+parameter(/.*)${local.salt_parameter_name_suffix}", local.salt_arn)[0]
 
-  bundle_from_s3 = startswith(var.path_to_function_zip, "s3://")
-  s3_bucket      = local.bundle_from_s3 ? regex("s3://([^/]+)/.*", var.path_to_function_zip)[0] : null
-  s3_key         = local.bundle_from_s3 ? regex("s3://[^/]+/(.*)", var.path_to_function_zip)[0] : null
+  bundle_from_s3  = startswith(var.path_to_function_zip, "s3://")
```
aperez-worklytics, line 43: This and other similar changes done as result of running `terraform fmt --recursive`

infra/examples-dev/gcp/msft-365-variables.tf
eschultink, line 1: assuming same as `aws-all` case, should add it to `tools/release/sync-examples.sh`

infra/modules/gcp-host/output.tf
```diff
@@ -17,6 +17,10 @@ output "pseudonym_salt" {
   value       = module.psoxy.pseudonym_salt
 }
 
+output "service_accounts" {
```
eschultink, line 20: this is vague as to what these pertain to.

and especially when looking at it, they're just those for `api_connector`s, not bulk cases.

so I would name more clearly (`api_connector_gcp_execution_service_accounts`) and add a description; OR pull from `api_connector_instances`

infra/modules/aws-psoxy-lambda/main.tf
eschultink: thanks, i try to routinely do that when prep'ing releases, but must have missed for last one

eschultink (changes requested):
great; only objecting to the `output` name

if this works, let's get it into the release

infra/modules/worklytics-connectors-msft-365/main.tf
```diff
@@ -1,7 +1,3 @@
-# NOTE: region used to be passed in as a variable; put it MUST match the region in which the lambda
-# is provisioned, and that's implicit in the provider - so we should just infer from the provider
-data "aws_region" "current" {}
```
aperez-worklytics, line 3: BTW I've forgot to comment this; not sure why it was here but I've removed it. I think is not going to break AWS implementation as this variable is not used here, but just for double checking @eschultink

eschultink: maybe to pull aws account ID or ARN of the user who's running terraform, or something like that. If stuff still works, shouldn't be a problem.

likely a relic of when we did the cognito stuff in there.

eschultink (approved):

eschultink (approved):

PR #507, opened 2023-08-25
GCP / Google auth fixes

eschultink:
### Fixes
  - avoid exception that isn't breaking things, but filling logs
  - ~~copy Credential, in case we're getting the same one repeatedly from the credential cache (which is likely to work 80% of the time, so could be broken in 0.4.35??)~~

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/GoogleCloudPlatformServiceAccountKeyAuthStrategy.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/GoogleCloudPlatformServiceAccountKeyAuthStrategy.java
@@ -124,8 +124,10 @@ public class GoogleCloudPlatformServiceAccountKeyAuthStrategy implements SourceA
         //even though GoogleCredentials implements `createDelegated`, it's a no-op if the
         // credential type doesn't support it.
         // similarly, createScoped() is no-op if not supported by GoogleCredentials implementation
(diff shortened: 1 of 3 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/GoogleCloudPlatformServiceAccountKeyAuthStrategy.java
```diff
@@ -122,7 +123,9 @@ GoogleCredentials buildImpersonatedCredentials(@NonNull String accountToImperson
 
         //even though GoogleCredentials implements `createDelegated`, it's a no-op if the
         // credential type doesn't support it.
+        // similarly, createScoped() is no-op if not supported by GoogleCredentials implementation
         return baseCredentials
+            .toBuilder().build() // force a copy of base, which we'll then mutate to be delegated, scoped
```
eschultink, line 128: so actually I'm wrong, it's OK in the cases that matter: 
psoxy____code_psoxy__–_ServiceAccountCredentials_java__Maven__com_google_auth_google-auth-library-oauth2-http_1_18_0_

eschultink: reverting; i had done this thinking it might be the source of auth problem; but it's OK

aperez-worklytics (approved):

PR #508, opened 2023-08-25
transform to preserve `proxyAddresses` format

eschultink:
### Fixes
 - serializing 'reversible' pseudonyms in URL_SAFE_TOKEN format not compatible with LEGACY pseudonym implementation; so will failover to JSON format in those cases

### Features
  - new transform to enable preserving `proxyAddresses` format

### Change implications

 - something to note in `CHANGELOG.md`? **yes** - some logic changes to be mindful of for users of LEGACY format

---
  - https://app.asana.com/0/0/1205345791601459

### DIFF
--- a/docs/sources/api-response-examples/microsoft-365/directory/sanitized-no-app-ids/user.json
+++ b/docs/sources/api-response-examples/microsoft-365/directory/sanitized-no-app-ids/user.json
@@ -11,7 +11,7 @@
     "{\"scope\":\"email\",\"domain\":\"gmail.com\",\"hash\":\"XRcSIwOXlgrxLZOIHxdtE3GCke5TsftXzOm5zmnm36g\"}"
   ],
   "proxyAddresses":[
-    "{\"scope\":\"email\",\"domain\":\"M365x214355.onmicrosoft.com\",\"hash\":\".tbhph3DloGSd5l9e9TeFqmEboAqU8o6BdigmHZOmxI\"}",
-    "{\"scope\":\"email\",\"domain\":\"M365x214355.onmicrosoft.com\",\"hash\":\"L282PmjOP7xZpTtW7M1sKbWHzVvzLV4Yfp.7x1r3Yk0\"}"
+    "SMTP:t~-tbhph3DloGSd5l9e9TeFqmEboAqU8o6BdigmHZOmxI@M365x214355.onmicrosoft.com",
+    "smtp:t~L282PmjOP7xZpTtW7M1sKbWHzVvzLV4Yfp-7x1r3Yk0@M365x214355.onmicrosoft.com"
   ]
 }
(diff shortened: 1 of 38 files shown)


docs/sources/api-response-examples/microsoft-365/directory/sanitized-no-app-ids/user.json
```diff
@@ -6,12 +6,12 @@
   "officeLocation":"12/1110",
   "preferredLanguage":"en-US",
   "userPrincipalName":"{\"scope\":\"email\",\"domain\":\"M365x214355.onmicrosoft.com\",\"hash\":\".tbhph3DloGSd5l9e9TeFqmEboAqU8o6BdigmHZOmxI\"}",
-  "id":"p~SIoJOpeSgYF7YUPQ28IWZexVuHyN9A80SJXrbfawKpDRcddGnKI4QDKyjQI9KtjJZDb8FZ27UE_toS68FyWz7Y22fnQYLP91SHJGVwQiN3E",
+  "id":"{\"scope\":\"azure-ad\",\"hash\":\"J0_C7jfy9KhwcfmhAjVNQOnzUnYaE3Kovzt2FAkc2W0\",\"reversible\":\"p~SIoJOpeSgYF7YUPQ28IWZexVuHyN9A80SJXrbfawKpDRcddGnKI4QDKyjQI9KtjJZDb8FZ27UE_toS68FyWz7Y22fnQYLP91SHJGVwQiN3E\"}",
```
eschultink, line 9: so if anyone were to use the "no app id" rules and 'LEGACY' format, the `hash` parsed from `reversible` won't match the non-reversible cases.

change in test here reflects that for LEGACY case, we're fail back to JSON. (tests, regretably, mostly run under `LEGACY` implementation)

docs/sources/api-response-examples/microsoft-365/directory/sanitized-no-app-ids/user.json
```diff
@@ -6,12 +6,12 @@
...
   "otherMails":[
     "{\"scope\":\"email\",\"domain\":\"gmail.com\",\"hash\":\"XRcSIwOXlgrxLZOIHxdtE3GCke5TsftXzOm5zmnm36g\"}"
   ],
   "proxyAddresses":[
-    "{\"scope\":\"email\",\"domain\":\"M365x214355.onmicrosoft.com\",\"hash\":\".tbhph3DloGSd5l9e9TeFqmEboAqU8o6BdigmHZOmxI\"}",
-    "{\"scope\":\"email\",\"domain\":\"M365x214355.onmicrosoft.com\",\"hash\":\"L282PmjOP7xZpTtW7M1sKbWHzVvzLV4Yfp.7x1r3Yk0\"}"
+    "SMTP:t~.tbhph3DloGSd5l9e9TeFqmEboAqU8o6BdigmHZOmxI@M365x214355.onmicrosoft.com",
+    "smtp:t~L282PmjOP7xZpTtW7M1sKbWHzVvzLV4Yfp.7x1r3Yk0@M365x214355.onmicrosoft.com"
```
eschultink, line 15: this just looks better; it's correct, bc URL_SAFE_TOKEN encoding of pseudonym without `reversible`; this is compatible with `LEGACY` implementation of pseudonyms

docs/sources/api-response-examples/microsoft-365/directory/sanitized-no-app-ids/user.json
eschultink: (for reference, `LEGACY` implementation uses hash of `(scope + secret + value)`; default is `(secret + value)`); `reversible` is computed at layer of code that is unaware of `scope`, and would be big refactor to put it there to account for very small number of people using `LEGACY` pseudonyms, which was v0.3)

aperez-worklytics (approved):

PR #515, opened 2023-08-30
csv field transforms

eschultink:
### Features
 - support transform of values in bulk data case, pre-pseudonymization

### Change implications

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
@@ -27,6 +27,7 @@ import org.apache.commons.csv.CSVPrinter;
 import org.apache.commons.csv.CSVRecord;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.function.TriFunction;
+import org.apache.commons.lang3.tuple.Pair;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
@@ -90,6 +91,14 @@ public class ColumnarBulkDataSanitizerImpl implements BulkDataSanitizer {
         Set<String> columnsToRedact = asSetWithCaseInsensitiveComparator(rules.getColumnsToRedact());
 
         Set<String> columnsToPseudonymize = asSetWithCaseInsensitiveComparator(rules.getColumnsToPseudonymize());
+
+        Map<String, String> columnsToPseudonymizeWithScope = rules.getColumnsToPseudonymizeWithScope()
+            .entrySet().stream()
+            .collect(Collectors.toMap(
+                entry -> entry.getKey().trim().toLowerCase(),
+                entry -> entry.getValue().trim(),
+                (a, b) -> a,
+                () -> new TreeMap<>(String.CASE_INSENSITIVE_ORDER)));
 
         Map<String, ColumnarRules.FieldValueTransform> columnsToTransform = rules.getColumnsToTransform()
             .entrySet().stream()
@@ -160,14 +169,22 @@ public class ColumnarBulkDataSanitizerImpl implements BulkDataSanitizer {
 
         TriFunction<String, String, Pseudonymizer, String> pseudonymizationFunction = buildPseudonymizationFunction(rules);
 
+        Map<String, Pseudonymizer> pseudonymizers = new HashMap<>();
+
         BiFunction<String, String, String> applyPseudonymizationIfAny = (outputColumnName, value) -> {
-            if (columnsToPseudonymize.contains(outputColumnName)) {
+            if (columnsToPseudonymize.contains(outputColumnName.toLowerCase())) {
                 return pseudonymizationFunction.apply(value, outputColumnName, pseudonymizer);
+            } else if (columnsToPseudonymizeWithScope.containsKey(outputColumnName.toLowerCase())) {
+                Pseudonymizer scopedPseudonymizer = pseudonymizer;
+                if (pseudonymizer.getOptions().getPseudonymImplementation() == PseudonymImplementation.LEGACY) {
+                    scopedPseudonymizer = pseudonymizers.computeIfAbsent(columnsToPseudonymizeWithScope.get(outputColumnName.toLowerCase()),
+                        scope -> pseudonymizerImplFactory.create(pseudonymizer.getOptions().withDefaultScopeId(scope)));
+                }
+
+                return pseudonymizationFunction.apply(value, outputColumnName, scopedPseudonymizer);
             }
             return value;
         };
-
-        Map<String, Pseudonymizer> pseudonymizers = new HashMap<>();
 
         BiFunction<String, String, String> applyTransformIfAny = (outputColumnName, value) -> {
           if (columnsToTransform.containsKey(outputColumnName.toLowerCase())) {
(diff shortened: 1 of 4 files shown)


java/core/src/test/java/co/worklytics/psoxy/storage/impl/BulkDataSanitizerImplTest.java
```diff
@@ -417,4 +417,54 @@ void pre_v0_4_30_bulk_pseudonym_URL_SAFE_TOKEN_ENCODING(String identifier) {
...
+            "4,,Engineering,2023-01-06,1,2018-06-03,,\r\n" +
+            "3,charles@workltycis.co,Engineering,2023-01-06,1,2019-10-06,2022-12-08,\r\n";
+
+        CsvRules rules = CsvRules.builder()
+            .columnsToDuplicate(Map.of("EMPLOYEE_EMAIL", "GITHUB_USERNAME"))
+            .columnsToTransform(Map.of("GITHUB_USERNAME", ColumnarRules.FieldValueTransform.builder()
+                                                                        .filterRegex("(.*)@worklytics.co")
+                                                                        .outputTemplate("%s_enterprise")
+                                                                        .build()))
+            .columnsToPseudonymizeWithScope(Map.of("GITHUB_USERNAME", "github"))
```
eschultink, line 438: this is a point for discussion; I've split the definition of the transform we want into essentially 3 steps that must be configured independently, instead of one 'transform' structure encoding those steps. 

advantage of this approach is that it keeps the usual, simple case simple: just put a list of `columnsToPseudonymize` and you're done; and there's no *other* way to pseudonymize that having the column in that list.

biggest con of this approach is that the order is not obvious; you have to know how it's implementation.

alternatives:

```yaml
- columnsToDuplicateAndTransform:
    EMPLOYEE_EMAIL:
         newColumn: GITHUB_USERNAME
         transforms: 
               - filter: "(.*)@worklytics.co"
               - template: "%s_enterprise"
               - pseudonymizeWithScope: 'github'
```
(Actual yaml would be more verbose than that, unless FieldValueTransform is java object with all the properties, with expectation that in practice only use one)

java/core/src/test/java/co/worklytics/psoxy/storage/impl/BulkDataSanitizerImplTest.java
```diff
@@ -417,4 +417,54 @@ void pre_v0_4_30_bulk_pseudonym_URL_SAFE_TOKEN_ENCODING(String identifier) {
...
+            "2,bob@workltyics.co,Sales,2023-01-06,1,2020-01-01,,\r\n" +
+            "1,alice@worklytics.co,Engineering,2023-01-06,,2019-11-11,,\"{\"\"scope\"\":\"\"github\"\",\"\"hash\"\":\"\"kwv9cWxo7TDgrt1qCegIJv7rA84s_895L_wG_y8hYjA\"\"}\"\r\n" +
+            "4,,Engineering,2023-01-06,1,2018-06-03,,\r\n" +
+            "3,charles@workltycis.co,Engineering,2023-01-06,1,2019-10-06,2022-12-08,\r\n";
+
+        CsvRules rules = CsvRules.builder()
+            .columnsToDuplicate(Map.of("EMPLOYEE_EMAIL", "GITHUB_USERNAME"))
+            .columnsToTransform(Map.of("GITHUB_USERNAME", ColumnarRules.FieldValueTransform.builder()
+                                                                        .filterRegex("(.*)@worklytics.co")
+                                                                        .outputTemplate("%s_enterprise")
```
jlorper, line 436: fine for a first approach, but could use named groups and apply replace directly
`(?<username>.*)@worklytics.co`
template `"${username}_enterprise"`, and that will apply directly `replaceAll(template)`
or using numerical capture groups rather than finding "%s" in the code

java/core/src/test/java/co/worklytics/psoxy/storage/impl/BulkDataSanitizerImplTest.java
jlorper: I was thinking in order too, 2nd approach is more obvious in that regard, easier to follow.

java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -82,6 +92,22 @@ public byte[] sanitize(@NonNull InputStreamReader reader,
 
         Set<String> columnsToPseudonymize = asSetWithCaseInsensitiveComparator(rules.getColumnsToPseudonymize());
 
+        Map<String, String> columnsToPseudonymizeWithScope = rules.getColumnsToPseudonymizeWithScope()
```
jlorper, line 95: I'd put some comments about the types
// Map<COLUMN_NAME, scope>

java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -141,39 +167,53 @@ public byte[] sanitize(@NonNull InputStreamReader reader,
...
+                      if (transform.getOutputTemplate() != null) {
+                          if (transform.getOutputTemplate().contains("%s")) {
+                              value = String.format(transform.getOutputTemplate(), value);
+                          } else {
+                              throw new IllegalArgumentException("Invalid outputTemplate:" + transform.getOutputTemplate() + " for column " + outputColumnName + ". Must contain %s");
+                          }
+                      }
+                  } else {
+                      value = null;
+                  }
```
jlorper, line 209: Quick sample of what I mentioned, works and more versatile as can leverage full power of regexps, not just one single replacement
```
Matcher matcher = Pattern.compile("(?<username>.*)@worklytics.co").matcher("jose@worklytics.co");

if (matcher.matches() && transform.getOutputTemplate() != null) {
			value = matcher.replaceAll("${username}_github");
} else {
value = null;
}
```

java/core/src/test/java/co/worklytics/psoxy/storage/impl/BulkDataSanitizerImplTest.java
```diff
@@ -417,4 +417,54 @@ void pre_v0_4_30_bulk_pseudonym_URL_SAFE_TOKEN_ENCODING(String identifier) {
...
             urlSafeTokenPseudonymEncoder.encode(pseudonymizedIdentity.asPseudonym()));
     }
+
+
+    @SneakyThrows
+    @Test
+    void transform_ghusername() {
+
+        final String EXPECTED = "EMPLOYEE_ID,EMPLOYEE_EMAIL,DEPARTMENT,SNAPSHOT,MANAGER_ID,JOIN_DATE,LEAVE_DATE,GITHUB_USERNAME\r\n" +
+            "2,bob@workltyics.co,Sales,2023-01-06,1,2020-01-01,,\r\n" +
```
jlorper, line 427: I was trying to understand why bob and charles didn't get a GITHUB_USERNAME... subtle domain change `workltyics.co`...
and `workltycis.co`...
a comment would be welcome here.
Why are the domains messed up in the first place? wouldn't it be easier to just do `@subdomain.worklytics.co` or `@example.com` ??

jlorper (approved):
Some suggestions, but no blockers given the tight deadline

java/core/src/test/java/co/worklytics/psoxy/storage/impl/BulkDataSanitizerImplTest.java
aperez-worklytics: Yeah I think the second one fits better; sometimes order is hard to follow when creating rules.

aperez-worklytics (approved):

jlorper (approved):
OK as is, I still think that forcing to just one replacement %s is constraining ourselves unnecessarily, but we don't expect this to be widely used, so it is fine.

PR #516, opened 2023-08-30
health check error feedback

eschultink:
### Fixes
 - 412 rather than 512 response code
 - log exceptions explicitly, although may be verbose/redundant

### Change implications

 - something to note in `CHANGELOG.md`? **arguably no; yes, it changes health check behavior a little, but would argue that relevant bit is non-200 - not specific error code**

---
  - https://app.asana.com/0/0/1205390449392905

### DIFF
--- a/java/core/src/test/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandlerTest.java
+++ b/java/core/src/test/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandlerTest.java
@@ -7,6 +7,7 @@ import co.worklytics.psoxy.gateway.HttpEventResponse;
 import co.worklytics.psoxy.gateway.ProxyConfigProperty;
 import co.worklytics.test.MockModules;
 import dagger.Component;
+import org.apache.hc.core5.http.HttpStatus;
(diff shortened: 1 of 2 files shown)


java/core/src/test/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandlerTest.java
```diff
@@ -7,6 +7,7 @@
 import co.worklytics.psoxy.gateway.ProxyConfigProperty;
 import co.worklytics.test.MockModules;
 import dagger.Component;
+import org.apache.hc.core5.http.HttpStatus;
```
jlorper, line 10: you meant `org.apache.http.HttpStatus;` ??

jlorper (approved):

java/core/src/test/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandlerTest.java
eschultink: i think the other is newer

PR #522, opened 2023-08-31
support path/query parameter validation

eschultink:
### Features
 - path/query parameter validation

### Change implications

 - something to note in `CHANGELOG.md`? **no, although perhaps should note that some rule sets will become stricter if this is actually used**

---
  - https://app.asana.com/0/0/1204472640908679


java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -474,17 +495,24 @@ Map<Endpoint, Pattern> getCompiledAllowedEndpoints() {
 
     @VisibleForTesting
     String effectiveRegex(Endpoint endpoint) {
+        //NOTE: java capturing groups names limited to A-Z, a-z and 0-9, and must start with a letter
```
eschultink, line 498: previously, we had no restriction on path parameter names, other than no `}`; so actually our github rules use snake_case, which is invalid

Openapi spec examples use camelCase
see : https://swagger.io/docs/specification/describing-parameters/#path-parameters

eschultink: i think restricting is better than trying to magically re-map `foo_bar` to `fooBar` or something, and hoping things don't collide.

while technically this is a breaking change, shouldn't affect many rule sets in the wild

aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -117,12 +121,28 @@ Predicate<Map.Entry<Endpoint, Pattern>> getHasPathRegexMatchingUrl(String relati
...
-                        && allowedQueryParams(entry.getKey(), URLUtils.queryParamNames(url));
+        return (entry) -> {
+            if (entry.getKey().getPathTemplate() != null) {
+                Matcher matcher = entry.getValue().matcher(url.getPath());
+                if (matcher.matches()) {
+                    boolean allParamsValid =
+                            entry.getKey().getPathParameterSchemasOptional()
+                                    .map(schemas -> schemas.entrySet().stream()
+                                        .map(paramSchema -> parameterSchemaUtils.validate(paramSchema.getValue(), matcher.group(paramSchema.getKey())))
+                                         .allMatch(b -> b))
```
jlorper, line 132:
```suggestion
                                        .allMatch(paramSchema -> parameterSchemaUtils.validate(paramSchema.getValue(), matcher.group(paramSchema.getKey()))))
```
don't need the map, right?

java/gateway-core/src/main/java/com/avaulta/gateway/rules/ParameterSchemaUtils.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param schema to validate against
+     * @param value to validate
+     * @return whether value conforms to schema
+     */
+    public boolean validate(ParameterSchema schema, String value) {
+        if (value != null) {
+            if (schema.getType() != null) {
+                if (schema.getType().equals(ParameterSchema.ValueType.INTEGER.getEncoding())) {
+                    try {
+                        Integer.parseInt(value);
```
jlorper, line 32: I'd be safe and do `Long.parseLong`, things can easily go beyond 2³² in autogenerated ids.

java/gateway-core/src/test/java/com/avaulta/gateway/rules/ParameterSchemaUtilsTest.java
```diff
@@ -0,0 +1,86 @@
...
+        ParameterSchema parameterSchema = ParameterSchema.builder()
+                .type(ParameterSchema.ValueType.STRING.getEncoding())
+                .enumValues(Arrays.asList("v0.1", "beta"))
+                .build();
+
+        assertTrue(parameterSchemaUtils.validate(parameterSchema, "beta"));
+        assertTrue(parameterSchemaUtils.validate(parameterSchema, "v0.1"));
+        assertFalse(parameterSchemaUtils.validate(parameterSchema, "v0.2"));
+
+        assertTrue(parameterSchemaUtils.validate(parameterSchema, null));
```
jlorper, line 42: this could also be an option, absence allowed or not. If we are expecting an enum, I don't see why null should be accepted unless it is not mandatory.
Same above for the reversible.

jlorper (approved):
No blockers, but a couple suggestions

PR #526, opened 2023-09-04
Missing rule for Salesforce email

aperez-worklytics:
Can be deployed as a hotfix of v4.0.37

### Fixes
- [Email not matching](https://app.asana.com/0/1205088991476808/1205397231714140/f)

### Change implications


docs/sources/api-response-examples/asana/sanitized/workspaces.json
```diff
@@ -0,0 +1,13 @@
+{
```
aperez-worklytics, line 1: As a result of couple of more tests added in Asana

eschultink (approved):

PR #532, opened 2023-09-08
document response schema

eschultink:
### Features
 - documentation on response schemas, default denyu

### Change implications

### DIFF
--- a/docs/api-data-sanitization.md
+++ b/docs/api-data-sanitization.md
@@ -147,7 +147,8 @@ not of expected type, will be removed from the response.
       - `object` a JSON object
       - `array` a JSON array
       - `string` a JSON string
-      - `number` a JSON number, including integers or decimal.
+      - `number` a JSON number, either integer or decimal.
+      - `integer` a JSON integer (not a decimal)
       - `boolean` a JSON boolean
   - `properties` - for `type == object`, a map of field names to schema to filter field's value
      against (eg, another `JsonSchemaFilter` for the field itself)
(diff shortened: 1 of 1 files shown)


docs/api-data-sanitization.md
```diff
@@ -122,5 +141,141 @@ Options:
...
+  - `items` - for `type == array`, a schema to filter each item in the array against (again, a `JsonSchemaFilter`)
+  - `format` - for `type == string`, a format to expect for the string value. As of v0.4.38, this is
+     not enforced by the proxy.
+  - `$ref` - a reference to a schema specified in the `definitions` property of the root schema.
+  - `definitions` - a map of schema names to schemas of type `JsonSchemaFilter`; only supported at
+     root schema of endpoint.
+
+Example:
+
+The following is for a User from the GitHub API. See: https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-a-user
```
aperez-worklytics, line 183: The example is actually from getting users from GraphAPI: https://docs.github.com/en/graphql/guides/forming-calls-with-graphql#the-graphql-endpoint

And I'd change the phrase to emphasize that with this example we are enforcing schema definition to match what we expect, avoiding to have any other field you could do over GraphQL

aperez-worklytics (approved):

jlorper (approved):

docs/api-data-sanitization.md
eschultink: 👍 thanks - yeah, I see that I skimmed past as `graphql` path template when I was looking at example.

we should think of how to deal with graphql endpoints more generally; I guess only solution is an `or` at the root of the schema? (Eg, from graphql we expect `!user` or `!commit` etc?) Workable, but perhaps there's a better way.

PR #533, opened 2023-09-08
Fix invalid JSON alias for schema

aperez-worklytics:
Fixed invalid JSON alias for schema

### Fixes
- [Fix version alias in Proxy](https://app.asana.com/0/1205397231714086/1205449579398900)

### Change implications


docs/sources/api-response-examples/github/issue_timeline.json
aperez-worklytics, line 1: Updated examples from GH; not related with this but not sure what other stuff I could have for including it, so as it is very very minor, doing it here.

eschultink (approved):

PR #542, opened 2023-09-21
detect gcp identity

eschultink:
### Fixes
 - google_client_openid_userinfo empty in Google Cloud shell

### Change implications


eschultink:
use of external should be avoided, and isn't gaining us anything vs #545

PR #543, opened 2023-09-21
 pass gcp identity in

eschultink:
given limitations of #545, this is the only viable solution to support service account impersonation for `google` provider case; given that impersonating service accounts is best practice, we really should get this working

### Fixes
 - data resources used by #545 don't expose email address of service account, in case that you're impersonating the service account via terraform (if application default credential is a service account, I think it will work)

### Change implications


jlorper (commented):
OK, this is in draft,but looks good to me

eschultink:
solutions on #549 should work for all scenarios EXCEPT possibly TF cloud using impersonation; if TF cloud is directly auth'd, I think it should work.

PR #545, opened 2023-09-22
detection of gcp identity in cloud shell

eschultink:
### Fixes
  - improve detection of GCP identity in Google Cloud Shell

### Change implications

---
  - https://app.asana.com/0/0/1205549827730455

### DIFF
--- a/infra/modules/gcp-tf-runner/main.tf
+++ b/infra/modules/gcp-tf-runner/main.tf
@@ -11,27 +11,24 @@ data "google_client_openid_userinfo" "me" {
 
 }
 
-data "external" "gcloud_active_account" {
+
+# if no 'email' field from 'google_client_openid_userinfo', attempt to generate id token for the
+# current user, which *should* be the principal that Terraform is running - hopefully will correctly
+# be a user if directly authenticated; and the service account if using impersonation.
+data "google_service_account_id_token" "identity" {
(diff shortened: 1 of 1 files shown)


infra/modules/gcp-tf-runner/main.tf
```diff
@@ -5,29 +5,68 @@
 # times throughout the code base, and includes some hard-coded convention stuff that imho is better
 # to have in one place.
 
+
+# in cloud shell, this seems to return {"email":"", "id":""}
+# in any env, this is NEVER the gcp service account configured via provider block
+# (eg, google.impersonate_service_account = "terraform@...")
```
jlorper, line 11: can't we use env variables and make the user explicitly set it in cloud shell as prereq??

export TF_VAR_email=whatever@iam.gserviceaccount.com

jlorper (approved):

infra/modules/gcp-tf-runner/main.tf
eschultink: does that work below the top-level? I would think it would not.

Our problem is that in individual modules, they need to know the identity TF is running as, to grant `ServiceAccountUser` to TF on the SA that TF itself created.  It's a little strange to require this to be passed into the module, as well as tedious because those modules are actually called 2-3 levels deep.

PR #549, opened 2023-09-26
parse SA for tf runner, in gcp case

eschultink:
### Features
 - parse 'gcp_terraform_sa_account_email' from `terraform.tfvars`, to support GCP terraform applies via SA impersonation

### Change implications

### DIFF
--- a/infra/modules/gcp-tf-runner/main.tf
+++ b/infra/modules/gcp-tf-runner/main.tf
@@ -30,6 +30,11 @@ data "google_service_account_id_token" "identity" {
   target_audience = "worklytics.co/gcp-tf-runner"
 }
 
+# as this runs `terraform console` underneath, it is hacky and SLOW!!!
+data "external" "identity" {
+  program = ["./${path.module}/read-tfvars.sh", "gcp_terraform_sa_account_email"]
+}
+
 # alternative ideas
 #  - data.google_client_config has an 'access_token', but that's opaque so can't parse email from it
 #  - call outs to gcloud CLI via data.external --> still won't know about impersonation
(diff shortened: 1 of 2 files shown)


infra/modules/gcp-tf-runner/main.tf
```diff
@@ -30,6 +30,11 @@ data "google_service_account_id_token" "identity" {
   target_audience = "worklytics.co/gcp-tf-runner"
 }
 
+# as this runs `terraform console` underneath, it is hacky and SLOW!!!
+data "external" "identity" {
```
jlorper, line 34: thinking something like this should exist just found this -> https://registry.terraform.io/providers/adminsentecacommerce/tfvars/latest

Not good examples on how to use though

infra/modules/gcp-tf-runner/main.tf
```diff
@@ -30,6 +30,11 @@ data "google_service_account_id_token" "identity" {
   target_audience = "worklytics.co/gcp-tf-runner"
 }
 
+# as this runs `terraform console` underneath, it is hacky and SLOW!!!
```
jlorper, line 33: seems you don't use `terraform console` anymore, right? remove this line?

jlorper (approved):

infra/modules/gcp-tf-runner/main.tf
jlorper: Also this -> https://terraform-docs.io/reference/tfvars-json/

eschultink: That would be fine for internal-use, but we need to minimize dependencies, esp those published by small 3rd parties.

PR #550, opened 2023-09-27
v0.4.39

eschultink:
v0.4.39 back to main

### DIFF
--- a/infra/modules/gcp-tf-runner/read-tfvars.sh
+++ b/infra/modules/gcp-tf-runner/read-tfvars.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+variable=$1
+filename="${2:-terraform.tfvars}"
+
+# TODO: extend to look at all *.tfvars files in the directory, not just terraform.tfvars
+
+json="{"
+
+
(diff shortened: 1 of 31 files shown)


infra/modules/gcp-tf-runner/read-tfvars.sh
```diff
@@ -0,0 +1,34 @@
...
+
+
+# Parse the file, find variable, convert it to JSON
+# this approach of parsing it from a `terr
+# alternative to parsing it ourselves would be `echo var.$variable | terraform console`, but that is
+# slow and brittle to errors/warnings from terraform console.
+if [ -f $filename ]; then
+  while IFS='=' read -r key value; do
+      # Skip commented lines
+      [[ $key =~ ^#.*$ ||  -z $key ]] && continue
```
jlorper, line 18: Forgot to tell you that this doesn't work for nested objects, which we luckily don't have

jlorper (approved):

infra/modules/gcp-tf-runner/read-tfvars.sh
eschultink: yeah; also lists won't work. can add a comment

PR #552, opened 2023-09-28
ip transforms

eschultink:
### Features
 - `hashIp` transform
 - `encryptIp` transform

NOTE: abandoned idea to make `hashIp` format-preserving, because I think that its preferable for it NOT to be format-preserving as long as clients can tolerate.  That clients (And humans) can tell based on the format that the data is encrypted/hashed is helpful.  It makes it evident that proxy is working, which if you preserved format you don't know. 

### Change implications

---
  - https://app.asana.com/0/0/1205533487577598


java/core/src/test/java/co/worklytics/psoxy/impl/RESTApiSanitizerImplTest.java
```diff
@@ -590,4 +594,69 @@ public void pseudonymizeWithRegexMatches_nonMatchingRedacted(String input, Strin
...
+
+            assertEquals(
+                base64encoder.encodeToString(hash),
+                base64encoder.encodeToString(Arrays.copyOfRange(decoded, 0, Md5DeterministicTokenizationStrategy.HASH_LENGTH_BYTES))
+            );
+
+            // TODO: fix the following; UrlSafeTokenPseudonymEncoder can't decode the encrypted value
+            // because it produces a 32-byte hash, but the EncryptIP uses 16-byte hash
+            // options:
+            //  - a different prefix (m~) that indicates 16-byte hash?
```
eschultink, line 650: or maybe `e16~` for encrypted w 16-byte IV? 
do we need to encode more info about encryption algorithm here??

or for `encryptIP` we relax the property that the IV == the result of the `hashIP` transform?

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -259,12 +271,72 @@ public MapFunction getTransformImpl(Transform transform) {
...
+                if (s != null) {
+                    log.warning("value matched by " + transform + " not of type String");
+                }
+                return null;
+            } else if (StringUtils.isBlank((String) s)) {
+                return s;
+            } else {
+                try {
+                    // Parse the IP address string and convert back to string to get the canonical form.
+                    InetAddress address = InetAddress.getByName((String) s);
```
jlorper, line 297: this is going to do an extra call, but OK as it has a cache built-in.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -259,12 +271,72 @@ public MapFunction getTransformImpl(Transform transform) {
...
+                    // already canonicalized, so can pass identity
+                    return urlSafePseudonymEncoder.encode(Pseudonym.builder()
+                            .hash(ipHashStrategy.getToken(address.getHostAddress()))
+                            .reversible(ipEncryptStrategy.getReversibleToken(address.getHostAddress()))
```
jlorper, line 301: why do we want to reverse ips on our side??

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -259,12 +271,72 @@ public MapFunction getTransformImpl(Transform transform) {
...
+                if (s != null) {
+                    log.warning("value matched by " + transform + " not of type String");
+                }
+                return null;
+            } else if (StringUtils.isBlank((String) s)) {
+                return s;
+            } else {
+                try {
+                    // Parse the IP address string and convert back to string to get the canonical form.
+                    InetAddress address = InetAddress.getByName((String) s);
```
jlorper, line 325: all this code can be reused for both hash and encrypt cases

java/core/src/test/java/co/worklytics/psoxy/impl/RESTApiSanitizerImplTest.java
```diff
@@ -590,4 +594,62 @@ public void pseudonymizeWithRegexMatches_nonMatchingRedacted(String input, Strin
...
+
+
+            assertEquals(
+                base64encoder.encodeToString(hash),
+                base64encoder.encodeToString(Arrays.copyOfRange(decoded, 0, Sha256DeterministicTokenizationStrategy.HASH_SIZE_BYTES))
+            );
+
+            UrlSafeTokenPseudonymEncoder encoder = new UrlSafeTokenPseudonymEncoder();
+            assertTrue(encoder.canBeDecoded(encrypted));
+            Pseudonym decodedPseudonym = encoder.decode(encrypted);
```
jlorper, line 649: this probably used in more tests, we should just have a test util that given a reversible token returns hash, reversed, decoded

java/gateway-core/src/main/java/com/avaulta/gateway/pseudonyms/impl/UrlSafeTokenPseudonymEncoder.java
```diff
@@ -43,9 +44,13 @@ public class UrlSafeTokenPseudonymEncoder implements PseudonymEncoder {
     public String encode(Pseudonym pseudonym) {
         String encoded;
         if (pseudonym.getReversible() == null) {
-            encoded = TOKEN_PREFIX + encoder.encodeToString(pseudonym.getHash());
+            encoded = TOKEN_PREFIX + encoder.encodeToString(pseudonym.getHash());;
         } else {
-            encoded = REVERSIBLE_PREFIX + encoder.encodeToString(pseudonym.getReversible());
+            if (!Arrays.equals(pseudonym.getHash(), 0, pseudonym.getHash().length, pseudonym.getReversible(), 0, pseudonym.getHash().length)) {
```
jlorper, line 49: :red_circle: is this correct?? last byte of the hash can be padding, so it may not match exactly.
from your tests
`        "123.234.252.12,t~7USliSM4GiS0Xfk1DXIAH-4nK-UkLJlSAA_5ZqQh_CI", `
`        "123.234.252.12,p~7USliSM4GiS0Xfk1DXIAH-4nK-UkLJlSAA_5ZqQh_CJStJlrfaIPHLZUvx-dHHQp",`

7USliSM4GiS0Xfk1DXIAH-4nK-UkLJlSAA_5ZqQh_**CI** != 7USliSM4GiS0Xfk1DXIAH-4nK-UkLJlSAA_5ZqQh_**CJ**

jlorper (changes requested):
Some questions, but flagged one check in the url safe token encoder

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
aperez-worklytics: But the cache is local, as part of the static instance of `InetAddress`, of course this is better than nothing, but if worth to use `ConfigService` to store the value and use it as a shared cache?

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -87,6 +94,11 @@ public RESTApiSanitizerImpl(@Assisted RESTRules rules,
     @Inject
     ParameterSchemaUtils parameterSchemaUtils;
 
+    @Inject @Named("ipEncryptionStrategy")
```
aperez-worklytics, line 97: I'd move `ipEncryptionStrategy` into a public property of `ReversibleTokenizationStrategy` (same for `ipHashStrategy` and `DeterministicTokenizationStrategy`), so we don't need to use them as hardcode values in modules and here.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
eschultink: I think I removed this, as I decided preserving format not worthwhile.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
eschultink: we don't reverse IPs on our side. 

We support `encryptIP` and `hashIP` transforms; customers can choose. I think use-case for the encryption would be that any IP that customer sees associated with above some number of people can be safely reversed/decrypted, as it's not identifying the home of a person. 

Customers may have use-case for having plain IPs to be able to localize the physical office / building / access point in such situations. With encryption, they could build a list of IPs by number of associated people, and decrypt only those that they're OK with from privacy perspective.

java/gateway-core/src/main/java/com/avaulta/gateway/pseudonyms/impl/UrlSafeTokenPseudonymEncoder.java
eschultink: that's just an artifact of the encoding. Each char of base64 encodes 6 bits of the binary sequence; 32 byte hash --> 43 bytes of base64

in 2nd case, you take 32 bytes + X bytes of encrypted content, and THEN base64 whole thing. so some bits from encrypted content alter the 43rd byte of the base64 of that.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
eschultink: It's a instance of `ReversibleTokenizationStrategy` with different config than the regular one ... I think exposing all through `ReversibleTokenizationsStrategy` is too much. If anything I'd inject a factory-type thing with methods for default, ip, etc ... but the whole point of using a DI FW is to avoid factorie

jlorper (approved):

PR #556, opened 2023-10-17
multi-file schema bulk rules

eschultink:
### Features
 - support case where multiple file types need to be processed via single Bulk connector
 - rules that support schemas beyond columnar; eg record-based, expecting NDJSON

### Change implications

---
  - https://app.asana.com/0/0/1205552701863602


java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -65,9 +66,12 @@ public class ColumnarBulkDataSanitizerImpl implements BulkDataSanitizer {
...
+    @Setter(onMethod_ = @VisibleForTesting)
+    @Inject
+    RuleSet bulkDataRules;
+
     @Override
     public byte[] sanitize(@NonNull InputStreamReader reader,
-                           @NonNull BulkDataRules bulkDataRules,
                            @NonNull Pseudonymizer pseudonymizer) throws IOException {
 
         if (!(bulkDataRules instanceof CsvRules)) {
```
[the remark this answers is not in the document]
aperez-worklytics: Another option, just use a injected constructor:

```
 @AssistedInject
    public ColumnarBulkDataSanitizerImpl(@Assisted ObjectMapper mapper,
UrlSafeTokenPseudonymEncoder urlSafeTokenPseudonymEncoder,
 PseudonymizerImplFactory pseudonymizerImplFactory,
RuleSet bulkDataRules) {
if (! (ruleSet instanceof CsvRules)) {
            throw new IllegalArgumentException("Configured RuleSet are not ColumnarRules");
        }

this.bulkDataRules = bulkDataRules;
this... = ....
    }
```

No need to modify DI and the factory but we could ensure that the object is built with the right dependencies and no need to check that as part of `sanitize` method.

java/core/src/main/java/co/worklytics/psoxy/storage/impl/RecordBulkDataSanitizerImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    public RecordBulkDataSanitizerImpl(@Assisted RecordRules rules) {
+        this.rules = rules;
+    }
+
+
+    @Override
+    public byte[] sanitize(InputStreamReader reader, Pseudonymizer pseudonymizer) throws IOException {
+
+        if (!(rules instanceof RecordRules)) {
+
```
aperez-worklytics, line 47:
```suggestion
```

java/gateway-core/src/main/java/com/avaulta/gateway/rules/MultiTypeBulkDataRules.java
```diff
@@ -0,0 +1,36 @@
...
+     * eg, /export/{week}/data_{shard}.csv -> CsvRules
+     *
+     * if provided, has the effect of pathRegex = "^/export/[^/]+/data_[^/]+\.csv$"
+     *
+     * files that trigger proxy instance but match NONE of the templates will not be processed
+     *
+     *
+     * q: support multiple jumps?
+     *
+     * q: what if file matches multiple path templates? pick lexicographically first?
```
aperez-worklytics, line 30: Do we need to cover that? I'll raise an exception in such case

java/gateway-core/src/main/java/com/avaulta/gateway/rules/transforms/Transform.java
```diff
@@ -54,6 +54,8 @@ public abstract class Transform {
      * specifies fields within content that identify value to be transformed
      * <p>
      * supported for CSV
+     *
+     * @Deprecated - CSV now uses columnar rules
```
aperez-worklytics, line 58: Ok, but please annotate it with `@deprecated` too

java/gateway-core/src/main/java/com/avaulta/gateway/rules/transforms/RecordTransform.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * would prefer that both
+ *
+ *  - redact: "foo"
+ *
+ * and
+ *  - redact
+ *    path: "foo"
+ *
+ * would be supported as equivalent ... but can't get that to work with Jackson YAML
```
aperez-worklytics, line 26: That *could* be done with a custom serializer, but I think as it is now is fine as we see the subtypes and the fields.

aperez-worklytics (approved):
Some code suggestion, but no blockers

java/core/src/main/java/co/worklytics/psoxy/PsoxyModule.java
```diff
@@ -248,6 +249,12 @@ Pseudonymizer pseudonymizer(PseudonymizerImplFactory factory, ConfigService conf
         return factory.create(factory.buildOptions(config, ruleSet.getDefaultScopeIdForSource()));
     }
 
+    @Provides
+    @Singleton
+    PathTemplateUtils pathTemplateUtils() {
+        return new PathTemplateUtils();
```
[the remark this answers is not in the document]
eschultink: yeah, i could be - but I think better to avoid static dependencies. Limits scope of refactor if we add dep in future

java/gateway-core/src/main/java/com/avaulta/gateway/rules/MultiTypeBulkDataRules.java
```diff
@@ -0,0 +1,36 @@
...
+@NoArgsConstructor //for Jackson
+@Getter
+@EqualsAndHashCode
+@JsonInclude(JsonInclude.Include.NON_NULL) //NOTE: despite name, also affects YAML encoding
+public class MultiTypeBulkDataRules implements RuleSet {
+
+    /**
+     * map of file path templates to rules for matching files
+     *
+     * eg, /export/{week}/data_{shard}.csv -> CsvRules
```
jlorper, line 21: I'd just be explicit with the names -> 
`ColumnarRules` -> `CSVRules`
`RecordRules` -> `NDJSONRules`

Actually both are `RecordRules` of different record types.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/PathTemplateUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class PathTemplateUtils {
+
+
+    //TODO: improve this; some special chars outside of {} are not accounted for
+    private static final String SPECIAL_CHAR_CLASS = "[\\.\\^\\$\\<\\>\\*\\+\\[\\]\\(\\)\\+\\-\\=\\?\\!]";
+
+    private static final String REGEX_ALPHANUMERIC_PATH_PARAM = "\\{([A-Za-z][A-Za-z0-9]*)\\}";
```
jlorper, line 14: just FYI, not sure if more readable or not, but these Posix "shortcuts" exists in java regexp
`\\{(\p{Alpha}\p{Alnum}*)\\}`

java/gateway-core/src/test/java/com/avaulta/gateway/rules/MultiTypeBulkDataRulesTest.java
```diff
@@ -0,0 +1,82 @@
...
+    public void toYaml() {
+
+        final String EXPECTED = "---\n" +
+            "fileRules:\n" +
+            "  /export/{week}/data_{shard}.csv:\n" +
+            "    columnsToPseudonymize:\n" +
+            "    - \"email\"\n" +
+            "    delimiter: \",\"\n" +
+            "    pseudonymFormat: \"JSON\"\n" +
+            "  /export/{week}/index_{shard}.ndjson:\n" +
```
jlorper, line 50: :warning: so we are not expecting gzipped content??? should we account for that too? like `export/{week}/index_{shard}.ndjson.gz` ?
I think we should, and explicitly put format = CSV or format = NDJSON always to choose one or another
and reader check if gzipped or not.

jlorper (changes requested):
Some concerns about OOO issues and gzipped files.
Also few suggestions here and there.

java/core/src/main/java/co/worklytics/psoxy/rules/RulesUtils.java
```diff
@@ -117,6 +129,44 @@ public List<StorageHandler.ObjectTransform> parseAdditionalTransforms(ConfigServ
         }
     }
 
+    public Optional<String> getDefaultScopeIdFromRules(com.avaulta.gateway.rules.RuleSet ruleSet) {
+        if (ruleSet instanceof RuleSet) {
+            return Optional.ofNullable(((RuleSet) ruleSet).getDefaultScopeIdForSource());
```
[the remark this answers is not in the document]
eschultink: if think if `null`, then`instance Ruleset` is `false` right? Although it shouldn't be `null`

java/core/src/main/java/co/worklytics/psoxy/rules/RulesUtils.java
```diff
@@ -117,6 +129,44 @@ public List<StorageHandler.ObjectTransform> parseAdditionalTransforms(ConfigServ
...
+        } else {
+            return Optional.empty();
+        }
+    }
 
+    static Map<String, String> defaultScopeIdBySource;
```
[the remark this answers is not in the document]
eschultink: yeah, this should go away in next version. scopeId only needed for legacy format pseudonyms, which is only a handful of early users - and no one new.

if missing, defaults to the value of `source` itself, which is correct for long-tail.

java/core/src/main/java/co/worklytics/psoxy/storage/impl/RecordBulkDataSanitizerImpl.java
```diff
@@ -0,0 +1,123 @@
...
+                } catch (UnmatchedPseudonymization e) {
+                    log.warning("Skipped record due to UnmatchedPseudonymizaiton: " + e.getPath());
+                }
+            }
+        }
+
+        outputStreamWriter.flush();
+        outputStreamWriter.close();
+
+        return outputStream.toByteArray();
```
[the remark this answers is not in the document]
eschultink: yeah, dealing with that on a separate branch. 

it's non-trivial to do streaming read/write for both aws/gcs.  Gcs actually seems more directly in SDK than what AWS offers.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/MultiTypeBulkDataRules.java
eschultink: yeah. Have gotten rid of CSVRules in favor of just ColumnarRules; can have `\t` as delimiter in theory to support TSV if we care.

But then yes - RecordRules could support CSV/TSV format and provide equivalent functionality. It's just not as simple in the simple case.  We could move to it from ColumnarRules in the future major version though if we want to consolidate.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/PathTemplateUtils.java
eschultink: yeah, imho not sure is more readable.

java/gateway-core/src/test/java/com/avaulta/gateway/rules/MultiTypeBulkDataRulesTest.java
eschultink: these are just example rules. I have a task to test out `gzip` on separate branch. I think for GCS it will just magically work; when you open the file stream, GCS client will by default detect compression and decompress it (unless you disable this behavior) - but need to test that. 

And then need to write compressed content back, which we currently don't support - but is certainly worth doing.

java/core/src/main/java/co/worklytics/psoxy/PsoxyModule.java
```diff
@@ -244,8 +248,27 @@ ParameterSchemaUtils parameterSchemaUtils() {
     }
 
     @Provides
-    Pseudonymizer pseudonymizer(PseudonymizerImplFactory factory, ConfigService config, co.worklytics.psoxy.rules.RuleSet ruleSet) {
-        return factory.create(factory.buildOptions(config, ruleSet.getDefaultScopeIdForSource()));
+    Pseudonymizer pseudonymizer(PseudonymizerImplFactory factory, ConfigService config, RulesUtils rulesUtils, com.avaulta.gateway.rules.RuleSet ruleSet) {
```
jlorper, line 251: I just realized this is weird. Why are we providing a `Pseudonymizer` but passing a factory inside??'
why do we have this at all, shouldn't all usages of `Provider<Pseudonymizer>` be gone in favor of constructing it explicitly like we do in `CommonRequestHandler`??

java/core/src/main/java/co/worklytics/psoxy/rules/RulesUtils.java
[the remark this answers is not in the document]
jlorper: you are right, so this can be straight, the casting is not needed
```suggestion
            return Optional.of(ruleSet.getDefaultScopeIdForSource());
```

java/core/src/main/java/co/worklytics/psoxy/storage/impl/RecordBulkDataSanitizerImpl.java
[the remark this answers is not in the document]
jlorper: OK, remember to do those changes there

jlorper (approved):
some comments, no blockers. Main thing I saw is the pseudonymizer injection

PR #562, opened 2023-10-23
process bulk GCS with streams

eschultink:
### Features
 - streaming reads/writes for GCS bulk case

### Change implications

---
  - https://app.asana.com/0/0/1205552701863598

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/StorageEventRequest.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/StorageEventRequest.java
@@ -5,6 +5,7 @@ import lombok.Getter;
 import lombok.NonNull;
 
 import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
 
 /**
  * Request received when a change is done in some storage service, such as "new" object created or "updated" object
@@ -20,9 +21,16 @@ public class StorageEventRequest {
     InputStreamReader readerStream;
 
     /**
+     * Stream to write output to
+     */
+    @NonNull
+    OutputStreamWriter destinationStream;
+
+    // REST OF THIS IS INFORMATION FOR SANITIZER TO INFORM SANITIZATION LOGIC BASED ON SOURCE/DESTINATION
+
+    /**
      * Name of the bucket where the object is. Example: "mybucket"
      */
-
     @NonNull
     String sourceBucketName;
 
@@ -44,6 +52,4 @@ public class StorageEventRequest {
      */
     @NonNull
     String destinationObjectPath;
-
-
 }
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/StorageEventRequest.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/StorageEventRequest.java
@@ -5,8 +5,8 @@ import lombok.Getter;
 import lombok.NonNull;
 import lombok.With;
 
-import java.io.InputStreamReader;
-import java.io.OutputStreamWriter;
+import java.io.Reader;
+import java.io.Writer;
 
 /**
  * Request received when a change is done in some storage service, such as "new" object created or "updated" object
@@ -16,14 +16,14 @@ import java.io.OutputStreamWriter;
 @Getter
 public class StorageEventRequest {
     /**
-     * Stream where the object can be read
+     * reader to read input from
      */
-    InputStreamReader readerStream;
+    Reader sourceReader;
 
     /**
-     * Stream to write output to
+     *  to write output to
      */
-    OutputStreamWriter destinationStream;
+    Writer destinationWriter;
 
     // REST OF THIS IS INFORMATION FOR SANITIZER TO INFORM SANITIZATION LOGIC BASED ON SOURCE/DESTINATION
 
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/StorageEventRequest.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/StorageEventRequest.java
@@ -3,6 +3,7 @@ package co.worklytics.psoxy.gateway;
 import lombok.Builder;
 import lombok.Getter;
 import lombok.NonNull;
+import lombok.With;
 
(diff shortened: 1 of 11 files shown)


java/impl/aws/src/main/java/co/worklytics/psoxy/S3Handler.java
```diff
@@ -71,21 +72,18 @@ StorageEventResponse process(String importBucket, String sourceKey, StorageHandl
             return null;
         }
 
+        S3Object s3Object = s3Client.getObject(new GetObjectRequest(importBucket, sourceKey));
         try (InputStream objectData = s3Object.getObjectContent();
              BOMInputStream is = new BOMInputStream(objectData);
-             InputStreamReader reader = new InputStreamReader(is, StandardCharsets.UTF_8)) {
+             InputStreamReader reader = new InputStreamReader(is, StandardCharsets.UTF_8);
+             PipedOutputStream outputStream = new PipedOutputStream();
+             OutputStreamWriter writer = new OutputStreamWriter(outputStream);
```
jlorper, line 80: > or top efficiency, consider wrapping an OutputStreamWriter within a BufferedWriter so as to avoid frequent converter invocations. For example:
> Writer out  = new BufferedWriter(new OutputStreamWriter(System.out));
https://docs.oracle.com/javase/8/docs/api/java/io/OutputStreamWriter.html

java/impl/aws/src/main/java/co/worklytics/psoxy/S3Handler.java
```diff
@@ -71,21 +72,18 @@ StorageEventResponse process(String importBucket, String sourceKey, StorageHandl
...
+             InputStream processedStream = new PipedInputStream(outputStream)) {
 
-            StorageEventRequest request = storageHandler.buildRequest(reader, importBucket, sourceKey, transform);
+            StorageEventRequest request = storageHandler.buildRequest(reader, writer, importBucket, sourceKey, transform);
 
             storageEventResponse = storageHandler.handle(request, transform.getRules());
```
jlorper, line 85: I might be totally wrong, IO in Java is always confusing:
I feel that this should be invoked after `s3Client.putObject` (obviously can't because it uses the destination objects)
but if handle does a flush it seems data will be held in memory until `putObject` starts processing, and might OOO?Will held the `writer` and `processedStream` until L98 flush?

java/core/src/main/java/co/worklytics/psoxy/gateway/StorageEventRequest.java
```diff
@@ -3,26 +3,33 @@
...
      */
-
-    @NonNull
     InputStreamReader readerStream;
 
     /**
-     * Name of the bucket where the object is. Example: "mybucket"
+     * Stream to write output to
      */
+    OutputStreamWriter destinationStream;
```
jlorper, line 26: OK, but it is kinda weird that this "request" class contains readers & writers, it's very attached to the implementation. But can't provide a better approach.
passing around IO is always tricky if closed/flushed when it is not supposed to happen

java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -76,8 +73,9 @@ public ColumnarBulkDataSanitizerImpl(@Assisted ColumnarRules rules) {
     }
 
     @Override
-    public byte[] sanitize(@NonNull InputStreamReader reader,
-                           @NonNull Pseudonymizer pseudonymizer) throws IOException {
+    public void sanitize(@NonNull InputStreamReader reader,
```
jlorper, line 76: > For top efficiency, consider wrapping an InputStreamReader within a BufferedReader. For example:
> BufferedReader in  = new BufferedReader(new InputStreamReader(System.in));
https://docs.oracle.com/javase/8/docs/api/java/io/InputStreamReader.html

```suggestion
    public void sanitize(@NonNull Reader reader,
```
consider leaving implementation to caller?

java/core/src/test/java/co/worklytics/psoxy/storage/StorageHandlerTest.java
```diff
@@ -72,7 +74,10 @@ void buildRequest() {
         when(config.getConfigPropertyAsOptional(eq(BulkModeConfigProperty.OUTPUT_BASE_PATH)))
             .thenReturn(Optional.empty());
 
-        StorageEventRequest request = handler.buildRequest(mock(InputStreamReader.class), "bucket-in", "directory/file.csv", handler.buildDefaultTransform());
+
+
+        StorageEventRequest request = handler.buildRequest(mock(InputStreamReader.class),
```
jlorper, line 79: I'd try testing with a big file to see how it behaves mem wise, but not a big deal

jlorper (approved):
some suggestions, no blockers.

java/impl/aws/src/main/java/co/worklytics/psoxy/S3Handler.java
eschultink: i agree with you, yeah ....

java/core/src/main/java/co/worklytics/psoxy/gateway/StorageEventRequest.java
eschultink: yeah, it should be split into two things

java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
eschultink: InputStreamReader still an interface ... but yeah, could be more generic

java/core/src/test/java/co/worklytics/psoxy/storage/StorageHandlerTest.java
eschultink: yeah, will have to; hard to replicate in tests exactly, but could try

PR #568, opened 2023-10-27
S160 return source always if not in rules

jlorper:
### Fixes
[lack of source in rules not fallbacking to env source](https://app.asana.com/0/1205719748958530/1205825821497086)

### Change implications


aperez-worklytics (commented):
Ok but better a double check with @eschultink

java/core/src/main/java/co/worklytics/psoxy/PsoxyModule.java
```diff
@@ -263,11 +264,13 @@ Pseudonymizer pseudonymizer(PseudonymizerImplFactory factory, ConfigService conf
                         .map(PseudonymImplementation::valueOf)
                         .map(implementation -> Objects.equals(implementation, PseudonymImplementation.LEGACY))
                         .orElse(false);
-                    if (legacy) {
-                        return rulesUtils.getDefaultScopeIdFromSource(config.getConfigPropertyOrError(ProxyConfigProperty.SOURCE));
-                    } else {
-                        return ""; //should only be used in legacy case
+                    String source = config.getConfigPropertyOrError(ProxyConfigProperty.SOURCE);
```
eschultink, line 267: this is a breaking change, as it will REQUIRE that `SOURCE` property is configured in non-legacy cases, even though has no function. I don't know why we want to introduce something that could break proxy deployments for no purpose.  `defaultScopeIdFromSource` is required ONLY for legacy cases, of which there are very few people - I don't like to introduce brittleness to everyone else. 

Existing code will still throw fatal exception if no `SOURCE`.

eschultink (approved):

PR #569, opened 2023-10-31
improve jira/zoom docs 

eschultink:
### Features
 - clarify Jira docs; clarify that terminal is required, `curl` doesn't work out of the box on Windows ...
 - screenshots, alignment on Zoom docs

### Change implications


docs/sources/atlassian/jira/README.md
```diff
@@ -17,6 +17,9 @@ And following granular scopes:
   - `read:group:jira`: for retrieving group members
   - `read:avatar:jira`: for retrieving group members
 
+You will need a web browser and a terminal with `curl` available (such as macOS terminal,
```
khnumdev, line 20: I'll add a note that `curl` is installed by default in Windows 10/11 -> https://curl.se/windows/microsoft.html.

Tested with my own Windows 11 and is installed in the powershell terminal

khnumdev (approved):

aperez-worklytics (approved):

docs/sources/atlassian/jira/README.md
eschultink: ok, interesting ... one Jira admin said that the commands didn't work for him, and had to enlist help from someone with a Mac.

aperez-worklytics: ok, let met check that later. Maybe PS version is an alias of `invoke-weburl` command, so the parameters may differ about what we have in the docs.

khnumdev: Yeah, from terminal/PS I had the issue described in the link:

```
cmdlet Invoke-WebRequest at command pipeline position 1
Supply values for the following parameters:
Uri:
```

So I think that adding that link as part of the documentation could help on this.

PR #572, opened 2023-11-13
fix java21

eschultink:
### Fixes
 - build breaks with jdk >= 21, which is now the default

as with past breakages in jdk 19/20, lombok is culprit: https://projectlombok.org/changelog - and in this case first fix was 1 day after the jdk 21 release, so 🤷 

--> change req'd is deeper than that; jdk 21 has broken Mockito, thus broken our tests; https://github.com/Worklytics/psoxy/actions/runs/6856732245/job/18644471668?pr=572 ; fix requires upgrade to Mockito v5, which has issues with java8 ...

### Change implications

 - dependencies added/changed? **yes**, bumped lombok from 1.18.28 --> 1.18.30; mockito to v5

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -23,8 +23,8 @@
         <dependency.google-cloud-bom.version>0.198.0</dependency.google-cloud-bom.version>
         <dependency.google-http-client.version>1.43.3</dependency.google-http-client.version>
         <dependency.google-auth-library-oauth2-http.version>1.18.0</dependency.google-auth-library-oauth2-http.version>
-        <dependency.junit-jupiter.version>5.9.1</dependency.junit-jupiter.version>
-        <dependency.mockito-junit-jupiter.version>4.11.0</dependency.mockito-junit-jupiter.version>
+        <dependency.junit-jupiter.version>5.10.1</dependency.junit-jupiter.version>
+        <dependency.mockito-junit-jupiter.version>5.0.0</dependency.mockito-junit-jupiter.version>
         <dependency.json-path.version>2.8.0</dependency.json-path.version>
         <dependency.bettercloud-vault-java-driver>5.1.0</dependency.bettercloud-vault-java-driver>
     </properties>
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -12,7 +12,7 @@
     <properties>
         <revision>0.4.41</revision>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <dependency.lombok.version>1.18.28</dependency.lombok.version>
+        <dependency.lombok.version>1.18.30</dependency.lombok.version> <!-- 1.18.30 is min compatible with jdk 21+ -->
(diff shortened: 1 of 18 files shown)


eschultink:
ugh, that fixed java 17-21, but broke 8 and 11.

eschultink:
ok, w hacks from https://github.com/Worklytics/psoxy/pull/572/commits/f4b38b9e8580922ee9a77c00e4dfc60b0e34ff02 and https://github.com/Worklytics/psoxy/pull/572/commits/5ced9cf906bcd390befa43cd45c3d35fd1d7472d all LTS versions of Java are working.

Java 20 though is broken; which while it's out-of-support, it would have been the default version of java people installed as of Sept 2023, less than 60 days ago ... so don't know that we really want to break it.  Plenty of people who think they have up-to-date machines are going to get build errors.

java/pom.xml
```diff
@@ -23,12 +23,25 @@
...
         <dependency.json-path.version>2.8.0</dependency.json-path.version>
         <dependency.bettercloud-vault-java-driver>5.1.0</dependency.bettercloud-vault-java-driver>
     </properties>
 
+    <dependencyManagement>
+        <dependencies>
+            <!-- https://mvnrepository.com/artifact/org.junit/junit-bom -->
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
```
jlorper, line 37: :+1: was going to suggest using the bom

jlorper:
Maybe we should only support LTS versions and fail otherwise

java/pom.xml
```diff
@@ -15,7 +15,7 @@
         <dependency.lombok.version>1.18.30</dependency.lombok.version> <!-- 1.18.30 is min compatible with jdk 21+ -->
         <dependency.dagger.version>2.40.5</dependency.dagger.version>
         <dependency.jackson.version>2.15.2</dependency.jackson.version>
-        <dependency.apache-commons-lang3.version>3.12.0</dependency.apache-commons-lang3.version>
+        <dependency.apache-commons-lang3.version>3.13.0</dependency.apache-commons-lang3.version> <!-- July 2023 release, doesn't actually have constants for java after 17 yet -->
```
jlorper, line 18: it does, doesn't it?
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/JavaVersion.html#JAVA_21

jlorper (approved):

java/pom.xml
eschultink: 🤔 intellij must have had it cached for me, still didn't recognize it

eschultink:
> Maybe we should only support LTS versions and fail otherwise

--> no, because if you just do `brew install openjdk` or something, you get the latest - not necessarily latest LTS. So if someone did that Sept 15th, they have java 20 right now, because java 21 was only released the same day java 20 support ended.

aperez-worklytics (approved):

PR #576, opened 2023-11-16
improve health check

eschultink:
### Features
 - reduce noise from health checks in logs
 - simplify/clarify version info in health checks 

### Change implications

 - something to note in `CHANGELOG.md`? **not really, healthcheck schema changes, but that shouldn't be something anyone is depending on (and healthcheck consumer POJOs designed to be flexible, deal with unexpected fields)**


java/core/src/main/java/co/worklytics/psoxy/HealthCheckResult.java
```diff
@@ -18,36 +19,23 @@
 @AllArgsConstructor
 public class HealthCheckResult {
 
-    @Builder.Default
-    String version = "rc-v0.4.41";
+    String javaSourceCodeVersion;
```
aperez-worklytics, line 22:
```suggestion
    private String javaSourceCodeVersion;
```

aperez-worklytics (approved):

PR #587, opened 2023-11-28
nodejs reference implementation of pseudonymization logic 

eschultink:
### Features
  - reference implementation of pseudonymization logic in nodejs; left as draft, first experiment 

### Change implications

 - dependencies added/changed? **yes - new npm script, depends on npm modules, including `ava` as 3rd party testing fw**


tools/js-reference/lib/lib.js
```diff
@@ -0,0 +1,34 @@
+const crypto = require('crypto');
```
davidfq, line 1: > npm WARN deprecated crypto@1.0.1: This package is no longer supported. It's now a built-in Node module. If you've depended on crypto, you should switch to the one that's built-in.

```suggestion
const crypto = require('node:crypto');
```

built-in since v16 (minimum we require); tests passing

eschultink:
@davidfq - removed from draft; I think worth having in mainline for a reference

davidfq (approved):

PR #589, opened 2023-11-29
slack bulk rules

eschultink:
### Features
 - example of bulk rules for slack, including tests against some example files

### Change implications

 - something to note in `CHANGELOG.md`? **prob min version for any of these multi-file rules to work in practice**


java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
```diff
@@ -35,7 +36,7 @@ public class StorageHandler {
     Pseudonymizer pseudonymizer;
 
     @Inject
-    BulkDataRules defaultRuleSet;
+    RuleSet defaultRuleSet;
```
eschultink, line 39: this change REQUIRED before `MultiTypeBulkDataRules` will actually work in practice, bc `MutliTypeBulkDataRules` don't implement `BulkDataRules`  (they could, but that's also a code change ...

docs/sources/slack/example-bulk/sanitized/messages-D06TX2RP0.json
```diff
@@ -0,0 +1,19 @@
+{
```
jlorper, line 1: shoudln't output be NDJSON too? Or left expanded for readabilty?

jlorper (approved):
Just one doubt about the content output format

PR #597, opened 2023-12-05
gzip support for bulk file processing, both in and out

eschultink:
### Fixes
 - dropped expectation of leading `/` in multi-file-type bulk rules, as requires big refactor and not well defined how should interact with `INPUT_BASE_PATH`/`OUTPUT_BASE_PATH` (used to support going in/out of same bucket, which some people do)

### Features
 - gzip output if `Content-Encoding` of input was `gzip`

### Change implications


docs/sources/slack/discovery-bulk-hierarchical.yaml
```diff
@@ -1,5 +1,5 @@
 fileRules:
```
eschultink, line 1: better name for this?

java/impl/aws/src/main/java/co/worklytics/psoxy/S3Handler.java
```diff
@@ -75,16 +75,18 @@ StorageEventResponse process(String importBucket, String sourceKey, StorageHandl
...
+             ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
 
-            StorageEventRequest request = storageHandler.buildRequest(reader, writer, importBucket, sourceKey, transform);
+            StorageEventRequest request = storageHandler.buildRequest(null, null, importBucket, sourceKey, transform);
+
+            boolean isCompressed = Optional.ofNullable(s3Object.getObjectMetadata().getContentEncoding())
+                .map(s -> s.contains("gzip"))
+                .orElse(false);
+
+            request = request.withCompressOutput(isCompressed).withDecompressInput(isCompressed);
```
aperez-worklytics, line 86: Invert?
```suggestion
            request = request.withDecompressInput(isCompressed).withCompressOutput(isCompressed);
```

aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
```diff
@@ -72,6 +78,38 @@ public String getMetaDataKey() {
...
+            Writer writer= new BufferedWriter(streamWriter, BUFFER_SIZE) //q: BufferedWriter needed if we're already buffering in GZIPOutputStream?
+        ) {
+
+            request = request
+                .withSourceReader(reader)
+                .withDestinationWriter(writer);
+
+            StorageEventResponse storageEventResponse = this.handle(request, transform.getRules());
+
+            log.info("Writing to: " + storageEventResponse.getDestinationBucketName() + "/" + storageEventResponse.getDestinationObjectPath());
```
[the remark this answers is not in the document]
eschultink: I don't find that more readable; you have to look to the end and count order to try to figure out what is going to be written there.  Until/unless we move to string interpolation, I prefer this.

imho, String.format is when you need real formatting.

PR #602, opened 2023-12-06
fix pseudonymize regex matches encoding

eschultink:
### Fixes
  - more proper encoding of pseudonyms out of `PseudonymizeRegexMatches`; will go through either Legacy or `URLSafePseudonymEncoder` path
  - fix `URLSafePseudonymEncoder` to let it decode legacy-encoded pseudonyms

### Features
  - for completeness, decoding support in `LegacyPseudonymEncoder`; although shouldn't be needed

### Change implications


java/gateway-core/src/main/java/com/avaulta/gateway/pseudonyms/impl/LegacyPseudonymTokenEncoder.java
```diff
@@ -0,0 +1,89 @@
...
+import com.avaulta.gateway.pseudonyms.PseudonymEncoder;
+import lombok.Builder;
+import lombok.Value;
+import lombok.extern.java.Log;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Base64;
+
+/*
+ * encodes legacy pseudonyms as tokens, starting with "l~" followed by hash value (presumed to be
```
jlorper, line 13: I guess all this comment is wrong, I don't see any `l~` prefix anywhere

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -490,19 +493,11 @@ public MapFunction getPseudonymizeRegexMatches(Transform.PseudonymizeRegexMatche
...
-                if (pseudonymizedIdentity.getReversible() != null) {
-                    if (getPseudonymizer().getOptions().getPseudonymImplementation() == PseudonymImplementation.LEGACY) {
-                        //exploits that already reversibly encoded, including prefix
-                        log.warning("Using transform PseudonymizeRegexMatches, with reversible==true; this is NOT supported for LEGACY pseudonym implementation, so non-reversible pseudonym encoded");
-                        pseudonymizedString = UrlSafeTokenPseudonymEncoder.TOKEN_PREFIX + pseudonymizedIdentity.getHash();
-                    } else {
-                        pseudonymizedString = pseudonymizedIdentity.getReversible();
-                    }
+
+                if (getPseudonymizer().getOptions().getPseudonymImplementation() == PseudonymImplementation.LEGACY) {
```
jlorper, line 497: I just saw there is a similar usage in ColumnarBulkDataSanitizerImpl L297.
Also legacyTokenEncoder could "embed" HashUtils class and get rid of that, deal with presumably legacy  `PseudonymizedIdentity` directly, rather than this one calling HashUtils internally.

jlorper (approved):

aperez-worklytics (approved):

PR #607, opened 2023-12-11
Teams: Not redact odata.type for messages (both chat and channels)

aperez-worklytics:
`odata.type` is not redacted (it was only for *no user ids* rules) to allow getting the type of the event detail.

### Change implications

     obviously a no-op? **no**
     change **no**


P-A-R-U-S (approved):
OK. Look good. 
I don't know **odata.type** should not be redacted.

aperez-worklytics:
> OK. Look good. I don't know **odata.type** should not be redacted.

Yeah thanks, it is required for https://github.com/Worklytics/evalengin/pull/6222

PR #608, opened 2023-12-11
S162 MSFT Teams via Proxy - Don't we need both Chat.Read and Chat.ReadBasic

P-A-R-U-S:
MS Team Proxuy don't we need both Chat.Read.All and Chat.ReadBasic.All. 
Chat.Read.All  will include  Chat.ReadBasic.All permissions.

**See:** https://graphpermissions.merill.net/permission/Chat.Read.All

### Features
[MSFT Teams via Proxy: do we need both Chat.Read and Chat.ReadBasic ](https://app.asana.com/0/1206131962756230/1206131962756225/f)

### Change implications

    * **_update terraform configuration._**


eschultink:
@P-A-R-U-S proxy branches should be branched from the latest `rc-`; and when PR open, the lastest `rc-` should be the base.  I assume that's why you have conflicts?  try creating new branch from the `rc-` and then cherry picking your commits specific to this directly on to that branch - then creating PR for that against the rc-

PR #609, opened 2023-12-11
S162 MS Team Proxy - Don't need both Chat.Read.All and Chat.ReadBasic.All. Chat.Read.All is sufficient.

P-A-R-U-S:
MS Team Proxuy don't we need both Chat.Read.All and Chat.ReadBasic.All. 
Chat.Read.All  will include  Chat.ReadBasic.All permissions.

**See:** https://graphpermissions.merill.net/permission/Chat.Read.All

### Features
[MSFT Teams via Proxy: do we need both Chat.Read and Chat.ReadBasic ](https://app.asana.com/0/1206131962756230/1206131962756225/f)

### Change implications

    * **_update terraform configuration._**


aperez-worklytics (approved):
ok, as a double check I'd deploy the proxy and run the tests to ensure all requests can be done

PR #613, opened 2023-12-12
more validation of bulk files before processing 

eschultink:
### Features
 - warn in logs if extension is `.gz`, but `Content-Encoding` is not `gzip`
 - refactor StorageHandler interface to be cleaner, 

### TODO
 - [x] test in actual AWS/GCS deployment scenarios

### Change implications

---
  - https://app.asana.com/0/0/1206122727528684

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
@@ -13,6 +13,7 @@ import lombok.*;
 import lombok.extern.java.Log;
 import org.apache.commons.lang3.StringUtils;
 
+import javax.annotation.Nullable;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import java.io.*;
@@ -32,6 +33,8 @@ public class StorageHandler {
 
     // as writing to remote storage, err on size of larger buffer
     private static final int DEFAULT_BUFFER_SIZE = 65_536; //64 KB
+
+    public static final String CONTENT_ENCODING_GZIP = "gzip";
 
     @Inject
     ConfigService config;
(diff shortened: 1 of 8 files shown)


java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
```diff
@@ -291,6 +275,88 @@ public Optional<BulkDataRules> getApplicableRules(RuleSet rules, String sourceOb
...
+                     Supplier<InputStream> inputStreamSupplier) {
+        int bufferSize = getBufferSize();
+        try (
+            InputStream decompressedStream = request.getDecompressInput() ? new GZIPInputStream(inputStreamSupplier.get(), bufferSize) : inputStreamSupplier.get();
+            BufferedReader reader = new BufferedReader(new InputStreamReader(decompressedStream, StandardCharsets.UTF_8), bufferSize);
+            OutputStream out = new ByteArrayOutputStream()
+        ) {
+
+            StringBuffer firstLines = new StringBuffer();
+            while (firstLines.length() < 5) {
```
aperez-worklytics, line 302: This is mostly for checking if the buffer has less than 5 characters length; maybe I'd change this to `firstLine` and just do the break after reading it or if the idea is to read first 5 lines, just count the lines read instead of `length()`

java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
```diff
@@ -291,6 +275,88 @@ public Optional<BulkDataRules> getApplicableRules(RuleSet rules, String sourceOb
...
+                    break;
+                }
+                firstLines.append(line);
+            }
+
+            Supplier<InputStream> firstLinesSupplier = () -> new ByteArrayInputStream(firstLines.toString().getBytes());
+
+            this.process(request, transform, firstLinesSupplier, () -> out);
+
+            return true;
```
aperez-worklytics, line 314: The way that `validate` fails is going to be if `process` raise an exception, right? Then, I'd drop the `return true` even not used on the caller; just return nothing and leave the exception flow.

java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
```diff
@@ -291,6 +275,88 @@ public Optional<BulkDataRules> getApplicableRules(RuleSet rules, String sourceOb
...
+        }
+    }
+
+
+    int getBufferSize() {
```
aperez-worklytics, line 319: Don't recall if used outside, but if not, `private` ?

aperez-worklytics (changes requested):

davidfq (approved):
given you're aware of the Web UI upload use-case, then OK

java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
eschultink: copilot fail

aperez-worklytics (approved):

PR #622, opened 2023-12-22
MSFT Teams: endpoints for Attendance Reports

aperez-worklytics:
- Rules for supporting fetching Attendance Reports (list and a particular report) for OnlineMeetings.

### Features
- [Teams: Fetch attendance reports](https://app.asana.com/0/1206131962756230/1206220295080462)

### Change implications

     obviously a no-op? `Outlook calendar` will require to grant new scopes for online meetings
     change


P-A-R-U-S (approved):
Looks good.

docs/sources/microsoft-365/README.md
```diff
@@ -50,12 +50,12 @@ Neither your proxy instances nor Worklytics ever hold any API key or certificate
...
+|--------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Active Directory                                                                                       | `User.Read.All` `Group.Read.All`                                                                                                                                                         |
+| Calendar                                                                                               | `User.Read.All` `Group.Read.All` `OnlineMeetings.Read.All` `OnlineMeetingArtifact.Read.All` `Calendars.Read` `MailboxSettings.Read`                                                                                      |
```
eschultink, line 56: also for the `Calendar`? this should go in CHANGELOG, as customers using that connector who `apply` after this will see addition perms added to the app, and probably need their MSFT 365 admin to make the grant again, right??

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -264,7 +264,8 @@ locals {
         "Chat.Read.All",
         "ChannelMessage.Read.All",
         "CallRecords.Read.All",
-        "OnlineMeetings.Read.All"
+        "OnlineMeetings.Read.All",
+        "OnlineMeetingArtifact.Read.All"
```
eschultink, line 268: why only in teams, when in docs you added to `calendar` to? which is correct?

eschultink (changes requested):
something is inconsistent here. 

`OnlineMeetingArtifact.Read.All` is either needed for calendar or not. If so, it needs to be added under `worklytics-connector-specs` module; if not, it needs to be removed from `README.md`s 

Also, probably something needs to be done to root `README.md` (adding this perm to either `Teams` or both Calendar + Teams.

docs/sources/microsoft-365/msft-teams/msft-teams.yaml
```diff
@@ -146,7 +146,57 @@ endpoints:
     transforms:
       - !<redact>
         jsonPaths:
-          - "$..user.displayName"
+          - "$.['@odata.context']"
+      - !<redact>
+        jsonPaths:
+          - "$..emailAddress"
```
jlorper, line 152: this should be pseudonymize, right? You are redacting all emailAddress here. Better to have 2 identities to fall back just in case

jlorper (changes requested):
Besides Erik's comments I would also pseudonymize emails, not redact them.

docs/sources/microsoft-365/README.md
aperez-worklytics: Yeah missing in the spec and changelog

java/core/src/main/java/co/worklytics/psoxy/rules/msft/PrebuiltSanitizerRules.java
```diff
@@ -26,7 +26,7 @@ public class PrebuiltSanitizerRules {
             .regex("^https://graph.microsoft.com/(.*)$")
             .build();
     static final Transform REDACT_ODATA_CONTEXT = Transform.Redact.builder()
-            .jsonPath("$.['@odata.context']")
+            .jsonPath("$..['@odata.context']")
```
aperez-worklytics, line 29: Change this, which affects to rest of MSFT yamls (because it is used across all MSFT-365). This is to redact all *context*, not only the one in the top of the JSON.

eschultink (approved):

jlorper (approved):

PR #625, opened 2023-12-26
GitHub Enterprise Server support

aperez-worklytics:
- Added enterprise server connector support. Only *host* is needed here, leaving cloud (enterprise or free/pro/team) host fixed, as it is not required to customize.
  - Rename host parameter to remark is the host for enterprise server
  - Rules for parameter validation only if enterprise server is enabled
- Updated tests and rules.

### Features
- [github-enterprise-server connector in proxy](https://app.asana.com/0/1201039336231823/1206231703141594)

### Change implications

     obviously a no-op? **yes, about deprecated parameter**
     change


docs/sources/github/README.md
```diff
@@ -46,9 +47,9 @@ openssl pkcs8 -topk8 -inform PEM -outform PEM -in {YOUR DOWNLOADED CERTIFICATE F
 - If the certificate is not converted to PKCS#8 connector will NOT work. You might see in logs a Java error `Invalid PKCS8 data.` if the format is not correct.
 - Command proposed has been successfully tested on Ubuntu; it may differ for other operating systems.
 
-4. Install the application in your organization.
+5Install the application in your organization.
```
jlorper, line 50:
```suggestion
5. Install the application in your organization.
```

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -470,15 +469,122 @@ https://github.com/organizations/{YOUR ORG}/settings/installations/{INSTALLATION
...
+        }
+      ],
+      environment_variables : {
+        GRANT_TYPE : "certificate_credentials"
+        TOKEN_RESPONSE_TYPE : "GITHUB_ACCESS_TOKEN"
+        REFRESH_ENDPOINT : "https://${var.github_enterprise_server_host}/app/installations/${local.github_installation_id}/access_tokens"
+        USE_SHARED_TOKEN : "TRUE"
+      }
+      settings_to_provide = {
+        "GitHub Organization" = local.github_organization
```
jlorper, line 516: @eschultink do we need to support multiple orgs per connector? Make this a csv / array?

docs/sources/github/github.yaml
```diff
@@ -78,6 +116,46 @@ endpoints:
       username:
         type: "string"
         format: "reversible-pseudonym"
+  - pathTemplate: "/api/{enterpriseServerVersion}/users/{username}"
```
jlorper, line 119: why cloning with diff template? Just outlining options, not sure what's best:
- different file for enterprise cloud and server
- use regex "(/api/{enterpriseServerVersion)?/users/{username}"

jlorper (approved):

docs/sources/github/github.yaml
aperez-worklytics: Because I had in the task that sourceKind should be the same, so only way to add more rules was to include them in the same file. Tried with regex but that would complicate a bit the rules, so just split in two files -with different sourceKindId but same scopeId

infra/modules/worklytics-connector-specs/main.tf
eschultink: in theory, yeah ... but they could just put is a CSV, right? should work. Still a string.

infra/examples-dev/aws-all/misc-data-source-variables.tf
```diff
@@ -26,10 +26,10 @@ variable "jira_example_issue_id" {
   description = "If using Jira Server/Cloud connector, provide id of an issue for only to be used as part of example calls for Jira (ex: ETV-12)"
 }
 
-variable "github_api_host" {
+variable "github_enterprise_server_host" {
```
eschultink, line 29: this is a breaking change; so need to leave the current one `github_api_host` as deprecated, and add the new one. then take new one unless the default value; and if new one is the default value, take the old one.

eschultink (changes requested):

infra/examples-dev/aws-all/misc-data-source-variables.tf
aperez-worklytics: Ok; more or less. Despite the prev version where there is a default value (if `github_api_host` is not provided it will use the cloud url for github) here is nothing for that, so it will try to take old one, then new one and if not a placeholder for the host that could make easier to identity what's pending on settings

eschultink (approved):

PR #632, opened 2024-01-16
AWS  vpc support 

eschultink:
### Features
 - support putting Lambdas on VPC (beta)

### Change implications

 - dependencies added/changed? **yes** - use of AWS VPC depends on VPC 
     obviously a no-op?
     change

## TODO
 - [ ] ~~move policy down to each connector~~ - *save for beta*
 - [x] CMEK for apigateway log group

---
  - https://app.asana.com/0/0/1206296312465377


eschultink:
problem in lambda's `init` phase, although seems unrelated vpc/gateway changes ... need to debug

infra/examples-dev/aws-all/variables.tf
```diff
@@ -186,6 +186,11 @@ variable "custom_api_connector_rules" {
   default     = {}
 }
 
+variable "use_api_gateway" {
```
aperez-worklytics, line 189: Rename to `use_api_gateway_v2` to make it clear that is for v2 ?

infra/modules/aws-psoxy-rest/main.tf
```diff
@@ -63,9 +70,56 @@ resource "aws_lambda_function_url" "lambda_url" {
...
+  payload_format_version = "2.0"
+
+  # q: do we need to specify this?? what's the default?
+  # docs saw it can be 30000 max, but ideally we'd get ~55 or longer
+  # timeout_milliseconds   = 30000
+}
+
+resource "aws_apigatewayv2_route" "methods" {
+  # TODO: allow configuration of methods
+  for_each = toset(local.use_api_gateway ? ["GET", "HEAD"] : [])
```
aperez-worklytics, line 91: We will need `POST` to for GH and Dropbox

infra/modules/aws/main.tf
```diff
@@ -144,6 +148,67 @@ module "psoxy_package" {
...
+
+resource "aws_iam_policy" "invoke_api" {
+  count = var.use_api_gateway ? 1 : 0
+
+  name_prefix = "${var.deployment_id}InvokeAPI"
+
+  policy = jsonencode({
+    "Version" : "2012-10-17",
+    "Statement" : [
+      {
```
aperez-worklytics, line 180: Yep, missing post for policies too

aperez-worklytics (changes requested):

infra/examples-dev/aws-all/variables.tf
eschultink: 🤔  feels hacky to me to expose version numbers, but OK

aperez-worklytics: Agree, it was a suggestion, but on the other hand we had `use_api_gateway` and then in the description we enforce that is *enable api gateway v2* (as opposite to use v1), so yeah, agree that is weird to expose version numbers...

infra/modules/aws/main.tf
```diff
@@ -197,3 +267,14 @@ output "pseudonym_salt" {
   value       = random_password.pseudonym_salt.result
   sensitive   = true
 }
+
+output "api_gateway_v2" {
+  # NOTE: filled based on `var.use_api_gateway`, which is sufficient for Terraform to understand
```
aperez-worklytics, line 272:
```suggestion
  # NOTE: filled based on `var.use_api_gateway_v2`, which is sufficient for Terraform to understand
```

aperez-worklytics (approved):

PR #636, opened 2024-01-24
retry refresh token writes

jlorper:
### Fixes
Related to https://app.asana.com/0/1205088991476808/1206353682725974 
Perform few attempts upon write access and refresh token attempts, to minimize lost tokens if write failure happens.

### Change implications

     obviously a no-op? no
     change - no

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
@@ -1,5 +1,6 @@
 package co.worklytics.psoxy.gateway;
 
+import com.google.common.util.concurrent.Uninterruptibles;
 import lombok.AccessLevel;
 import lombok.Builder;
 import lombok.Getter;
@@ -7,6 +8,7 @@ import lombok.Value;
 import lombok.extern.java.Log;
 
 import java.io.Serializable;
+import java.time.Duration;
 import java.time.Instant;
 import java.util.Optional;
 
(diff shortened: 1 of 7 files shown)


java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
```diff
@@ -171,39 +177,48 @@ public boolean acquire(@NonNull String lockId, @NonNull Duration expires) {
...
 
                 if (lockedAt.isBefore(clock.instant().minusSeconds(expires.getSeconds()))) {
                     log.warning("Lock " + lockParameterName + " is stale, removing");
 
-                    //q: add random delay here, in case multiple instances have been waiting to
-                    // acquire the lock?
+                    // release it and let the caller decide if the retry
                     release(lockId);
-
-                    return acquire(lockId);
```
jlorper, line 195: removed this, so now will wait in caller WAIT_AFTER_FAILED_LOCK_ATTEMPTS + buffer

java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
```diff
@@ -50,6 +53,30 @@ default boolean supportsWriting() {
...
+    default void putConfigProperty(ConfigProperty property, String value, int retries) throws WritePropertyRetriesExhaustedException {
+        if (retries <= 0) {
+            // use the non-retry version
+            throw new IllegalArgumentException("retries must be > 0");
+        }
+        do {
+            try {
+                putConfigProperty(property, value);
+                return;
+            } catch (Exception ignore) {
```
eschultink, line 72: add logging?

eschultink (approved):

java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
jlorper: :+1: passed through the cause exception. Can't use `@Log` because is an interface

PR #637, opened 2024-01-26
KMS improvements

eschultink:
### Fixes
 - with explicit default encryption configuration, attempts to set configuration outside get overwritten (race)

### Change implications

 - something important to note in future release notes **yes; see CHANGELOG**

### DIFF
--- a/infra/modules/aws-psoxy-bulk/migration_legacy.tf
+++ b/infra/modules/aws-psoxy-bulk/migration_legacy.tf
@@ -3,10 +3,12 @@ moved {
   from = aws_iam_policy.output_bucket_write_policy
   to   = aws_iam_policy.sanitized_bucket_write_policy
 }
+
 moved {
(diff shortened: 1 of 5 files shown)


infra/modules/aws-psoxy-bulk/migration_legacy.tf
```diff
@@ -0,0 +1,27 @@
...
+  to   = aws_iam_policy.sanitized_bucket_read
+}
+
+moved {
+  from = aws_s3_bucket.output
+  to   = aws_s3_bucket.sanitized
+}
+
+moved {
+  from = aws_s3_bucket_server_side_encryption_configuration.output
```
aperez-worklytics, line 21: If `aws_s3_bucket_server_side_encryption_configuration` is destroyed, do we need to keep the movement from output to sanitized on the migration legacy ?

aperez-worklytics (approved):
Approved but just a question in the review comments

PR #643, opened 2024-02-05
AWS Secrets Manager support (alpha)

eschultink:
### Features
 - alpha support for AWS Secrets Manager as secrets store implementation
 - bash script to help Jira token exchange

### Change implications

 - dependencies added/changed? **yes, if used requires' AWS secret manager client lib, infra created; perms to use it**
 - something important to note in future release notes? **some moves**

---
  - https://app.asana.com/0/0/1206145456646562

### DIFF
--- a/docs/development/alpha-features/aws-secrets-manager.md
+++ b/docs/development/alpha-features/aws-secrets-manager.md
@@ -3,6 +3,15 @@
 With `v0.4.47`, we're adding **alpha** support for AWS Secrets Manager. This feature is not yet
 fully documented or stable.
 
+A couple notes:
+  - some connectors, in particular Zoom/Jira, rotate tokens frequently so will generate a lot of
+    versions of secrets.  AFIAK, AWS will still bill you for just one secret, as only one should be
+    staged as the 'current' version.  But you should monitor this and review the particular terms
+    and pricing model of your AWS contract.
+  - our modules will create secrets ONLY in the region into which your proxy infra is being deployed,
+    based on the value set in your `terraform.tfvars` file.
+
+
 Migration from Parameter Store: the default storage for secrets is as AWS Systems Manager Parameter
 Store SecureString parameters. If you have existing secrets in Parameter Store that *aren't* managed
 by terraform, you can copy them to a secure location to avoid needing to re-create them for every
--- a/docs/development/alpha-features/aws-secrets-manager.md
+++ b/docs/development/alpha-features/aws-secrets-manager.md
@@ -4,7 +4,11 @@ With `v0.4.47`, we're adding **alpha** support for AWS Secrets Manager. This fea
 fully documented or stable.
 
 Migration from Parameter Store: the default storage for secrets is as AWS Systems Manager Parameter
-Store SecureString parameters. If you have existing secrets in Parameter Store, s
+Store SecureString parameters. If you have existing secrets in Parameter Store that *aren't* managed
+by terraform, you can copy them to a secure location to avoid needing to re-create them for every
+source.
+
+
 
 ## Setup
 
@@ -23,7 +27,16 @@ secrets_store_implementation="aws_secrets_manager"
   5. Fill values of secrets in Secrets Manager that you copied from Parameter Store. If you did not
      copy the values, see the `TODO 1..` files for each connector to obtain new values.
 
+## Filling Secret Manager Secret values via AWS Console
 
+Navigate to the AWS Secrets Manager console and find the secret you need to fill. If there's not an
+option to fill the value, click 'Retrieve secret value'; it should then prompt you with option to
+fill it.
+
+IMPORTANT: Choose 'Plain Text' and remove the brackets (`{}`) that AWS prefills the input with!
+
+Then copy-paste the value EXACTLY as-is. Ensure no leading/trailing whitespace or newlines, and no
+encoding problems.
 
 ## NOTES
   - AWS Secret Manager secrets will be stored/accessed with same path as SSM parameters. Eg, at value of
--- a/docs/development/alpha-features/aws-secrets-manager.md
+++ b/docs/development/alpha-features/aws-secrets-manager.md
@@ -1,0 +1,35 @@
+# AWS Secrets Manager Support
+
+With `v0.4.47`, we're adding **alpha** support for AWS Secrets Manager. This feature is not yet
+fully documented or stable.
+
+Migration from Parameter Store: the default storage for secrets is as AWS Systems Manager Parameter
+Store SecureString parameters. If you have existing secrets in Parameter Store, s
+
+## Setup
+
+  1. If you forked the `psoxy-example-aws` repo prior to `v0.4.47`, you should copy a `main.tf` and
+`variables.tf` from that version of later of the repo and unify the version numbers with your own. (>=0.4.47)
+
+  2. Add the following to your `terraform.tfvars`:
+```hcl
+secrets_store_implementation="aws_secrets_manager"
+```
+  3. If you previously filled any secret values via AWS web console (such as API secrets you were
+     directed to create in `TODO 1` files for certain sources, you should copy those values now).
+
+  4. `terraform apply`; review plan and confirm when ready
+
+  5. Fill values of secrets in Secrets Manager that you copied from Parameter Store. If you did not
+     copy the values, see the `TODO 1..` files for each connector to obtain new values.
+
+
+
+## NOTES
+  - AWS Secret Manager secrets will be stored/accessed with same path as SSM parameters. Eg, at value of
+`aws_ssm_param_root_path`, if any.
+
+## Future work
+q: support distinct path for secrets? or generalize parameter naming?
+
+
--- a/docs/development/alpha-features/aws-secrets-manager.md
+++ b/docs/development/alpha-features/aws-secrets-manager.md
@@ -5,7 +5,7 @@ fully documented or stable.
 
 A couple notes:
   - some connectors, in particular Zoom/Jira, rotate tokens frequently so will generate a lot of
-    versions of secrets.  AFIAK, AWS will still bill you for just one secret, as only one should be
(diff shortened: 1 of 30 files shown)


docs/development/alpha-features/aws-secrets-manager.md
```diff
@@ -0,0 +1,57 @@
+# AWS Secrets Manager Support
+
+With `v0.4.47`, we're adding **alpha** support for AWS Secrets Manager. This feature is not yet
+fully documented or stable.
+
+A couple notes:
+  - some connectors, in particular Zoom/Jira, rotate tokens frequently so will generate a lot of
+    versions of secrets.  AFIAK, AWS will still bill you for just one secret, as only one should be
```
jlorper, line 8:
```suggestion
    versions of secrets.  AFAIK, AWS will still bill you for just one secret, as only one should be
```

docs/development/alpha-features/aws-secrets-manager.md
```diff
@@ -0,0 +1,57 @@
...
+## Filling Secret Manager Secret values via AWS Console
+
+Navigate to the AWS Secrets Manager console and find the secret you need to fill. If there's not an
+option to fill the value, click 'Retrieve secret value'; it should then prompt you with option to
+fill it.
+
+IMPORTANT: Choose 'Plain Text' and remove the brackets (`{}`) that AWS prefills the input with!
+
+Then copy-paste the value EXACTLY as-is. Ensure no leading/trailing whitespace or newlines, and no
+encoding problems.
```
jlorper, line 48: Idea: Maybe we could provide a simple command like (not tested and not sure if affected by the brackets thing you mention), but if the secret names are the same could be straightforward
```
value=$(aws ssm get-parameter --name "SECRET_NAME")
aws secretsmanager put-secret-value --secret-id SECRET_NAME --secret-string $value
```

infra/modules/aws-host/migration_0_4_47.tf
```diff
@@ -0,0 +1,11 @@
+# 0.4.47 will add support for Secrets Manager instead of SSM Parameter Store
+# customers will seem moves
```
jlorper, line 2:
```suggestion
# customers will see moves
```

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AwsModule.java
```diff
@@ -66,6 +67,7 @@ static SsmClient ssmClient(AwsEnvironment awsEnvironment) {
                 .build();
     }
 
+    //NOTE: also used in SecretsManager case for simplicity
     static String asParameterStoreNamespace(String functionName) {
```
jlorper, line 71: rename to a more general `asSecretNamespace` ?

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AwsModule.java
```diff
@@ -86,20 +88,54 @@ static CognitoIdentityClient cognitoClient(AwsEnvironment awsEnvironment) {
...
+            .map(String::toUpperCase) //case-insensitive, so accept 'aws_ssm_parameter_store'
+            .map(AwsEnvironment.SecretStoreImplementations::valueOf)
+            .orElse(AwsEnvironment.SecretStoreImplementations.AWS_SSM_PARAMETER_STORE);
+
+        ConfigService sharedConfigService;
+        ConfigService instanceConfigService;
+        if (secretStoreImpl == AwsEnvironment.SecretStoreImplementations.AWS_SECRETS_MANAGER) {
+            instanceConfigService = CompositeConfigService.builder()
+                .preferred(instanceScopedSecretsManagerConfigService)
+                .fallback(instanceScopedParameterConfigService)
```
jlorper, line 112: I don't think we should fallback to SSM. If the customer has specified the secrets manager we should stick to that. Will that even work from an infra perspective? If for whatever reason on infra side this would work could give the false sense that everything is under SM but is not.
Is this to support migrations from SSM to SM?

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
```diff
@@ -93,6 +93,7 @@ public void putConfigProperty(ConfigProperty property, String value) {
             log.info(String.format("Property: %s, stored version %d", key, parameterResponse.version()));
         } catch (SsmException e) {
             log.log(Level.SEVERE, "Could not store property " + key, e);
+            //q: should we be re-throwing here? caller will be unaware of failure
```
jlorper, line 96: yeah, it should! in `ConfigService` putConfigProperty (with retries version) expects an exception to retry!

jlorper (approved):
Some suggestions, no blockers. In SSM param config , please rethrow the exception

infra/examples-dev/aws-all/variables.tf
```diff
@@ -53,6 +53,12 @@ variable "aws_ssm_param_root_path" {
   }
 }
 
+variable "secrets_store_implementation" {
+  type        = string
+  description = "one of 'aws_ssm_parameter_store' (default) or 'aws_secrets_manager'"
+  default     = "aws_ssm_parameter_store"
```
aperez-worklytics, line 59: Adding validation with a constraint for avoiding random values here? (Same for other places where this new parameter is used)

```suggestion
  default     = "aws_ssm_parameter_store"
  
  validation {
    condition     = contains(["aws_ssm_parameter_store", "aws_secrets_manager"], var.secrets_store_implementation)
    error_message = "Allowed values for secrets_store_implementation are \"aws_ssm_parameter_store\" or \"aws_secrets_manager\"."
  }
```

infra/modules/aws-psoxy-lambda/main.tf
```diff
@@ -78,11 +71,12 @@ resource "aws_lambda_function" "instance" {
       var.path_to_config == null ? {} : yamldecode(file(var.path_to_config)),
       var.environment_variables,
       {
-        EXECUTION_ROLE  = aws_iam_role.iam_for_lambda.arn,
+        EXECUTION_ROLE  = aws_iam_role.iam_for_lambda.arn, # q: used for anything? doesn't seem accessed by Java code ...
```
aperez-worklytics, line 74: Yeah don't remember.. maybe we need it time ago for something?

infra/modules/aws-secretsmanager-secrets/main.tf
```diff
@@ -0,0 +1,43 @@
+# stores secrets as AWS Secret Manager SEcrets
+# NOTE: value of this module is a consistent interface across potential Secret store implementations
+#   eg, GCP Secret Manager, AWS SSM Parameter Store, Hashicorp Vault, AWS Secrets Manager, etc.
+#  but is this good Terraform style? clearly in AWS case, this module doesn't do much ...
```
aperez-worklytics, line 4: let's see how it grows..

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AwsModule.java
aperez-worklytics: Yeah, now is not only used for ParameterStore

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/SecretsManagerConfigService.java
```diff
@@ -0,0 +1,120 @@
...
+            PutSecretValueRequest request = PutSecretValueRequest.builder()
+                .secretId(id)
+                .secretString(value)
+                .build();
+            PutSecretValueResponse response = client.putSecretValue(request);
+
+            log.info(String.format("Property: %s, stored version %d", id, response.versionId()));
+        } catch (SecretsManagerException e) {
+            log.log(Level.SEVERE, "failed to write secret", e);
+            //q: re-throw??
```
aperez-worklytics, line 67: I think so, otherwise could cause inconsistencies

tools/jira-cloud-auth.sh
aperez-worklytics, line 1: :+1:

aperez-worklytics (approved):

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AwsModule.java
eschultink: our problem there is config values that AREN'T secrets.  Those are easier to inspect and change via AWS console than Secrets Manager secret.  

Although in the end for simplicity I left the best case of this (client id) as a secret just in case.

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/ParameterStoreConfigService.java
eschultink: yeah, it's not now

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AwsModule.java
eschultink: and actually, also problem for `RULES` and `ADDITIONAL_TRANSFORMS`; those don't fit as env vars.

what we need is to add the concept of whether a config value is 'secret' or not; and then pull secrets from the "secret store" and non-secret from regular config service (for which env vars falling through to SSM is more practical)

that said, env vars are legitimate for secrets as well, tbh.  just hard to manage as AWS ux renders as ~80 char html input.  YAML/JSON/etc that's multi-line is unworkable. And unlike SSM SecureString, values shown unmasked in the UX.

infra/modules/aws-secretsmanager-secrets/main.tf
```diff
@@ -0,0 +1,43 @@
+# stores secrets as AWS Secret Manager SEcrets
```
P-A-R-U-S, line 1:
```suggestion
# stores secrets as AWS Secret Manager Secrets
```

PR #649, opened 2024-02-09
GH: Fix double transform when pseudonymizing commits from timeline/single/pull commits

aperez-worklytics:
When reading timeline, single and pull commit, there was a rule for pseudonymizing emails and then the *user* rule was being applied, so the identities were pseudonymized twice.

### Fixes
[Fix Commit identities for GH](https://app.asana.com/0/1206513197716007/1206557848212702)

 - []()

### Change implications

     obviously a no-op?
     change

### DIFF
--- a/docs/sources/github/example-api-responses/original/issue_timeline.json
+++ b/docs/sources/github/example-api-responses/original/issue_timeline.json
@@ -5,8 +5,8 @@
     "url": "https://api.github.com/repos/khnum-heket/test2/git/commits/66249c83914e7ad8e9976eededda069b936ee426",
     "html_url": "https://github.com/khnum-heket/test2/commit/66249c83914e7ad8e9976eededda069b936ee426",
     "author": {
-      "name": "Octocat",
-      "email": "octocat@some-domain.com",
+      "name": "AuthorUser",
+      "email": "authorUser@some-domain.com",
       "date": "2023-09-18T10:46:54Z"
     },
     "committer": {
--- a/docs/sources/github/example-api-responses/original/issue_timeline.json
+++ b/docs/sources/github/example-api-responses/original/issue_timeline.json
@@ -1,5 +1,40 @@
 [
   {
+    "sha": "66249c83914e7ad8e9976eededda069b936ee426",
+    "node_id": "C_kwDOKU_SqtoAKDY2MjQ5YzgzOTE0ZTdhZDhlOTk3NmVlZGVkZGEwNjliOTM2ZWU0MjY",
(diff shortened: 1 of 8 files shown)


docs/sources/github/example-api-responses/sanitized/commit.json
```diff
@@ -7,11 +7,11 @@
   "commit":{
     "url":"https://api.github.com/repos/org/Hello-World/git/commits/6dcb09b5b57875f334f61aebed695e2e4193db5e",
     "author":{
-      "email":"{\"scope\":\"github\",\"hash\":\"T-sBXXi6k0bWw-zNtcdQenxAuqN5C51_ehZ5lJDWeqw\"}",
+      "email":"{\"scope\":\"email\",\"domain\":\"github.com\",\"hash\":\"39uRnKxs0PYIueRoGGc7fyyp-hSTtamEhvB6YywFNVM\"}",
```
aperez-worklytics, line 10: This is what it was wrong, the double pseudonymization generates an invalid identity, considered it as *github* when it is actually an email from someone

docs/sources/github/example-api-responses/original/issue_timeline.json
```diff
@@ -1,4 +1,39 @@
 [
+  {
+    "sha": "66249c83914e7ad8e9976eededda069b936ee426",
+    "node_id": "C_kwDOKU_SqtoAKDY2MjQ5YzgzOTE0ZTdhZDhlOTk3NmVlZGVkZGEwNjliOTM2ZWU0MjY",
+    "url": "https://api.github.com/repos/khnum-heket/test2/git/commits/66249c83914e7ad8e9976eededda069b936ee426",
+    "html_url": "https://github.com/khnum-heket/test2/commit/66249c83914e7ad8e9976eededda069b936ee426",
+    "author": {
+      "name": "AuthorUser",
+      "email": "authorUser@some-domain.com",
```
eschultink, line 9: this is just whatever user sets in `.gitconfig` right? not validated or anything, right?

we should duck-type it, and if is formated like an email address, we'll scope it as such, right?

eschultink (approved):

docs/sources/github/example-api-responses/original/issue_timeline.json
aperez-worklytics: Yes, that was due the invalid config rules proxy wasn't doing it. Now with the changes it is what is doing, but I can include a test to cover the case if you want (if work on that way before)

jlorper (approved):

PR #652, opened 2024-02-13
prettier formatting markdown

eschultink:
### Features
 - use `prettier` to format markdown; apply it

### Change implications


aperez-worklytics (approved):
Cool; just as a suggestion, maybe we could refer `style.sh` in some doc (sorry if exist and I've missed that), maybe in the README or in other main document just to know we have that tool and it is a good practice to use it when writing documents

eschultink:
> Cool; just as a suggestion, maybe we could refer `style.sh` in some doc (sorry if exist and I've missed that), maybe in the README or in other main document just to know we have that tool and it is a good practice to use it when writing documents

yeah, i haven't yet ... probably will integrate it into release process.

PR #656, opened 2024-02-21
GCP: Using artifact registry

aperez-worklytics:
In May 2024, new GCP Cloud Functions will use *Artifact Registry* instead of *Container Registry*. However I see that for creating a new function today, artifact registry needs to be enabled for that as the deployment will use that way.

This PR just enables the API for artifact registry without anything else. Old functions will work as before, using container registry and no action needs to be done. But any new deployment it will put everything in the artifact repository.

### Features
[Psoxy: Enable artifact registry for GCP](https://app.asana.com/0/1206626185959023/1206654468494641)

### Change implications

     obviously a no-op?
   - breaking changes? **no**


infra/modules/gcp-psoxy-rest/main.tf
```diff
@@ -70,6 +70,8 @@ resource "google_cloudfunctions_function" "function" {
   entry_point                  = "co.worklytics.psoxy.Route"
   service_account_email        = var.service_account_email
   labels                       = var.default_labels
+  docker_registry              = "ARTIFACT_REGISTRY"
```
aperez-worklytics, line 73: Now, it seems that functions are deployed by default in artifact registry if they are new, so specifing here that for all of them.

infra/modules/gcp/main.tf
```diff
@@ -26,6 +27,36 @@ resource "google_project_service" "gcp_infra_api" {
   disable_on_destroy         = false # disabling on destroy has potential to conflict with other uses of the project
 }
 
+resource "google_artifact_registry_repository" "psoxy-functions-repo" {
```
aperez-worklytics, line 30: Actually this is *not* required, as if we don't specify the repo GCP will create a new one:

image

But I think it is cleaner if we create one for psoxy images, so we could customize some data there...

infra/modules/gcp/main.tf
```diff
@@ -26,6 +27,36 @@ resource "google_project_service" "gcp_infra_api" {
...
+  ## Not supported in current google providers, needs 5.14 as there it is GA
+  # See https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#5140-jan-29-2024
+  # but even is present in the documentation (https://registry.terraform.io/providers/hashicorp/google/4.80.0/docs/resources/artifact_registry_repository#argument-reference)
+  # when applied it throws an error with the message: "An argument named "cleanup_policy_dry_run" is not expected here"
+  # and "no block for cleanup_policies" is expected
+  /*cleanup_policy_dry_run = false
+
+  # https://cloud.google.com/artifact-registry/docs/repositories/cleanup-policy#json_2
+  # https://registry.terraform.io/providers/hashicorp/google/4.80.0/docs/resources/artifact_registry_repository#argument-reference
+  cleanup_policies {
```
aperez-worklytics, line 46: ... so mainly for this. Adding a `cleanup_policy` we could just put rules for keeping the most recent images instead of storing all of them.

But it doesn't work with 4.84.0, 5.0.0 and it will only work since 5.14.0, which I have to modify the versioning of the gcp providers (from 3.74 to 5.14). Not sure if we can do it, but if so, it could be interesting. In any case, I've left the commented code here, at some point we could use it once we update the versions

infra/modules/gcp-psoxy-rest/variables.tf
```diff
@@ -56,6 +56,11 @@ variable "deployment_bundle_object_name" {
   description = "Name of the object containing the deployment bundle"
 }
 
+variable "artifact_repository_id" {
+  type        = string
+  description = "ID of the artifact repository"
+}
```
eschultink, line 62: adding this w/o a default is a breaking change. older examples/etc that reference module directly (although shouldn't be common) will break unless we specify a default

infra/modules/gcp/main.tf
eschultink: can we make it conditional or something?

Otherwise it's a breaking change, as technically we have the version for google provider in examples at       `version = ">= 3.74, <= 5.0"` -- and this is still in beta < 4.0

eschultink (changes requested):
seems OK, but we should EITHER avoid the breaking changes OR hold for 0.5 of proxy, which could bump google provider to >= 5.0 too. 

That might be worth doing, but if this is really blocking deployment of new GCP cloud functions, we should have some transitional solution - so GCP customers can deploy additional sources w/o full upgrade to 0.5

aperez-worklytics:
> seems OK, but we should EITHER avoid the breaking changes OR hold for 0.5 of proxy, which could bump google provider to >= 5.0 too.
> 
> That might be worth doing, but if this is really blocking deployment of new GCP cloud functions, we should have some transitional solution - so GCP customers can deploy additional sources w/o full upgrade to 0.5

Ok, tried to have that without artifact repository to leave the default as GCP will create that for us -as same as is doing for deploying the functions in their container registry-. Name of that default, according to my tests, is `gcf-artifacts`

So, what we can do is:
1) Comment (or drop) the code for creating an artifact repository. With that, new deployments will happen through `gcf-artifacts` artifact repository.
2) Once we are in 0.5, include the part of creating of our repository with the deletion rules. By default the repo name used will be `gcf-artifacts` to keep the compatibility.

Issues when doing 1) is that we need to warn to remove old images in artifact repository, as that cannot be done with terraform. Then, the other issue is that we'll need to use the default artifact repository, so in case since 0.5 can be customized some work needs to be done to drop old images too.

@eschultink thoughts ?

infra/modules/gcp-host/main.tf
```diff
@@ -175,7 +177,8 @@ module "api_connector" {
 
   secret_bindings = merge(
     local.secrets_bound_as_env_vars[each.key],
-    module.psoxy.secrets
+    module.psoxy.secrets,
+    module.psoxy.artifact_repository
```
aperez-worklytics, line 181: Need some dependency here. `artifactrepository` API needs to be activated before running this, so:
1) the output variable has a dependency of that
2) we *force* the dependency here in the module too.

Otherwise following error can happen:

`Error while updating cloudfunction configuration: googleapi: Error 403: Cloud Functions uses Artifact Registry to store function docker images. Artifact Registry API is not enabled in your project. To enable the API, visit https://console.developers.google.com/apis/api/artifactregistry.googleapis.com/overview?project=[REDACTED] or use the gcloud command 'gcloud services enable artifactregistry.googleapis.com' then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry., forbidden
`

And it will require two deployments for doing the same: one for activating the API and then, a second one for running it without errors (to avoid the delay between the terraform service and the real activation in the API)

infra/modules/gcp/main.tf
aperez-worklytics: Leaving this commented with a TODO

infra/modules/gcp-psoxy-bulk/variables.tf
```diff
@@ -52,6 +52,11 @@ variable "artifacts_bucket_name" {
   description = "Name of the bucket where artifacts are stored"
 }
 
+variable "artifact_repository_id" {
+  type        = string
+  description = "ID of the artifact repository"
+}
```
eschultink, line 58: provide `default = null` so anyone using this module directly won't break? (not aware of anyone using it directly in gcp case, but a couple are using aws directly)

infra/modules/gcp-psoxy-rest/variables.tf
eschultink: yeah - as above, it's a breaking change - can we put default as `null` or empty and deal with it?

infra/modules/gcp/main.tf
```diff
@@ -315,3 +347,17 @@ output "pseudonym_salt" {
...
+  #value = google_artifact_registry_repository.psoxy-functions-repo.id
+  value = null # by default, GCP Artifact Registry will use "gcf-artifacts" repository
+
+  depends_on = [
+    # For ensuring the API is enabled, otherwise following error can happen:
+    # "Error while updating cloudfunction [...] Cloud Functions uses Artifact Registry to store function docker images.
+    # Artifact Registry API is not enabled in your project. To enable the API, visit [...]
+    # or use the gcloud command 'gcloud services enable artifactregistry.googleapis.com' then retry.
+    # If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry., forbidden"
+    google_project_service.gcp_infra_api
```
eschultink, line 361: yeah, ok - in practice I think will be fine; we can also flag as prereq/bootstrap case, and everyone should be ok; worst case a couple applies will get it

eschultink (changes requested):

infra/modules/gcp-psoxy-rest/variables.tf
aperez-worklytics: Yeah sorry, miss that before.

eschultink (approved):

PR #657, opened 2024-02-26
destroy rather than disable GCP secret versions when updating

eschultink:
### Fixes
 - disabled secrets still have costs; so we need to delete, rather than disable

SCR-20240226-lvio

### Change implications

 - something important to note in future release notes? **no, albeit some behavior change; not impactful**

### DIFF
--- a/java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
+++ b/java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
@@ -21,6 +21,8 @@ import java.io.IOException;
 import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -209,30 +209,61 @@ private static void updateLabelFromSecret(SecretManagerServiceClient client, Sec
...
+        // in Notes section of their pricing paging,
+        // "1. A secret version is 'active' when it's in any of these states: ENABLED, DISABLED"
+       // so we need to DESTROY any versions of these secrets that are merely DISABLED
+        try  {
+            Instant until = Instant.ofEpochSecond(exceptVersion.getCreateTime().getSeconds())
+                .minus(Duration.ofDays(7));
+
+            SecretManagerServiceClient.ListSecretVersionsPage disabledVersions =
+                client.listSecretVersions(ListSecretVersionsRequest.newBuilder()
+                    .setFilter("state:DISABLED AND create_time < " + until.getEpochSecond() + "s")
```
jlorper, line 251: first time I see epochSecond as filter, but OK. Guess @aperez-worklytics  can test this on his proxy, I don't have a GCP one either

jlorper (approved):

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
aperez-worklytics: Added this commit https://github.com/Worklytics/psoxy/pull/657/commits/ce769c9a3db1aaacb6b4cbe70b072f3f6772e46c @eschultink after trying it which fixes the filtering. Working on my env after that.

aperez-worklytics (approved):

PR #660, opened 2024-02-27
Bulk sanitizer fix transformation order

jlorper:
### Fixes
[Lookup table generation failing since 0.4.47]
(https://app.asana.com/0/1206626185959023/1206708363382894)

### Change implications

     obviously a no-op?
     change

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
@@ -29,7 +29,9 @@ import org.apache.commons.csv.CSVPrinter;
 import org.apache.commons.csv.CSVRecord;
 import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.function.TriConsumer;
 import org.apache.commons.lang3.function.TriFunction;
+import org.apache.commons.lang3.tuple.Triple;
 
 import javax.inject.Inject;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -94,8 +96,14 @@ public void sanitize(@NonNull Reader reader,
          * K = new column name
          * C = function to apply to the original column
          * V = original column name
+         * Note, can't use Guava Table as order of the transformations to apply is not deterministic.
+         * Using LinkedHashMap will use insertion order
          */
-        Table<String, Function<String, Optional<String>>, String> transformTable = TreeBasedTable.create(String.CASE_INSENSITIVE_ORDER, Comparator.comparingInt(Object::hashCode));
+        Map<String, Map<Function<String, Optional<String>>, String>> columnTransforms = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
eschultink, line 102: clear as `Map<String, List<Pair<String, FUnction<??>>>`

eschultink (changes requested):

java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
eschultink: or `Map<String, Pair<String, List<Function<String, Optional<String>>>>`

eschultink: eg, `nameInOutput` --> (`source`, `transformsToApply...`)

eschultink (approved):

PR #678, opened 2024-03-12
generalize bulk record handling

eschultink:
### Features
  - support 'CSV' as a format for record-type rules; just matches expectation and provides for a more readable pipeline approach
  - improve maintainability of tests by preserving line endings; don't know why didn't realize this was configurable, but easy to using apache CSV writer to preserve the same line ending, rather than convert everything to `\r\n`.

### Change implications


java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -293,6 +293,7 @@ public void sanitize(@NonNull Reader reader,
 
         CSVFormat csvFormat = CSVFormat.Builder.create()
             .setHeader(columnNamesForOutputFile.toArray(new String[0]))
+            .setRecordSeparator(records.getFirstEndOfLine())
```
eschultink, line 296: technically, a breaking change; previously separated records with `\r\n` regardless of what was in the input file.

jlorper (approved):

PR #685, opened 2024-03-20
misc fixes to beta features (AWS Secrets Manager; API Gateway)

eschultink:
### Fixes
 - [fix AWS Secrets Manager default prefix via aws-host case](https://github.com/Worklytics/psoxy/commit/3b57aa71f71dbf3f9e473a74b9a14e394fa90e54)
 - [fix healthcheck example in todo_2 output in api_gateway case](https://github.com/Worklytics/psoxy/commit/399d9861bc9ad938517150cf425203a4b89af534)

### Change implications

 - something important to note in future release notes? **no; behavior change in beta features only (AWS Secrets Manager, API gateway)**

---
  - https://app.asana.com/0/0/1206469318133980

### DIFF
--- a/infra/modules/aws-host/variables.tf
+++ b/infra/modules/aws-host/variables.tf
@@ -49,6 +49,12 @@ variable "logs_kms_key_arn" {
   default     = null
 }
 
+variable "log_retention_days" {
+  type        = number
+  description = "Number of days to retain logs in CloudWatch."
+  default     = 7
+}
+
 variable "aws_lambda_execution_role_policy_arn" {
   type        = string
   description = "*beta* The ARN of policy to attach to the lambda execution role, if you want one other than the default. (usually, AWSLambdaBasicExecutionRole)."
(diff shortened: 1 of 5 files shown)


P-A-R-U-S (approved):

infra/modules/aws-host/variables.tf
```diff
@@ -49,6 +49,12 @@ variable "logs_kms_key_arn" {
   default     = null
 }
 
+variable "log_retention_days" {
```
aperez-worklytics, line 52: Ok, variable is passed but I don't see where it is set

eschultink: SCR-20240321-oyyq

aperez-worklytics: ah yeah sorry :facepalm: ; I was expecting to see the direct application on some AWS resource, but that was already there through `aws-psoxy-rest` and `aws-psoxy-bulk` so this was only providing the parameter instead of using the default

PR #688, opened 2024-03-22
add bulk file compression test

eschultink:
### Features
 - add test case for bulk of compressed as well as uncompressed example

### Change implications

### DIFF
--- a/tools/psoxy-test/lib/gcp.js
+++ b/tools/psoxy-test/lib/gcp.js
@@ -247,7 +247,7 @@ async function download(bucketName, fileName, destination, client, logger) {
   }
 
   const downloadFunction = async () => client.bucket(bucketName).file(fileName)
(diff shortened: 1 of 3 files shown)


tools/psoxy-test/lib/gcp.js
```diff
@@ -247,7 +247,7 @@ async function download(bucketName, fileName, destination, client, logger) {
   }
 
   const downloadFunction = async () => client.bucket(bucketName).file(fileName)
-    .download({ destination: destination });
+    .download({ destination: destination, decompress: false });
```
eschultink, line 250: as-is, was failing due to auto-magic decompress by GCS when we properly set `Content-Encoding`, which we do now (didn't before)

aperez-worklytics (approved):

PR #693, opened 2024-03-26
relax tf version contraints

eschultink:
### Features
  - add tests against tf 1.7.x, and relax constraints
  - add docs on troubleshooting terraform version issues
  - document Terraform compatibility targets for both developers and users

### Change implications

 - dependencies added/changed? **yes - adding support for 1.7.x and relaxing terraform constraints**

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ You will need all the following in your deployment environment (eg, your laptop)
 | [git](https://git-scm.com/)                     | 2.17+                  | `git --version`           |
 | [Maven](https://maven.apache.org/)              | 3.6+                   | `mvn -v`                 |
 | [Java JDK 11+](https://openjdk.org/install/) | 11, 17, 21 (see notes) | `mvn -v &#124; grep Java` |
-| [Terraform](https://www.terraform.io/)          | 1.3.x, <= 1.6          | `terraform version`       |
+| [Terraform](https://www.terraform.io/)          | 1.3.x, <= 1.7          | `terraform version`       |
 
 NOTE: we will support Java versions for duration of official support windows, in particular the
 LTS versions. As of Nov 2023, we  still support java 11 but may end this at any time. Minor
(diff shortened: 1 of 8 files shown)


docs/development/terraform-versions.md
```diff
@@ -0,0 +1,36 @@
+# Terraform Version Compatibility
+
+We use CI to automate testing against *latest* version of each minor version of Terraform
+(eg 1.3.x, 1.4.x, etc that we support).
+
+Given that Hashicorp has changed Terraform licensing to ELv2; and that Linux foundation has forked
+Terraform from 1.5.x to [create OpenTofu](https://opentofu.org/blog/the-opentofu-fork-is-now-available/),
+we are targeting compatibility with 1.6.x feature set.
+
+Do NOT use any features from 1.7.x or later, as we wish to allow all our terraform modules/examples
```
aperez-worklytics, line 10: Not clear this point. features listed below are introduced in >= 1.4. So not sure if want to say something like 

```suggestion
Do NOT use any features from 1.3.x or later, as we wish to allow all our terraform modules/examples
```

or just clarify why something introduced before 1.7.x (like `strcontains`) cannot be used

docs/development/terraform-versions.md
```diff
@@ -0,0 +1,36 @@
...
+to work with Terraform back to 1.3.x, which is vintage Sept 2022.
+
+In particular, **features to NOT use**:
```
aperez-worklytics, line 13: Thinking about how we can prevent some of this features to be used in future reviews... any idea?

README.md
```diff
@@ -218,7 +218,7 @@ You will need all the following in your deployment environment (eg, your laptop)
 | [git](https://git-scm.com/)                     | 2.17+                  | `git --version`           |
 | [Maven](https://maven.apache.org/)              | 3.6+                   | `mvn -v`                 |
 | [Java JDK 11+](https://openjdk.org/install/) | 11, 17, 21 (see notes) | `mvn -v &#124; grep Java` |
-| [Terraform](https://www.terraform.io/)          | 1.3.x, <= 1.6          | `terraform version`       |
+| [Terraform](https://www.terraform.io/)          | 1.3.x, <= 1.7          | `terraform version`       |
```
aperez-worklytics, line 221: From YAML https://github.com/Worklytics/psoxy/pull/693/files#diff-224777d94940d8246633bd46b1a26f8cf36bc33a4ee0046fb9891c66d55f1fc0L24 version is `~1.7.0`. 

Maybe .. ?

```suggestion
| [Terraform](https://www.terraform.io/)          | 1.3.x, <= 1.7.x          | `terraform version`       |
```

aperez-worklytics (approved):
Approved but with a couple of questions

docs/development/terraform-versions.md
eschultink: in theory, the CI runs validate with 1.3.x that should catch it (would be language syntax error i believe)

PR #698, opened 2024-04-16
fix conditional file

eschultink:
### Fixes
 - tf apply fails with `todos_as_local_files == false` without this

### Change implications

### DIFF
--- a/infra/modules/azuread-grant-all-users/main.tf
+++ b/infra/modules/azuread-grant-all-users/main.tf
@@ -44,7 +44,7 @@ output "todo" {
 }
 
 output "filename" {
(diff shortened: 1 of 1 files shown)


infra/modules/azuread-grant-all-users/main.tf
```diff
@@ -44,9 +44,9 @@ output "todo" {
 }
 
 output "filename" {
-  value = local_file.todo[0].filename
+  value = var.todos_as_local_files ? local_file.todo[0].filename : null
```
aperez-worklytics, line 47: :facepalm:

PR #700, opened 2024-04-17
auth config fallback to config, if not in secrets

eschultink:
### Features
 - various auth properties (client id, account id) aren't in and of themselves *secret*; so allow to get them from Config if not found in `SecretStore`; we have customers who think of them as secret, but others who prob don't want to pay overhead of handling them as secrets

### Change implications

 - something important to note in future release notes? **should be backwards compatible**

---
  - https://app.asana.com/0/0/1207091086964724

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestBuilder.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestBuilder.java
@@ -180,20 +180,23 @@ public class ClientCredentialsGrantTokenRequestBuilder
         Map<String, String> data = new HashMap<>();
 
         data.put("grant_type", getGrantType());
-        data.put("client_id",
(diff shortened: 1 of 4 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestBuilder.java
```diff
@@ -180,20 +180,23 @@ private Map<String, String> buildClientSecretPayload() {
...
-        data.put("client_id",
-            secretStore.getConfigPropertyOrError(OAuthRefreshTokenSourceAuthStrategy.ConfigProperty.CLIENT_ID));
+        data.put("client_id", getClientId());
         data.put("client_secret",
             secretStore.getConfigPropertyOrError(ConfigProperty.CLIENT_SECRET));
 
         return data;
     }
 
+    private String getClientId() {
```
aperez-worklytics, line 190: I'd move this on the bottom of the file

aperez-worklytics (approved):

PR #705, opened 2024-05-02
Jira cloud drop cloudid from deep linking

aperez-worklytics:
~~**NOTE** Mark as draft, as until https://github.com/Worklytics/evalengin/pull/6537 is released this cannot be integrated in psoxy.~~

- Drop `jira_cloud_id` parameter from deep linking
- Minor update on instructions

 - []()

### Features
[Update Jira Proxy doc and remove parameter from deep linking](https://app.asana.com/0/0/1206475184639833)

### Change implications

     obviously a no-op? **no**
     change **no**

### DIFF
--- a/infra/modules/worklytics-connector-specs/main.tf
+++ b/infra/modules/worklytics-connector-specs/main.tf
@@ -166,7 +166,7 @@ locals {
   k => merge(v, { example_calls : v.example_api_calls }) }
 
 
-  jira_cloud_id                    = coalesce(var.jira_cloud_id, "YOUR_JIRA_CLOUD_ID")
+  jira_example_cloud_id                    = coalesce(var.jira_cloud_id, "YOUR_JIRA_CLOUD_ID")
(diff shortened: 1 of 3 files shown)


infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -166,7 +166,7 @@ locals {
   k => merge(v, { example_calls : v.example_api_calls }) }
 
 
-  jira_cloud_id                    = coalesce(var.jira_cloud_id, "YOUR_JIRA_CLOUD_ID")
+  jira_example_cloud_id                    = coalesce(var.jira_cloud_id, "YOUR_JIRA_CLOUD_ID")
```
aperez-worklytics, line 169: Minor refactor here for naming, as now this is only for the example calls. But if I change this across all the modules is going to be a breaking change. Another option is to duplicate the variable and choose the not null; not sure what approach you want to follow for this @eschultink

eschultink (approved):

eschultink:
👍 sry i missed this; for some reason i thought it was dependent on something else

PR #707, opened 2024-05-10
improve init

eschultink:
### Features
  - customers keep leaving `google-workspace` references in their configs, even though not using it as a source; try to improve prompts and provide troubleshooting guidance for this case

### Change implications

     obviously a no-op?
     change

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -4,12 +4,41 @@
 
 - [Microsoft 365](docs/sources/msft-365/troubleshooting.md)
 
+
+### Error: Attempted to load application default credentials (Google provider authentication failure)
+
+This is related to `gcloud` not being authenticated (or installed?) in the environment where you're
+running terraform, which the `google` terraform provider requires.
+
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -4,13 +4,42 @@
...
 
 - [AWS](docs/aws/troubleshooting.md)
 - [GCP](docs/gcp/troubleshooting.md)
 
 ## General Tips
 
+### Verify Pre-Requisites
+
+Our example templates include a script to check for the prerequisites for running the psoxy. You
+can run this prior to `./init` to get feedback/suggestions on what prerequisites you may be missing
```
jlorper, line 36:
```suggestion
can run this prior to `./init-tfvars.sh` to get feedback/suggestions on what prerequisites you may be missing
```
You meant that??

Also, we could include in the `init-tfvars.sh` a Y/N question: **Check prerequisites installed?** and run the other script.

jlorper (approved):
One suggestion

aperez-worklytics (approved):

PR #728, opened 2024-06-25
avoid gcp key creation

eschultink:
### Features
 - to deal w customer policies that restrict key creation in GCP

### Change implications

---
  - https://app.asana.com/0/0/1207661870677980

### DIFF
--- a/infra/examples-dev/gcp/google-workspace-variables.tf
+++ b/infra/examples-dev/gcp/google-workspace-variables.tf
@@ -26,6 +26,12 @@ variable "google_workspace_example_admin" {
   default     = null # will failover to user
 }
 
+variable "google_workspace_provision_keys" {
+  type        = bool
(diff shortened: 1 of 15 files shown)


jlorper (approved):

infra/examples-dev/gcp/google-workspace-variables.tf
```diff
@@ -26,6 +26,12 @@ variable "google_workspace_example_admin" {
   default     = null # will failover to user
 }
 
+variable "google_workspace_provision_keys" {
```
aperez-worklytics, line 29: Maybe the name could be `provision_google_workspace_keys` to clarify this is the flag for creating the keys, not the *keys* itself to use.

aperez-worklytics (approved):
Approved; but I'd change the name of the variable

infra/examples-dev/gcp/google-workspace-variables.tf
eschultink: yeah ... i was just trying group all the variables with common prefix, to make them easy to find by source at this level.

PR #733, opened 2024-06-28
improve msft rules

eschultink:
### Features
 - as discussed, exclude `jobTitle` by default

### Change implications

 - something important to note in future release notes? **could note, but as makes default rules more conservative (not more permissive) - not really breaking**


aperez-worklytics (approved):
AS I have to work with [refine proxy rules wrt to `beta` v `1.0` msft endpoints](https://app.asana.com/0/1200304402089071/1207472295155600)[refine proxy rules wrt to `beta` v `1.0` msft endpoints](https://app.asana.com/0/1200304402089071/1207472295155600) (hopefully monday next week) if there is any other field we need to remove, just include it in the task and I'll do it.

PR #748, opened 2024-07-10
gcp functions-framework-api as provided dep

eschultink:
**this does NOT work, although unclear it's for reasons related to this change**

 - []()

 - []()

### Change implications

     obviously a no-op?
     change


eschultink:
superceded by #751

PR #738, opened 2024-07-02
Support online meetings for Outlook-Cal to work with Teams

aperez-worklytics:
`outlook-cal` needs support for fetching online meetings. Permissions required for that were included in https://github.com/Worklytics/psoxy/pull/622 but it was missing there the extra configuration required on Teams side and the rules for online meetings in outlook-cal.

### Features
[Include teams config in outlook-cal setup for proxy](https://app.asana.com/0/1207553650922163/1207676092252177)

### Change implications

 - dependencies added/changed? **no**
 - something important to note in future release notes?
   - NOTE in `CHANGELOG.md` anything that will show up in `terraform plan`/`apply` that isn't
     obviously a no-op?
   - breaking changes? if in module/example that is NOT marked `alpha`, requires major version
     change

PR #735, opened 2024-07-01
MSFT rules update for using /v1.0

aperez-worklytics:
- Only allowing v1.0 endpoints in MSFT endpoints (entra-id, outlook-cal, outlook-mail, onedrive, teams)
- Dropping beta support for those endpoints (NOTE: https://github.com/Worklytics/evalengin/pull/6711 related)
- Adding an example of calendar view with online meetings
- Updated docs and specs to remove any beta reference

### Features
[Psoxy rules](https://app.asana.com/0/0/1207702373764545)

### Change implications

 - dependencies added/changed? **no**
 - something important to note in future release notes?
   - NOTE in `CHANGELOG.md` anything that will show up in `terraform plan`/`apply` that isn't
     obviously a no-op?
   - breaking changes? if in module/example that is NOT marked `alpha`, requires major version
     change

PR #756, opened 2024-07-11
Teams cleanup

aperez-worklytics:
Following #735 and part of #738:
- Path in Teams was supporting any api version, not only 1.0
- Dropping online meetings permission from outlook-cal (as per https://github.com/Worklytics/evalengin/pull/6719 is not required that permission there)
- Some cleanup: moved salesforce tests under `rules` folder (it was outside, not sure why).

### Features
[MSFT Graph v1.0 for Teams](https://app.asana.com/0/1207688469675086/1207785655059390)

### Change implications

     obviously a no-op?
     change


infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -260,8 +260,6 @@ locals {
       target_host : "graph.microsoft.com"
       required_oauth2_permission_scopes : []
       required_app_roles : [
-        "OnlineMeetings.Read.All",
-        "OnlineMeetingArtifact.Read.All",
```
eschultink, line 264: shouldn't this also be reflected in docs updates? docs/README, and the outlook-cal specific docs page w scopes?

eschultink (changes requested):

eschultink (approved):

PR #757, opened 2024-07-11
java17 runtimes by default

eschultink:
### Features
 - [java17 runtimes by default](https://app.asana.com/0/1201039336231823/1207783492196045)

NOTE: this does NOT change source code or byte code java versions - both still target to java 11.  So for clarity, this means we write java 11 code, which will be compiled to java 11 bytecode (using any JDK >= 11), which will be run by a java 17 JRE.

main driver of this is GCP, which is deprecating java11 runtime in Sept 2024; and generally doesn't maintain the java11 containers as robustly as the java17 ones.

### Change implications

 - something important to note in future release notes? **terraform apply will show `java11` --> `java17` runtime change**

### DIFF
--- a/infra/modules/aws-psoxy-lambda/main.tf
+++ b/infra/modules/aws-psoxy-lambda/main.tf
@@ -44,7 +44,7 @@ resource "aws_lambda_function" "instance" {
   function_name                  = local.function_name
   role                           = aws_iam_role.iam_for_lambda.arn
   architectures                  = ["arm64"] # 20% cheaper per ms exec time than x86_64
-  runtime                        = "java11"
+  runtime                        = "java17"
   filename                       = local.bundle_from_s3 ? null : var.path_to_function_zip
   s3_bucket                      = local.s3_bucket # null if local file
   s3_key                         = local.s3_key    # null if local file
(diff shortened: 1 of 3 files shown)


aperez-worklytics (approved):

infra/modules/aws-psoxy-lambda/main.tf
```diff
@@ -44,7 +44,7 @@ resource "aws_lambda_function" "instance" {
   function_name                  = local.function_name
   role                           = aws_iam_role.iam_for_lambda.arn
   architectures                  = ["arm64"] # 20% cheaper per ms exec time than x86_64
-  runtime                        = "java11"
+  runtime                        = "java17"
```
aperez-worklytics, line 47: BTW; worth to parametrize this ? By default `java17`, but it can be customized with any other value depending on the needs

aperez-worklytics (approved):

infra/modules/aws-psoxy-lambda/main.tf
eschultink: YAGNI ... I don't see utility of it. java11 is EoL, deprecated by Oracle since Sept 2023; will be deprecated by GCP as a runtime in Sept 2024.  

if someone asks for running in java11 jre, then we can do it  - but I just don't what the use-case would be. If anything, I think they'd want opposite (forcing to java17 or java21, as those are more fully supported)

PR #763, opened 2024-07-11
drop basic exec policy from caller role 

eschultink:
### Fixes
 - lambda basic exec role is effectively just logging perms, not needed - as lambdas have their own exec rules with those perms

### Change implications

 - something important to note in future release notes? **yes - will see role attachment being destroyed**

---
  - https://app.asana.com/0/0/1206355928380269

### DIFF
--- a/infra/modules/aws/main.tf
+++ b/infra/modules/aws/main.tf
@@ -103,11 +103,6 @@ resource "aws_iam_policy" "execution_lambda_to_caller" {
   }
 }
 
-resource "aws_iam_role_policy_attachment" "invoker_lambda_execution" {
-  role       = aws_iam_role.api-caller.name
-  policy_arn = "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
-}
-
 resource "aws_iam_role_policy_attachment" "invoker_url_lambda_execution" {
   count = var.use_api_gateway_v2 ? 0 : 1
 


infra/modules/aws/main.tf
```diff
@@ -103,11 +103,6 @@ resource "aws_iam_policy" "execution_lambda_to_caller" {
   }
 }
 
-resource "aws_iam_role_policy_attachment" "invoker_lambda_execution" {
```
jlorper, line 106: Are you sure this works? Do lambdas log because they have the rights to do so, or because the caller (that can do InvokeURL) also has logging permissions?
Also this policy is mentioned in various doc files, see https://github.com/search?q=repo%3AWorklytics%2Fpsoxy%20AWSLambdaBasicExecutionRole&type=code . Does it need to be dropped in certain cases?

eschultink: lambdas don't execute as this role

they each have their own execution role; and the lambda execution permission is attached to each here:
https://github.com/Worklytics/psoxy/blob/02b8dc5de0951b0cec3695634fdfa0705851ddb1/infra/modules/aws-psoxy-lambda/main.tf#L140

jlorper: got it. Looks good then

jlorper (approved):

aperez-worklytics (approved):

PR #771, opened 2024-08-08
GittHub: include email in SAML graphQL

aperez-worklytics:
Adding support for emails in GraphQL response for Github.

### Features
[Update Psoxy rules](https://app.asana.com/0/0/1208008421755524)

### Change implications

     obviously a no-op?
     change

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/rules/github/PrebuiltSanitizerRules.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/rules/github/PrebuiltSanitizerRules.java
@@ -111,7 +111,7 @@ public class PrebuiltSanitizerRules {
             .transform(Transform.Pseudonymize.builder()
                     .jsonPath("$..nameId")
                     .jsonPath("$..email")
-                    .jsonPath("$..value")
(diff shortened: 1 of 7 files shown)


java/core/src/main/java/co/worklytics/psoxy/rules/github/PrebuiltSanitizerRules.java
```diff
@@ -111,6 +111,7 @@ public class PrebuiltSanitizerRules {
             .transform(Transform.Pseudonymize.builder()
                     .jsonPath("$..nameId")
                     .jsonPath("$..email")
+                    .jsonPath("$..value")
```
eschultink, line 114: 🤔 this doesn't seem great; this will pseudsonymize any field named `value`, anywhere in the response. make it `$..emails[].value` or something like that? 

(not sure that's exactly correct JSONPath; but idea is .value fields below emails[] array)

eschultink (changes requested):
don't love the rule; can you make it a little more precise?

eschultink (approved):

PR #774, opened 2024-08-14
Fixing tokenization for MSFT cal, mail and teams

aperez-worklytics:
For outlook-cal, outlook-mail the tokenization link rule was generating full URL. For example, following URL:

```
https://graph.microsoft.com/v1.0/users/12345/calendar/calendarView?startDateTime=2020-01-01T00%3a00%3a00Z&endDateTime=2024-08-14T00%3a00%3a00Z&%24top=100&%24skip=100
```

was converted to:

```
https://graph.microsoft.com/p~7H9ZtewbbjdL-2NIg1WTcEF8MZ4UG4onBKBT7WGIfeU8j2Z36rBS9WPPHMeOTwVJZ6ReEPB45IEBANImwzxkLPAtL5Vha3VC0-iCP2AuKxT8LY3cWs2b9ucgPM2CZsndXuSLW0RlEiz0TbU65xNRT3PNR4g9y6SBwX-4Ql9tfgDVIFW4_GuJ7HvtEgWY_39s8_bGtqw3aFAUTQd3DAoDbrmK-2YDAWgOyhB8B_Tq84DUkh75VXdTT_jJvetcQgsNHUgDXp3oQahqslP-b1hGog
```

 Note that the conversion is mostly the host + rest of things, as rule used this regex: `^https://graph.microsoft.com/(.*)$")`, The token value is `v1.0/users/12345/calendar/calendarView?startDateTime=2020-01-01T00%3a00%3a00Z&endDateTime=2024-08-14T00%3a00%3a00Z&%24top=100&%24skip=100`, which is the path and the query of the request.

That had two issues when requesting the URL to the proxy generated by the tokenization,:
1) As the url is *host* + *path*, the reversed result is `v1.0/users....` which is not matching any rule, because rules expect to have the slash at the beginning: `/v1.0/...` That explain the blocked by rules error detected on the task associated to this PR.
2) Then, after reversing it the result is the URL without the tokenization: `v1.0/users/12345/calendar/calendarView?startDateTime=2020-01-01T00%3a00%3a00Z&endDateTime=2024-08-14T00%3a00%3a00Z&%24top=100&%24skip=100`. The problem is that we use tokenization to avoid passing *PII* values to the path, and in case of *no_app_id* rules endpoints should be only allowed with pseudonymized values, not the plain ones. That was working because rules for *no app id/ no groups app id* supports both: plain and pseudonymized. So the second change on this PR is the update of the rules for MSFT to only support right endpoints.
As a side effect when reviewing this, I detected that MSFT Teams rules were not right for links and some tokenization (it was rules for that that are not actually needed and probably it could break teams pagination on client) 

 [Tokenized paths not working?](https://app.asana.com/0/1207909702220871/1208022262261466)

 - []()

### Change implications

     obviously a no-op? **yes**
     change **no**


java/core/src/test/java/co/worklytics/psoxy/rules/msft/Calendar_NoAppIds_NoGroups_Tests.java
```diff
@@ -3,11 +3,12 @@
 import co.worklytics.psoxy.rules.Rules2;
 import lombok.Getter;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 //TODO: fix this re-use via inheritance; makes tests brittle; we should inject this rule set into
 // the directory tests, or something like that
-public class Calendar_NoAppIds_NoGroups_Tests extends CalendarTests {
```
aperez-worklytics, line 10: Note that all no app id tests, inherits from calendar... because they were used same set of rules + no app ids

docs/sources/microsoft-365/msft-teams/example-api-responses/sanitized_no-userIds/Communications_callRecords_v1.0.json
```diff
@@ -212,5 +212,5 @@
       ]
     }
   ],
-  "sessions@odata.nextLink":"https://graph.microsoft.com/p~qVi1Nt_0lOgW1F8drMalgXVuclaDbOeXpPWd2BvUvOsUGfPfqoz7UTpwS9flY4NcDvlsK1T4X6e9ZOiUksIJwVXZTzeqtAXgBYuO5SRB8vRv-m7pvZUsNE_RMw2rx-biUwF3RWbel3mKm1_hRPERQjCvT8xjXGuBpEI4UrrNEg4LHTyzNLYt_zyzYzUJRkAmeTAihCawIU94B60X1lxl0g"
-}
+  "sessions@odata.nextLink":"https://graph.microsoft.com/v1.0/$metadata#communications/callRecords('e523d2ed-2966-4b6b-925b-754a88034cc5')/sessions?$expand=segments&$skiptoken=abc"
```
eschultink, line 215: OK, as long as we're certain this is not a user acccount guid.

eschultink (approved):
OK, thanks. nice catch on the regex thing. 

I think I had hashed nextLink generally bc structure is opaque, so we are never *certain* it doesn't contain PII. By exempting it in some cases, you're presuming some knowledge about the contents. 

We might have to go back to always tokenizing it, bc that's the most conservative approach.  Although perhaps extreme - since microsoft uses uuid for everything, seeing uuid in a nextLink doesn't mean it refers to a user, let along which user it refers to.

PR #780, opened 2024-08-16
aws secrets manager fixes

eschultink:
### Fixes
 - global (shared) path not set properly in lambda env

### Change implications

### DIFF
--- a/infra/modules/aws-host/main.tf
+++ b/infra/modules/aws-host/main.tf
@@ -81,7 +81,7 @@ resource "aws_iam_role_policy_attachment" "invoker_url_lambda_execution" {
 
 # secrets shared across all instances
 locals {
-  path_to_global_secrets = var.secrets_store_implementation == "aws_secrets_manager" ? var.aws_secrets_manager_path : var.aws_ssm_param_root_path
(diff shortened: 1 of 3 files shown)


infra/modules/aws-host/main.tf
```diff
@@ -163,7 +167,7 @@ module "api_connector" {
   global_parameter_arns                 = try(module.global_secrets_ssm[0].secret_arns, [])
   global_secrets_manager_secret_arns    = try(module.global_secrets_secrets_manager[0].secret_arns, {})
   path_to_instance_ssm_parameters       = "${local.instance_ssm_prefix}${replace(upper(each.key), "-", "_")}_"
-  path_to_shared_ssm_parameters         = var.aws_ssm_param_root_path
+  path_to_shared_ssm_parameters         = local.path_to_shared_secrets
```
eschultink, line 170: this would basically be always wrong if someone actually wanted to prefix the shared variables. but none of the beta users of AWS Secrets Manager have actually used this, so haven't hit it. We just did in demo env when testing, as needed prefixes to avoid collisions in shared AWS accounts.

jlorper (approved):

PR #782, opened 2024-08-16
Tokenize MSFT Teams PSTN call ids

aperez-worklytics:
Reviewing #778 I've seen the `callId` may contain the IP of the caller.

### Features
[Psoxy: tokenize PSTN callId](https://app.asana.com/0/1208032421449960/1208075034541288)

### Change implications

 - dependencies added/changed? **no**
 - something important to note in future release notes?
   - NOTE in `CHANGELOG.md` anything that will show up in `terraform plan`/`apply` that isn't
     obviously a no-op? **no**
   - breaking changes? if in module/example that is NOT marked `alpha`, requires major version
     change

PR #783, opened 2024-08-16
Fix invalid link files

aperez-worklytics:
Meanwhile doing #782, seen that since v.4.58 some invalid links (gcal and qualitrics) were preventing to run the build for tests in intelliJ. Worked fine from mvn command line, but not in the IDE with following error: *Maven Resources Compiler: Failed to copy [...]*:

- In case of gcal, fixing the link path
- In case of qualitrics, adding a README to the folder (so the link will be right). 

 - []()

 - []()

### Change implications

     obviously a no-op?
     change

### DIFF
--- a/docs/sources/gcal
+++ b/docs/sources/gcal
@@ -1,1 +1,1 @@
-google-workspace/gcal/README.md
+google-workspace/calendar/README.md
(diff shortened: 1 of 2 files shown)


eschultink (approved):
thanks yes - had also fixed on a different branch

PR #784, opened 2024-08-16
bump ephermal amount for aws bulk mode

eschultink:
### Features
 - increase ephemeral storage amount; free default wasn't enough in practice for some users

### Change implications

 - something important to note in future release notes? **yes, users will see a change on apply; and possibly trivial cost increase**

### DIFF
--- a/infra/modules/aws-psoxy-lambda/variables.tf
+++ b/infra/modules/aws-psoxy-lambda/variables.tf
@@ -119,7 +119,7 @@ variable "memory_size_mb" {
 variable "ephemeral_storage_mb" {
   type        = number
   description = "ephemeral storage size in MB"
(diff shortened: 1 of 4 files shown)


infra/modules/aws-psoxy-lambda/variables.tf
```diff
@@ -116,6 +116,12 @@ variable "memory_size_mb" {
   default     = 512
 }
 
+variable "ephemeral_storage_mb" {
+  type        = number
+  description = "ephemeral storage size in MB"
+  default     = 512 # this is the free amount; over this though it's pretty
```
jlorper, line 122: Incomplete comment?

jlorper (approved):

PR #788, opened 2024-08-19
Always use current directory to save files

davidfq:
### Fixes
 - [test bulk with arbitrary file fails](https://app.asana.com/0/1201039336231823/1208068981362738/f)

### Features
.

### Change implications


eschultink (approved):
thanks - tested, wfm 👍

PR #793, opened 2024-08-27
cut calendar fields

eschultink:
### Features
 - remove some calendar fields, just to simplify reviews

### Change implications

---
  - https://app.asana.com/0/0/1208065497055580


.idea/compiler.xml
```diff
@@ -41,11 +41,68 @@
         </processorPath>
         <module name="core" />
         <module name="psoxy-aws" />
-        <module name="gateway-core" />
         <module name="psoxy-gcp" />
         <module name="psoxy-cmd-line" />
         <module name="psoxy-core" />
       </profile>
+      <profile name="Annotation profile for gateway-core" enabled="true">
```
aperez-worklytics, line 48: This is how I have mine on my local, it seems to.

docs/sources/microsoft-365/outlook-cal/example-api-responses/sanitized_no-app-ids_no-groups/Events_v1.0.json
```diff
@@ -1036,12 +1000,9 @@
       "lastModifiedDateTime":"2021-02-25T01:31:48.0159305Z",
       "changeKey":"IiLKjG2I7E+Xv0+ys6MD0wADW1wcFg==",
       "categories":[ ],
-      "transactionId":null,
       "originalStartTimeZone":"UTC",
       "originalEndTimeZone":"UTC",
       "iCalUId":"040000008200E00074C5B7101A82E008000000004B36F3526728D3010000000000000000100000001E18A3027CF8964892BF097ACA3679E8",
-      "reminderMinutesBeforeStart":15,
```
aperez-worklytics, line 1043: I'd include this on the unit test too, to ensure this is redacted when sanitized

aperez-worklytics (approved):

docs/sources/microsoft-365/outlook-cal/example-api-responses/sanitized_no-app-ids_no-groups/Events_v1.0.json
eschultink: yeah, this is from the `sanitized` version used for tests. so should be covered by unit tests

.idea/compiler.xml
```diff
@@ -41,11 +41,68 @@
...
+          <entry name="$MAVEN_REPOSITORY$/org/jetbrains/kotlin/kotlin-stdlib-jdk7/1.5.30/kotlin-stdlib-jdk7-1.5.30.jar" />
+          <entry name="$MAVEN_REPOSITORY$/com/squareup/javapoet/1.13.0/javapoet-1.13.0.jar" />
+          <entry name="$MAVEN_REPOSITORY$/com/google/googlejavaformat/google-java-format/1.5/google-java-format-1.5.jar" />
+          <entry name="$MAVEN_REPOSITORY$/com/google/errorprone/javac-shaded/9-dev-r4023-3/javac-shaded-9-dev-r4023-3.jar" />
+          <entry name="$MAVEN_REPOSITORY$/net/ltgt/gradle/incap/incap/0.2/incap-0.2.jar" />
+          <entry name="$MAVEN_REPOSITORY$/org/jetbrains/kotlinx/kotlinx-metadata-jvm/0.3.0/kotlinx-metadata-jvm-0.3.0.jar" />
+          <entry name="$MAVEN_REPOSITORY$/org/projectlombok/lombok/1.18.30/lombok-1.18.30.jar" />
+        </processorPath>
+        <module name="gateway-core" />
+      </profile>
```
eschultink, line 105: @aperez-worklytics - how does about compare for you? i always have local changes made by IntelliJ here, but avoid committing usually - I can revert this unless seems consistent w yours.

aperez-worklytics: Same as me; I've compared two files (can send you mine if you prefer) and the differences are:
- I don't have this part:
```
    <module name="core" />
        <module name="psoxy-aws" />
        <module name="psoxy-gcp" />
        <module name="psoxy-cmd-line" />
```

and I have same profile as there is added here `Annotation profile for gateway-core`. Then, my differences is that I have more than one profile, one per module but with same but with same entries in *processor path* in all of them.

I think you can include yours here, not an issue for me.

docs/sources/microsoft-365/outlook-cal/example-api-responses/sanitized_no-app-ids_no-groups/Events_v1.0.json
aperez-worklytics: ok, but if the *original* file doesn't include that field we are not seeing the rule applied in the tests or in the files, but I was referring something like this:

https://github.com/Worklytics/psoxy/blob/6836d172f5de7dddcd48eecfdda21f4ed18b5638/java/core/src/test/java/co/worklytics/psoxy/rules/msft/CalendarTests.java#L35

so adding in the redacted assert the text  `reminderMinutesBeforeStart`, so at least we have something to check that prior the files we generate

jlorper (approved):

PR #800, opened 2024-09-06
OOO through proxy for meeting titles

eschultink:
### Features
  - allow OOO-related tokens through proxy rules

### Change implications


java/core/src/main/java/co/worklytics/psoxy/rules/generics/Calendar.java
```diff
@@ -42,16 +46,20 @@ public class Calendar {
...
-            .map(s -> Pattern.quote(s))
-            .collect(Collectors.joining("|"))
-        + ")";
+    public static final List<String> OOO_TITLE_SNIPPETS = Arrays.asList(
+        "OOO",
+        "OOO:",
+        "Out of Office",
+        "Out of Office:",
+        "Out of the Office",
+        "Out of the Office:"
```
aperez-worklytics, line 55: Worth to add?.. 

```
O-O-O
Out-of-Office
O-O-O:
Out-of-Office:
```

?

aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/rules/generics/Calendar.java
eschultink: that's not in tokens we currently deal with for Calendar 3 methodology.  I don't think those cases common enough to deal with.

(most calendar tools have native OOO fields anyways, so the parsing out of calendar titles imho is likely to be pretty incremental; but might as well do it for now bc I do think historically is still relevant)

PR #804, opened 2024-09-16
S183 : extended meeting title tokens

eschultink:
### Features
 - coming out of customer discussion, further expansion of tokens in meeting titles

### Change implications

 - something important to note in future release notes? **yes, although consistent with what is already noted for v0.4.60**


aperez-worklytics (approved):
Ok; for consistency we'll need to do the same for onlineMeetings in Teams:

https://github.com/Worklytics/psoxy/blob/cb265e5f6b413743882d22b5ddce2df704a39370/java/core/src/main/java/co/worklytics/psoxy/rules/msft/PrebuiltSanitizerRules.java#L445

I can do that in a different PR after this is merged.

PR #811, opened 2024-09-25
concise phrase transform

eschultink:
### Features
 - transform more suited to the calendar event title use-case

### Change implications

 - something important to note in future release notes? **not yet**

---
  - https://app.asana.com/0/0/1208336236290115


java/core/src/test/java/co/worklytics/psoxy/impl/RESTApiSanitizerImplTest.java
```diff
@@ -186,6 +186,27 @@ void redactRegexMatches(String source) {
     }
 
 
+    @SneakyThrows
```
jlorper, line 189: test by copilot?? typo PHASE instead of PHRASE everywhere :)

java/core/src/test/java/co/worklytics/psoxy/rules/generics/CalendarTest.java
```diff
@@ -53,19 +34,19 @@ public void focusTimeBlockTitleSnippets_noMatch(String input) {
...
-        "Secret Project Focus Time,Focus Time",
-        "Focus Time Block,Focus Time Block",
-        "Focus: Secret Project,Focus: ",
-        "No Meeting Wednesday,No Meeting ",
-        " No Meetings,No Meetings",
-        "Prep Time,Prep Time",
-        "Prep Customer Meeting,Prep ",
-        "Prep: Customer,Prep: ",
+        "Out of the Office: Vacation,Out of the Office",
+        "Focus Time,'Focus Time,Focus'",
```
eschultink, line 28: so, this is reality of this; bc transform products ALL matches, not just first. In effect, it's more like categorization than redaction.

behavior is needed if you want "Team weekly" to match both "team" and "weekly" cases. that will end up as `team,weekly`.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -301,6 +303,37 @@ String canonicalizeIp(String ip) {
...
+                    log.warning("value matched by " + transform + " not of type String");
+                }
+                return null;
+            } else if (StringUtils.isBlank((String) s)) {
+                return s;
+            } else {
+                return patterns.stream()
+                    .map(p -> p.matcher((String) s))
+                    .filter(Matcher::matches)
+                    .map(m -> m.group(1)) //group 1, bc we created caputuring group in regex above
```
jlorper, line 330:
```suggestion
                    .map(m -> m.group(1)) //group 1, bc we created capturing group in regex above
```

java/core/src/test/java/co/worklytics/psoxy/rules/generics/CalendarTest.java
```diff
@@ -16,56 +17,23 @@
...
+        "Secret Project Focus Time,'Focus Time,Focus'",
+        "Focus Time Block,'Focus Time,Focus'",
+        "Focus: Secret Project,Focus",
+        "No Meeting Wednesday,No Meeting",
+        " No Meetings,'No Meetings,No Meetings'", // q: why????
```
jlorper, line 33: Because matches with `No Meeting` and `No Meetings`

jlorper (approved):
a minor typo

java/core/src/test/java/co/worklytics/psoxy/rules/generics/CalendarTest.java
eschultink: 🤔  it shouldn't though ... `No Meeting` token string shouldn't match `No Meetings`, bc I'm transforming it into a pattern with `\b` tokens on either side to only match "word boundaries. So it shouldn't match prefixes.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -301,6 +303,37 @@ String canonicalizeIp(String ip) {
...
     }
 
+    MapFunction getRedactExceptPhrases(Transform.RedactExceptPhrases transform) {
+
+        //TODO: alternatively, all different patterns, and preserve ALL matches?
+        // --> means we might enlarge, if the same phrase matches multiple times
+
+        List<Pattern> patterns = transform.getAllowedPhrases().stream()
+            .map(p -> "\\Q" + p + "\\E") // quote it
+            .map(p -> "\\b" + p + "[\\\\s:]*\\b") //boundary match, with optional whitespace or colon at end
```
eschultink, line 313: @jlorper this is where should be limiting to word boundaries.

maybe bug is that `"[\\\\s:]*"` is wrong, so it's matching character 's' specifically, instead of `\s` for whitespace as I intended ...

eschultink: yeah, it was that - fixed now and works as expected.

aperez-worklytics (approved):

PR #819, opened 2024-10-04
drop zoom oauth refresh token

eschultink:
### Features
 - zoom account id, client id don't need to be sensitive

### Fixes
  - fix AWS parameter store stuff so `String` rather than `SecureString` used

TODO
  - [ ] test carefully that bug previously noted is no longer a problem; possibly pin AWS provider version to avoid it

### Change implications

 - something important to note in future release notes? **yes, ppl will see a change**

### DIFF
--- a/docs/sources/zoom/README.md
+++ b/docs/sources/zoom/README.md
@@ -81,3 +81,21 @@ sufficient, but as of May 2024 are no longer available for newly created Zoom ap
 Once the scopes are added, click on `Done` and then `Continue`.
 
 6. Activate the app
+
+## Troubleshooting
+
+### Zoom API Error : 400 invalid client
+
+`{"reason":"Invalid client_id or client_secret","error":"invalid_client"}`
+
+Causes:
+   - extra chars in Client ID; or incorrect Client ID
+
+Confirm that the `Client ID` and `Client Secret` are correctly set in your secret store solution (AWS Parameter Store, Secrets Manager, or GCP Secret Manager).
+
+### Zoom API Error : 400 Bad Request
+
+`{"reason":"Bad Request","error":"invalid_request"}`
+
+Causes:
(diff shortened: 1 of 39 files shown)


jlorper (approved):

infra/modules/aws-ssm-secrets/main.tf
```diff
@@ -41,17 +38,17 @@ resource "aws_ssm_parameter" "secret" {
...
+  type           = each.value.sensitive ? "SecureString" : "String"
+  description    = each.value.description
+  value          = each.value.sensitive ? sensitive(coalesce(each.value.value, local.PLACEHOLDER_VALUE)) : null
+  insecure_value = each.value.sensitive ? null : coalesce(each.value.value, local.PLACEHOLDER_VALUE)
+  key_id         = coalesce(var.kms_key_id, "alias/aws/ssm")
 
   lifecycle {
     ignore_changes = [
       value, # key difference here; we don't want to overwrite values filled by the external process
+      insecure_value,
```
aperez-worklytics, line 51: What happens if for example `client_id` is wrongly provided if is not sensitive? Does `value` will update `insecure_value` on second attempt ?

docs/sources/zoom/README.md
```diff
@@ -81,3 +81,21 @@ sufficient, but as of May 2024 are no longer available for newly created Zoom ap
 Once the scopes are added, click on `Done` and then `Continue`.
 
 6. Activate the app
```
aperez-worklytics, line 83: Maybe you can include an screenshot like this on this step too:

image

infra/modules/aws-ssm-secrets/main.tf
eschultink: if this is in lifecycle, then TF shouldn't change resource even if it believes changes are needed in EITHER `value` or `insecure_value`

PR #823, opened 2024-10-09
env var only config

eschultink:
### Features
  - only allow some config as env vars

### Change implications

 - something important to note in future release notes? **Yes**
     obviously a no-op? **if anyone is depending on overriding env vars with secrets/parameter store, this will break that**
     change


java/core/src/test/java/co/worklytics/psoxy/gateway/ProxyConfigPropertyTest.java
```diff
@@ -0,0 +1,43 @@
...
+        strings = {
+            "PSOXY_ENCRYPTION_KEY",
+            "ENCRYPTION_KEY_IP",
+            "RULES",
+            "PSOXY_SALT",
+            "SALT_IP",
+        }
+    )
+    @ParameterizedTest
+    public void remoteConfigVars(String paramName) {
```
jlorper, line 23: :red_circle: missing test body

java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java
```diff
@@ -61,6 +61,10 @@ public SecretManagerConfigService(@Assisted("projectId") @NonNull String project
 
     @Override
     public void putConfigProperty(ConfigProperty property, String value) {
+        if (property.isEnvVarOnly()) {
```
jlorper, line 64: Tip: I like preconditions for these kind of checks.
`Preconditions.checkArgument(!property.isEnvVarOnly(), "Can't put env-only config property: " + property);`

jlorper (changes requested):
missing test body

aperez-worklytics (approved):
OK after @jlorper changes

jlorper (approved):

PR #828, opened 2024-10-28
improve error handling, cleanup config

eschultink:
### Features
 - improve error handling for a couple of cases we've seen
 - remove deprecated usage of isDevelopmentMode env var

### Change implications

     obviously a no-op?
     change

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
@@ -21,8 +21,8 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import lombok.NoArgsConstructor;
 import lombok.SneakyThrows;
+import lombok.Value;
 import lombok.extern.java.Log;
-import org.apache.commons.lang3.ObjectUtils;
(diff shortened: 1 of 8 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -147,30 +149,14 @@ public HttpEventResponse handle(HttpEventRequest request) {
...
-
-        // Using original URL to check sanitized rules, as they should match the original URL. It could contain tokenized components.
-        // Examples:
-        // /v1/accounts/p~12adsfasdfasdf31
-        // /v1/accounts/12345
-        URL originalRequestedURL = new URL(requestedTargetUrl);
-        // And the URL to use for source request; it could contain the reversed tokenized components
-        URL targetForSourceApiRequest = new URL(clearTargetUrl);
-
+        RequestUrls requestUrls = buildRequestedUrls(request);
```
jlorper, line 157: I'd wrap from here the try-catch that starts in L201

eschultink: added more

jlorper (approved):

PR #833, opened 2024-11-11
Add teams with default scopes

aperez-worklytics:
`msft-teams` should use `azure-ad` as default scope, but it was using a deprecated property on that and no value was provided, leaving it as the same as source.

### Fixes
[Psoxy: msft-teams scope](https://app.asana.com/0/0/1208737224528874)

 - []()

### Change implications

     obviously a no-op?
     change


eschultink (approved):
yeah - this is fine; although it shouldn't affect behavior of proxy for anyone not using LEGACY pseudonyms

PR #836, opened 2024-11-28
Slack conversations ids support

aperez-worklytics:
### Features
  - [Psoxy support for new Slack API endpoint](https://app.asana.com/0/0/1208864444154089)

### Change implications

     change **no**

### DIFF
--- a/docs/sources/slack/discovery.yaml
+++ b/docs/sources/slack/discovery.yaml
@@ -53,6 +53,7 @@ endpoints:
       - !<pseudonymize>
         jsonPaths:
           - "$.messages[*].user"
+          - "$.messages[*].inviter"
           - "$.messages[*].files[*].user"
           - "$.messages[*].reactions[*].users[*]"
           - "$.messages[*].replies[*].user"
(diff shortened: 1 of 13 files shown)


docs/sources/slack/discovery.yaml
```diff
@@ -11,6 +11,10 @@ endpoints:
       - !<pseudonymize>
         jsonPaths:
           - "$.users[*].id"
+        includeReversible: true
+        encoding: "URL_SAFE_TOKEN"
```
aperez-worklytics, line 15: Now *userId* from listing users needs to be url safe, as it is provided as as parameter for user conversation endpoint.

docs/sources/slack/discovery.yaml
```diff
@@ -44,6 +53,7 @@ endpoints:
       - !<pseudonymize>
         jsonPaths:
           - "$.messages[*].user"
+          - "$.messages[*].inviter"
```
aperez-worklytics, line 56: Seen this during my tests, so include it to be pseudonymized

eschultink (approved):

PR #841, opened 2024-12-04
bump java source/target to java17

eschultink:
### Fixes
 - bad merge issues - changes made in 0.4.62 AFTER branched for 0.5 which conflict with 0.5 stuff; mainly test cases that should have been updated with `scope` being dropped, but weren't on the base

### Features
 - java source target to java 17

### Change implications

 - dependencies added/changed? **yes - now MUST have java17; 11 not supported**
 - something important to note in future release notes? **java17**

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,91 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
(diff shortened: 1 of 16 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,91 @@
...
+#
+name: "CodeQL Advanced"
+
+on:
+    push:
+        branches: [ "main", "rc-*" ]
+    pull_request:
+        branches: [ "main", "rc-*" ]
+    schedule:
+        - cron: '25 4 * * 5'
```
aperez-worklytics, line 20: Better on monday? So we could fix it quickly instead of having it broken during the weekend

aperez-worklytics (approved):

.github/workflows/codeql.yml
eschultink: this was github's generated default, when converting from the basic CodeQL scanning (default) to the advanced (explicit yaml-defined workflow you can customize)

PR #844, opened 2024-12-10
GCP secret access robustness

eschultink:
### Fixes
  - logic to access GCP secrets via explicit version stored on label on the secret MAY be causing issues, due to the label value being stale if someone updates the token outside of our code

### Features
  - more logging around GCP secret access
  - more logging when oauth refresh tokens updated

### Change implications

---
  - https://app.asana.com/0/0/1208948645711048


aperez-worklytics (approved):
OK; LGMT

jlorper (approved):

PR #847, opened 2024-12-11
misc zoom cleanup

eschultink:
parts of #845 actually worth keeping 

### Features
  - zoom client id not sensitive; document that zoom renders it in the clear in consoles
  - add some Zoom trouble shooting insights
  - improve logs/warnings, add some trims - to avoid misconfigurations of Zoom connector as well as provide better error feedback

### Change implications

 - something important to note in future release notes? **flipping zoom client ID -might- show changes in tf plan; we'll need to watch in pre-release testing, although i believe it will at most appear pretty innocuous**

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/AccountCredentialsGrantTokenRequestBuilder.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/AccountCredentialsGrantTokenRequestBuilder.java
@@ -8,6 +8,7 @@ import com.google.api.client.http.HttpHeaders;
 import com.google.api.client.http.UrlEncodedContent;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
+import org.apache.commons.lang3.StringUtils;
 
 import javax.inject.Inject;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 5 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/AccountCredentialsGrantTokenRequestBuilder.java
```diff
@@ -60,23 +61,26 @@ public enum ConfigProperty implements ConfigService.ConfigProperty {
 
     @Override
     public HttpContent buildPayload() {
+
+        //trim to avoid copy-paste errors
+        //q : check for non-printable chars or anything like that
```
aperez-worklytics, line 66: yeah, I'd include a warning if accountId != trim(accountId)

aperez-worklytics (approved):

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/AccountCredentialsGrantTokenRequestBuilder.java
eschultink: that will do the whitespace before/after, but easy to trim out; my concern is encapsulated stuff, like if they mixed up encodings or something.

PR #863, opened 2025-03-17
Zoom meeting summary

aperez-worklytics:
Supporting `meetings/{MEETING_ID}/meeting_summary` endpoint for Zoom. Redacted all sensitive fields there too.

 - []()

### Features
[Update psoxy rules](https://app.asana.com/0/0/1209656368911012)

### Change implications

 - dependencies added/changed? **no**
 - something important to note in future release notes?
   - NOTE in `CHANGELOG.md` anything that will show up in `terraform plan`/`apply` that isn't
     obviously a no-op? **yes**
   - breaking changes? if in module/example that is NOT marked `alpha`, requires major version
     change **no**

PR #864, opened 2025-03-25
Microsoft 365 Copilot support

aperez-worklytics:
**NOTE** Parent PR is #863 to use `TextDigest` transformation

Support for Microsoft 365 connector:
- Adding rules for regular and *no app id* version
- Expanding `TextDigest` to process JSON escaped content

### Features
[Add psoxy connector](https://app.asana.com/0/0/1209672290439370)

### Change implications

     obviously a no-op? **no**
     change **no**


docs/sources/microsoft-365/msft-copilot/README.md
```diff
@@ -0,0 +1,50 @@
+# Microsoft Copilot
```
eschultink, line 1: mark this as _alpha_; i think worth releasing it in this form, but 1) source API is only beta so unstable; and 2) I have a couple high-level doubts about our design here:

1) I think we should use rule-set with `responseSchema`, as this is stronger approach; rather than specifying what we redact, specify only what gets included.
2) `textDigest` - i worry that this just replacing value at a node with new value may not make it clear enough that value has been transformed.  maybe we need some convention that replaces node with digest of a different name - eg, `content~digest` or something. This would generally allow clients to consume it if and only if understand it.  But i guess we use replacement for pseudonym case now anyways - we just do something closer to format-preservation of the value in such cases (or can, for non-JSON format output of pseudonyms)

eschultink (approved):
cool - I think definitely worth releasing as an `alpha`; just mark as such throughout the docs.

docs/sources/microsoft-365/msft-copilot/README.md
aperez-worklytics: Agree on both; I think this case with TextDigest is a bit particular, probably we can improve it in next releases.

PR #873, opened 2025-04-08
better `User-Agent` string

eschultink:
### Features
 - improve `User-Agent` string sent to API servers

### Change implications

---
  - https://app.asana.com/0/0/1209899304445589

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
@@ -19,9 +19,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
-import lombok.NoArgsConstructor;
-import lombok.SneakyThrows;
-import lombok.Value;
+import lombok.*;
 import lombok.extern.java.Log;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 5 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -430,11 +431,19 @@ HttpRequestFactory getRequestFactory(HttpEventRequest request) {
 
         ComposedHttpRequestInitializer initializer =
                 ComposedHttpRequestInitializer.of(initializeWithCredentials,
-                        new GzipedContentHttpRequestInitializer("Psoxy"));
+                        new GzipedContentHttpRequestInitializer(getProxyUserAgentString()));
```
jlorper, line 434: I don't think this does anything at all, as is used as fallback

https://github.com/Worklytics/psoxy/blob/331a45abe814ae046b5f37f8fb4907aa17b9d83d/java/core/src/main/java/co/worklytics/psoxy/utils/GzipedContentHttpRequestInitializer.java#L25C9-L26C39

so presumably we are sending this instead always:

https://github.com/googleapis/google-http-java-client/blob/f2e14545476c83ed36ef46839a00f2fa000acf75/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java#L67

Also, should that class be  `Gzipped...` with double P?

jlorper (changes requested):

eschultink:
closing in favor of https://github.com/Worklytics/psoxy/pull/1161

PR #877, opened 2025-04-21
 new slack rules

eschultink:
### Fixes

- Addressed scope deserialization issue (link to relevant issue if available)

### Features

- slack rules that enforce response schemas

### Change implications

- something important to note in future release notes? **yes; slack rules change**


docs/sources/slack/example-api-responses/sanitized/discovery-users-list.json
```diff
@@ -13,23 +12,14 @@
         "team":"T02DDBANBM5"
       },
       "is_admin":true,
-      "is_owner":true,
-      "is_primary_owner":false,
-      "is_restricted":false,
-      "is_ultra_restricted":false,
-      "is_bot":false,
```
jlorper, line 20: should we let `is_bot` pass always?

eschultink: yes; it was passing as `bot`; there's a `@JsonAlias("is_bot")` on the property, but seems to not be coming through;

docs/sources/slack/example-api-responses/sanitized/discovery-conversations-info.json
```diff
@@ -8,7 +8,7 @@
       "is_archived":false,
       "is_general":false,
       "unlinked":0,
-      "creator":"{\"hash\":\"MOMSVXXFHYCrH2e8Zrt_0G9lcd_GK4cUKRXBmO2QAa0\"}",
+      "creator":"MOMSVXXFHYCrH2e8Zrt_0G9lcd_GK4cUKRXBmO2QAa0",
```
jlorper, line 11: why is the hash / json gone? I don't find the change that makes that happen... shouldn't at least be t~HASH ??

jlorper (approved):
Ok to the is_bot, but one doubt about why the hashes are gone now... I don't see where that change comes from.

docs/sources/slack/example-api-responses/sanitized/discovery-conversations-info.json
eschultink: it's just the format change; default is now `URL_SAFE_TOKEN` instead of `JSON`

eschultink: but yeah, i agree should have `t~` for claritify - I've found a path through which that wasn't being added

eschultink: seems like not anything introduced here; but path that was never hit bc we used JSON format by default in most places.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -512,21 +512,7 @@ public MapFunction getPseudonymizeRegexMatches(Transform.PseudonymizeRegexMatche
...
-                        pseudonymizedString = UrlSafeTokenPseudonymEncoder.TOKEN_PREFIX + pseudonymizedIdentity.getHash();
-                    } else {
-                        pseudonymizedString = pseudonymizedIdentity.getReversible();
-                    }
-                } else {
-                    pseudonymizedString = UrlSafeTokenPseudonymEncoder.TOKEN_PREFIX + pseudonymizedIdentity.getHash();
-                }
-                if (pseudonymizedIdentity.getDomain() != null) {
-                    pseudonymizedString += UrlSafeTokenPseudonymEncoder.DOMAIN_SEPARATOR + pseudonymizedIdentity.getDomain();
-                }
```
eschultink, line 529: all of the above is internal to `UrlSafePseudonymEncoder`

docs/sources/slack/example-api-responses/sanitized/discovery-users-list.json
```diff
@@ -2,77 +2,52 @@
...
-      "is_email_confirmed":false,
-      "who_can_share_contact_card":"EVERYONE",
-      "is_invited_user":true,
       "teams":[
         "T02DDBANBM5"
       ]
     },
     {
-      "id":"p~mJlhl3xv52fNY6wyUps5mBYPCl1nVWHT442eQ7v-lJyurnRXe1xte4bUgJLOS0DP",
+      "id":"t~mJlhl3xv52fNY6wyUps5mBYPCl1nVWHT442eQ7v-lJw",
```
jlorper, line 39: :red_circle: this needs to be reversible for the new endpoint user.conversations, see https://github.com/Worklytics/psoxy/pull/836

jlorper (changes requested):
the user id needs to be reversible

docs/sources/slack/example-api-responses/sanitized/discovery-users-list.json
eschultink: good catch; bug in rules generation wasn't respecting the `includeReversible` fix; corrected now.

jlorper (approved):

PR #878, opened 2025-04-22
avoid auth config fallback

eschultink:
### Features
 - avoid some cases of config fallback
 - refactor to avoid lots of booleans in constructors

### Change implications

---
  - https://app.asana.com/0/0/1206510514036210
  - https://app.asana.com/0/0/1209961501149179

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
@@ -18,8 +18,7 @@ public interface ConfigService {
 
         String name();
 
-        // shared? (across multiple instances?)
-        // local? (per-instance secrets)
+
 
         // sensitive? (eg, value should not be exposed to 3rd parties)
         // secret? (eg, value should be handled as a secret; obscured/acl even internally; avoid in logs, etc)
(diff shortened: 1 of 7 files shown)


aperez-worklytics (commented):
ok, but I don't see where flags are currently used to get them only through env or env+secrets; is that going to be in a different PR ?

java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
```diff
@@ -38,7 +37,29 @@ default Boolean noCache() {
...
+
+            /**
+             * config value MUST be set in the remote source; env var is NOT Supported
+             */
+            REMOTE,
+
+            /**
+             * config value may be set either via environment variable or remote config, with former taking precedence
+             */
+            ENV_VAR_OR_REMOTE,
```
jlorper, line 62: nitpick: I'd probably remove this and use a List of supported values, with order implicit in list.
define a constant here
`List<SupportedSource> ENV_VAR_OR_REMOTE = List.of(ENV_VAR, REMOTE)`

jlorper: said this, I don't think this will grow... so fine as is

jlorper (approved):

java/core/src/main/java/co/worklytics/psoxy/gateway/ConfigService.java
eschultink: yeah. I don't think it will grow.

I do think having these as some kind of `Option`, with a variadic constructor `(Option...)` would be more readable - the default case (either) not so straightforward though.

Going to leave as-is for now; easy future change.

PR #879, opened 2025-04-22
GCP from Cloud Functions gen1 to Cloud Run Functions gen2

eschultink:
### Features
  - migrate GCP to Cloud Run Gen2, as this now seem *strongly* preferred by GCP and has various advantages

### Change implications

 - dependencies added/changed? **yes - depends on having some additional perms (CLoud Run Developer); will also require enabling some APIs, but our terraform modules will handle that internally**
 - something important to note in future release notes? **yes - add the role, prob wait for v0.6 with this, as it's technically breaking**

---
  - https://app.asana.com/0/0/1206108944334925

### DIFF
--- a/infra/modules/gcp-psoxy-rest/main.tf
+++ b/infra/modules/gcp-psoxy-rest/main.tf
@@ -54,7 +54,7 @@ resource "google_cloudfunctions2_function" "function" {
   location = var.region
 
   build_config {
-    runtime           = "java17"
+    runtime           = "java21"
     docker_repository = var.artifact_repository_id
     entry_point       = "co.worklytics.psoxy.Route"
 
--- a/infra/modules/gcp-psoxy-rest/main.tf
+++ b/infra/modules/gcp-psoxy-rest/main.tf
@@ -46,43 +46,53 @@ resource "google_service_account_iam_member" "act_as" {
 }
 
 
-resource "google_cloudfunctions_function" "function" {
+resource "google_cloudfunctions2_function" "function" {
   name        = "${var.environment_id_prefix}${var.instance_id}"
   description = "Psoxy Connector - ${var.source_kind}"
-  runtime     = "java17"
-  project     = var.project_id
-  region      = var.region
 
-  trigger_http                 = true
-  https_trigger_security_level = "SECURE_ALWAYS"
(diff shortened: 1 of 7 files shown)


infra/modules/gcp-psoxy-rest/main.tf
```diff
@@ -46,44 +46,54 @@ resource "google_service_account_iam_member" "act_as" {
...
+  project  = var.project_id
+  location = var.region
+
+  build_config {
+    runtime           = "java21"
+    docker_repository = var.artifact_repository_id
+    entry_point       = "co.worklytics.psoxy.Route"
+
+    source {
+      storage_source {
```
aperez-worklytics, line 62: I was thinking if this is still required, but maybe we can start using artifact repository if possible instead of using buckets for this. We have a todo for that: https://github.com/Worklytics/psoxy/pull/879/files#diff-1863f04a027d498e88a5f196aab09d85de06ea0cfce398427b4444cc07abcd1aL34 so maybe we can do this in this or next version?

infra/modules/psoxy-constants/main.tf
```diff
@@ -460,11 +460,16 @@ locals {
       display_name    = "Service Usage Admin",
       description_url = "https://cloud.google.com/iam/docs/understanding-roles#serviceusage.serviceUsageAdmin"
     },
-    "roles/cloudfunctions.admin" = {
+    "roles/cloudfunctions.admin" = { # q: can we replace this with Cloud Run Developer ??
```
aperez-worklytics, line 463: it seems so

aperez-worklytics (approved):

infra/modules/gcp-psoxy-rest/main.tf
eschultink: yeah, i saw that and looked at it a bit.

it would appear that cloud functions is creating Artifact Registry underneath automatically; perhaps if we create and then specify one explicitly, that's avoided?

aperez-worklytics: I think that was the original idea, but we faced with the update of google provider to support it to have it with the right clean policies.

Ideally, I'd like to try with artifact repository to see if we can improve deployment time with multiple functions too.

PR #880, opened 2025-04-25
avoid email parser dependency

eschultink:
### Features
  - replace email parser/validator lib with combination of their stuff (apache 2 license) copied over and our own cleaner/simpler wrappers

why? they have a weird jakarta/javax dep that's problematic and unhelpful.  In particular, conflicts with some deps in core worklytics codebase, requiring exclusion stuff; and generally messy to deal with

### Change implications

  - The dependency on `EmailParser` has been removed. Please check the project's dependency list to ensure compatibility with the new validation changes.
- breaking changes? **shouldn't be; certainly risk of minor changes to email parsing/validation**

---
  - https://app.asana.com/0/0/1209929312359673

### DIFF
--- a/java/core/pom.xml
+++ b/java/core/pom.xml
@@ -128,20 +128,6 @@
             <version>${dependency.google-auth-library-oauth2-http.version}</version>
         </dependency>
 
-
-        <!-- RFC 2822 validator -->
-        <!-- see https://github.com/bbottema/email-rfc2822-validator -->
-        <dependency>
-            <groupId>com.github.bbottema</groupId>
-            <artifactId>emailaddress-rfc2822</artifactId>
-            <version>2.3.1</version>
-        </dependency>
-        <dependency> <!-- for email validation to work in lib above; used in RESTApiSanitizerImpl, PseudonymizerImpl as of June 2023 -->
-            <groupId>com.sun.mail</groupId>
-            <artifactId>jakarta.mail</artifactId> <!-- CDDL + GPLv2 with classpath exception license, so ideally would remove -->
-            <version>[1.6.3,1.9.9)</version>
-        </dependency>
-
         <!-- explicit direct dep on apache HTTP Client (psoxy-core has transitive dep on it) -->
         <dependency>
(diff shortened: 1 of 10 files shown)


java/core/src/main/java/co/worklytics/psoxy/utils/email/EmailAddress.java
```diff
@@ -0,0 +1,21 @@
+package co.worklytics.psoxy.utils.email;
+
+import lombok.Builder;
+import lombok.Value;
+
+@Value
+@Builder
+public class EmailAddress {
```
eschultink, line 8: our own structure, in lieu of javax / jakarta `InternetAddress` stuff; i do think this possibly removes/changes some implicit IP/domain validation ... but don't care so much for our use-case

java/core/pom.xml
```diff
@@ -128,20 +128,6 @@
...
-        <!-- see https://github.com/bbottema/email-rfc2822-validator -->
-        <dependency>
-            <groupId>com.github.bbottema</groupId>
-            <artifactId>emailaddress-rfc2822</artifactId>
-            <version>2.3.1</version>
-        </dependency>
-        <dependency> <!-- for email validation to work in lib above; used in RESTApiSanitizerImpl, PseudonymizerImpl as of June 2023 -->
-            <groupId>com.sun.mail</groupId>
-            <artifactId>jakarta.mail</artifactId> <!-- CDDL + GPLv2 with classpath exception license, so ideally would remove -->
-            <version>[1.6.3,1.9.9)</version>
```
eschultink, line 142: this was optional dep of their email address parser, which was tricky to manage - as we also have it in worklytics code base - so some conflicts between them.

additionally - the CDDL + GPLv2 stuff was possibly a problem - certainly a more exotic license type to contend with; and really all they used out of here was `InternetAddress` pojo (? or does it validate domains???) - which wasn't adding value for our use cases.  We don't care about email addresses being *valid* in the sense of sending mail to them ... just syntactically/pragmatically valid.

jlorper (approved):

PR #887, opened 2025-05-07
fix node build

eschultink:
### Fixes
 - try to fix nodejs24 incompatibility

### Change implications

   - The version of Google Cloud Logging has been updated to address compatibility issues with Node.js 24.
   - No additional notes are necessary.
   - There are no breaking changes; this is a straightforward dependency bump to fix compatibility issues.

### DIFF
--- a/.github/workflows/ci-tools.yaml
+++ b/.github/workflows/ci-tools.yaml
@@ -14,9 +14,8 @@ jobs:
         # 16 is min version recommended to users; but unmaintained since 2023-10
         # 18 released 2022-04-19 - maintained until June 2025
         # 20 maintained until June 2026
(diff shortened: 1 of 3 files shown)


eschultink:
eh, nm - that didn't fix anything ... same problem still dep on same version of `buffer-equal-constant-time` via `@google-cloud/logging`

aperez-worklytics:
> eh, nm - that didn't fix anything ... same problem still dep on same version of `buffer-equal-constant-time` via `@google-cloud/logging`

FYI https://github.com/googleapis/google-auth-library-nodejs/issues/1650#issuecomment-2856673965

davidfq:
@eschultink I think we can close this one, right?

PR #889, opened 2025-05-13
finalize release v0.5.2

eschultink:
upgrade version number references from rc to rlease throughout the code base

### Change implications

none

---
  - https://app.asana.com/0/0/1210246445501431


java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
```diff
@@ -124,14 +119,13 @@ public Rules2 clone() {
     static ObjectMapper mapper = new YAMLMapper();
 
     public static synchronized Rules2 load(String path) {
```
aperez-worklytics, line 121: Ok, as part of #890 I've dropped this, as it was only used from slack

aperez-worklytics (approved):
Approved; after merge #891 we could have the release ready

jlorper (approved):

davidfq (approved):

PR #890, opened 2025-05-14
Fix slack loader

aperez-worklytics:
Slack was being loaded as a file path and not as a resource

### Fixes
[500s in gcp-hosted asana, others](https://app.asana.com/1/27145998307022/project/1201039336231823/task/1210246445501431)

 - []()

### Change implications

     obviously a no-op?
     change

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
@@ -117,22 +117,4 @@ public class Rules2 implements RESTRules {
             .build();
         return clone;
     }
-
-    //TODO: fix YAML serialization with something like
-    // https://stackoverflow.com/questions/55878770/how-to-use-jsonsubtypes-for-polymorphic-type-handling-with-jackson-yaml-mapper
-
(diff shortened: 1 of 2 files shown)


java/core/src/main/java/co/worklytics/psoxy/rules/Rules2.java
```diff
@@ -117,22 +117,4 @@ public Rules2 clone() {
...
-        try {
-            URL url = PrebuiltSanitizerRules.class.getClassLoader().getResource(path);
-            if (url == null) {
-                throw new IllegalArgumentException("No such file: " + path);
-            }
-            return mapper.readerFor(Rules2.class).readValue(Files.readAllBytes(Paths.get(url.toURI())));
-        } catch (IOException | URISyntaxException e) {
-            throw new Error(e);
-        }
-    }
```
eschultink, line 136: this is not really slack specific ... why not have in `Rules2?

aperez-worklytics: Well, it wasn't used in any other place after fixing Slack thing.

PR #899, opened 2025-05-28
Limiting GH Copilot Audit log requests

aperez-worklytics:
Updating Copilot rules to ensure only "copilot" events can be fetched from audit log when using Copilot connector

 - []()

### Features
[Ensure that only Copilot events are being requested on GH Copilot proxy audit log](https://app.asana.com/1/27145998307022/project/1210126515330332/task/1210351394180858)

### Change implications

     obviously a no-op? **no**
     change **no**


docs/sources/github/github-copilot.yaml
```diff
@@ -480,7 +480,10 @@ endpoints:
       - !<tokenize>
         jsonPaths:
           - "$..slug"
-  - pathTemplate: "/orgs/{org}/audit-log"
+  - pathRegex: "^/orgs/[^/]+/audit-log\\?(?=[A-Za-z0-9=&%+:\\._\\-]*phrase=action:copilot(?:\\\
```
eschultink, line 483: this is better accomplished with `ParameterSchema` constraints on the query parameters. see: https://docs.worklytics.co/psoxy/configuration/api-data-sanitization#path-query-parameter-schemas

I think will be more readable and easier to validate. Leave `pathTemplate` as is. add `phrase` as `queryParameterSchema`

eschultink (changes requested):

docs/sources/github/github-copilot.yaml
aperez-worklytics: Thanks; spent a while with this... Finally I had to use both, because agree that with query parameters the validation is easy to understand, but I also need that `phrase` and `include` parameters are always present when audit log is invoked, so I've supported that through regex.
We probably can implement something like *mandatoryParameters* or mark as part of *allowedParameters* the ones are mandatories, but I think this is the first case we have wth this and not sure if worth to do it right now.

docs/sources/github/copilot/github-copilot.yaml
```diff
@@ -480,9 +480,7 @@ endpoints:
       - !<tokenize>
         jsonPaths:
           - "$..slug"
-  - pathRegex: "^/orgs/[^/]+/audit-log\\?(?=[A-Za-z0-9=&%+:\\._\\-]*phrase=action:copilot(?:\\\
-    +created=[^&]+)?)(?=[A-Za-z0-9=&%+:\\._\\-]*include=web)[A-Za-z0-9=&%+:\\._\\\
-    -]+$"
+  - pathRegex: "^/orgs/[^/]+/audit-log\\\\?(?=.*[\\\\?&]phrase=)(?=.*[\\\\?&]include=).*$"
```
eschultink, line 483: i don't think you can have BOTH a `pathRegex` and a `pathTemplate` ... or at least, that shouldn't work.

eschultink (approved):

PR #901, opened 2025-05-30
Miro AI bulk support

aperez-worklytics:
Adding docs for defining a bulk importer for Miro Audit log CSV for Miro AI

 - []()

### Features
[Psoxy documentation](https://app.asana.com/1/27145998307022/task/1210426317485632)

### Change implications

     obviously a no-op? **no**
     change **no**

---
  - https://app.asana.com/0/0/1210426317485632


eschultink (approved):
OK - but:
 1) why not formalized in connector-specs, even if `alpha`?
 2) why not link from top `README`?  
 
I don't really understand why this is here, not just in Worklytics knowledge base. It's just a custom config of proxy for worklytics-specific use-case.

PR #902, opened 2025-06-03
Zoom user settings and cloud recordings endpoints

aperez-worklytics:
Open two new endpoints on Zoom: user settings and listing cloud recordings. A
Adding rules, examples, etc.
Documentation has been updated too.

 - []()

### Features
[Add psoxy support](https://app.asana.com/1/27145998307022/task/1210448932508138)

### Change implications

     obviously a no-op? **yes**
     change **no**

---
  - https://app.asana.com/0/0/1210448932508138
  - https://app.asana.com/0/0/1210487649633399


docs/sources/zoom/scopes.png
aperez-worklytics, line 1: Updated with all permissions to show how the page should like after adding all of them

docs/sources/zoom/example-api-responses/sanitized/user-settings.json
```diff
@@ -0,0 +1,285 @@
...
+    "call_out":true,
+    "call_out_countries":[
+      { }
+    ],
+    "show_international_numbers_link":true
+  },
+  "whiteboard":{
+    "out_meeting_advanced_whiteboard":true,
+    "in_meeting_advanced_whiteboard":true
+  }
```
eschultink, line 284: the above has a TON of stuff; should use a Json schema filter to cut down to what we need.

PR #903, opened 2025-06-03
more robust bulk file processing

eschultink:
### Fixes
 - syntactic-sugar loop `for (CSVRecord record : records)` can throw `UncheckedIOException` if individual line is malformed. This avoids that to some extent, by parsing line-by-line.

### Change implications

While arguably good to avoid failing a huge file due to a couple malformed lines, may lead to encoding issues being too subtle to be discovered ... so perhaps we don't want this.

---
  - https://app.asana.com/0/0/1210453538185821

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
@@ -32,10 +32,7 @@ import org.apache.commons.lang3.function.TriFunction;
 import org.apache.commons.lang3.tuple.Pair;
 
 import javax.inject.Inject;
-import java.io.IOException;
-import java.io.Reader;
-import java.io.UncheckedIOException;
-import java.io.Writer;
+import java.io.*;
 import java.util.*;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
@@ -76,8 +73,9 @@ public class ColumnarBulkDataSanitizerImpl implements BulkDataSanitizer {
         this.rules = rules;
     }
 
+
     @Override
-    public void sanitize(@NonNull Reader reader,
+    public void sanitize(@NonNull BufferedReader reader,
                          @NonNull Writer writer,
                          @NonNull Pseudonymizer pseudonymizer) throws IOException {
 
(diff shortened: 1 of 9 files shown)


java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -275,44 +276,90 @@ public void sanitize(@NonNull Reader reader,
...
-            ProcessingBuffer<ProcessedRecord> buffer = getRecordsProcessingBuffer(printer);
+    /**
+     * processes records into outputBuffer, applying the transformations to each
+     * @param columnNamesForOutput to render in the output buffer; so columns which DON'T have values in particular record will be filled
+     * @param columnTransformsToApply to apply to each column in the record
+     * @param bodyRecords to process; should be a CSVParser with headers
+     * @param outputBuffer to write the processed records to
+     * @return number of records which could not be processed due to errors
+     */
+    int processRecords(
```
aperez-worklytics, line 317: I'd return an object with the count of errors and processed lines too

java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -328,19 +375,20 @@ public void sanitize(@NonNull Reader reader,
...
                 });
 
-                if (buffer.addAndAttemptFlush(ProcessedRecord.of(Lists.newArrayList(newRecord.values())))) {
-                    log.info(String.format("Processed records: %d", buffer.getProcessed()));
-                };
-            }
-            if (buffer.flush()) {
-                log.info(String.format("Processed records: %d", buffer.getProcessed()));
+                if (outputBuffer.addAndAttemptFlush(ProcessedRecord.of(Lists.newArrayList(newRecord.values())))) {
+                    log.info(String.format("Processed records: %d", outputBuffer.getProcessed()));
```
aperez-worklytics, line 381: Yeah true, that appears in logs but just the buffer; I think if we expose the total as part of metadata could be useful

java/core/src/main/java/co/worklytics/psoxy/storage/impl/ColumnarBulkDataSanitizerImpl.java
```diff
@@ -87,8 +87,9 @@ public void sanitize(@NonNull Reader reader,
             .setTrim(true)
             .build();
 
+        ParsedFirstLine parsedFirstLine = parseFirstLine(reader);
```
aperez-worklytics, line 90: I'd include a comment here about why use this and not `records.getFirstEndOfLine()`

aperez-worklytics (changes requested):
I'd include the number of rows processed too as part of metadata.

And 🔴 ; pending to include that in `GcsFileEventHandler` for GCP (AWS and terminal are included here; missing GCP?)

PR #904, opened 2025-06-03
improve `SideOutput` design

eschultink:
### Features
 - move some general stuff out of host-specific `SideOutput` implementations
 - `contentEncoding` should be on `ProcessedContent` pojo

### Change implications


java/impl/aws/src/main/java/co/worklytics/psoxy/aws/S3SideOutput.java
```diff
@@ -36,27 +36,29 @@ public S3SideOutput(@Assisted("bucket") String bucket,
...
 
     @Override
     public void write(@NonNull HttpEventRequest request, @NonNull ProcessedContent content) {
         try {
             AmazonS3 s3Client = s3ClientProvider.get();
 
-            byte[] compressedContent =
-                sideOutputUtils.gzipContent(content.getContent(), content.getContentCharset());
+            //q: correct to do the compression *inside* of every SideOutput implementation?
+            //q: why not have Content-Encoding on ProcessedContent, compress conditionally?
```
aperez-worklytics, line 48: outdated questions? As finally you've included `contentEncoding` in `ProcessedContent`

aperez-worklytics (commented):
ok; does https://github.com/Worklytics/psoxy/pull/904/files#diff-13cad7c0c6986bc926b158c0c4f6eccfd88cfc3623a4f76b37b2e94d2099c6c0L281  thread usage needs to be fixed ?

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/S3SideOutput.java
eschultink: yeah - I have - thanks.

eschultink:
@aperez-worklytics  - I'm just removing no-response case, which is unused and untested, which makes the need for threading irrelevant. We should bring it back at some point ... but it's not immediately needed.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CommonRequestHandler.java
```diff
@@ -275,38 +271,24 @@ public HttpEventResponse handle(HttpEventRequest request) {
...
+                writeSideOutput(request, sourceApiResponse, sanitizationResult);
+
 
                 if (skipSanitization) {
-                    proxyResponseContent = original.getContent();
+                    proxyResponseContent = original.getContentString();
                 }
             } else {
                 //write error, which shouldn't contain PII, directly
                 log.log(Level.WARNING, "Source API Error " + original.getContent());
```
jlorper, line 289: sorry, late to this: should this be `original.getContentString()`  too?

PR #906, opened 2025-06-04
fix hybrid pseudonym deserialization

eschultink:
Case is customers legacy pseudonyms, but with a version that ALSO sends the new pseudonyms scheme to facilitate migration.

### Fixes
 - pseudonyms configured with legacy scheme, then serialized with proxy versions that actually fill the `h_4` proerpty with the new scheme (sometime after 0.4.39), aren't readable with v0.5.x libraries

### Change implications

 - something important to note in future release notes? **anything consuming pseudonyms that might be legacy scheme, but serialized with verions 0.4.40 - 0.4.62, should upgrade to this**

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/PseudonymizedIdentity.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/PseudonymizedIdentity.java
@@ -40,6 +40,19 @@ public class PseudonymizedIdentity {
      */
     String hash;
 
+    /**
+     * future-use. 0.4 hash for pseudonym, if `hash` is NOT 0.4.
+     *
+     * in the future, will be filled for LEGACY (0.3) cases; but for now, will be null (and always
+     * absent from JSON-serialized form)
+     *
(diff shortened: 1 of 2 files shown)


java/core/src/main/java/co/worklytics/psoxy/PseudonymizedIdentity.java
```diff
@@ -40,6 +40,19 @@ public class PseudonymizedIdentity {
...
+     *
+     * in the future, will be filled for LEGACY (0.3) cases; but for now, will be null (and always
+     * absent from JSON-serialized form)
+     *
+     * this will give both DEFAULT and LEGACY hashes for pseudonyms, allowing for eventual migration
+     * of LEGACY customers to DEFAULT (0.4)
+     *
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    String h_4;
```
eschultink, line 54: must maintain this until EVERYONE has migrated.

jlorper: do we need the scope back too??

jlorper (approved):

PR #909, opened 2025-06-16
webhook collectors 2

eschultink:
build out AWS webhook collector use-case fully, including authentication via JWT auth tokens and integrity checks of payloads AGAINST values in the tokens.

NOTE: testing tool does not yet support generating auth tokens or sending them to proxy instances, so that's not yet tested in a real lambda yet.

### Change implications

- **Dependencies added/changed?**  **AWS KMS lib required**


davidfq (commented):
Initial review: LGTM.
However, I have one concern about Data Flow **Scenario A** that we might want to discuss offline: the Client Tool sending the JWT in an `Authorization` header—although this is standard practice—could lead to CORS issues due to [preflight requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request). I haven’t reviewed the code in depth, but it appears that InboundWebhookHandler doesn’t handle these. In my last review of the JavaScript SDK, I changed the transport method to `navigator.sendBeacon` to avoid this issue in the Web Application `/in` endpoint.

java/impl/aws/src/main/java/co/worklytics/psoxy/SQSBatchMergeHandler.java
```diff
@@ -0,0 +1,52 @@
...
+import co.worklytics.psoxy.aws.SQSOutput;
+import co.worklytics.psoxy.gateway.ProcessedContent;
+import co.worklytics.psoxy.gateway.impl.BatchMergeHandler;
+import co.worklytics.psoxy.aws.DaggerAwsContainer;
+
+import java.nio.charset.StandardCharsets;
+import java.util.stream.Stream;
+
+/**
+ * @deprecated  - not sure we ultimately want/need this case, so marking as deprecated for now
```
aperez-worklytics, line 13: I think we can drop it, as the `WebhookCollectionModeHandler` will fit better. Other thing is if we'll need this in case of api v1/vpc stuff that is not *standard* in the configuration

java/impl/aws/src/main/java/co/worklytics/psoxy/WebhookCollectionModeHandler.java
```diff
@@ -0,0 +1,174 @@
...
+            APIGatewayV2HTTPResponse resp = APIGatewayV2HTTPResponse.builder()
+                .withStatusCode(400)
+                .withBody("Unsupported event type")
+                .build();
+            mapper.writeValue(output, resp);
+        }
+    }
+
+    private boolean isSQSEvent(JsonNode node) {
+        return node.has("Records") &&
```
aperez-worklytics, line 96: What happen if it is a function and the response contains that structure? I think we don't have that in the source, but wondering if we could face an issue here if we compare that structure and not other thing (not sure if we could check headers or other stuff apart from json payload to help identifying this)

infra/modules/aws-webhook-collector/main.tf
```diff
@@ -0,0 +1,406 @@
...
+
+  environment_name                     = var.environment_name
+  instance_id                          = var.instance_id
+  handler_class                        = var.handler_class
+  path_to_function_zip                 = var.path_to_function_zip
+  function_zip_hash                    = var.function_zip_hash
+  function_env_kms_key_arn             = var.function_env_kms_key_arn
+  logs_kms_key_arn                     = var.logs_kms_key_arn
+  memory_size_mb                       = var.memory_size_mb
+  timeout_seconds                      = 55
```
aperez-worklytics, line 77: as var with a default value ?

java/impl/gcp/src/main/java/co/worklytics/psoxy/GcpEnvironment.java
```diff
@@ -21,7 +21,7 @@ public String getInstanceId() {
 
 
     @Override
-    public Set<String> getSupportedSideOutputUriProtocols() {
-        return Set.of("gs://");
+    public Set<String> getSupportedOutputKinds() {
+        return Set.of("gs", "pubsub", "bq");
```
aperez-worklytics, line 25: As infra not ready yet, I'd not include pubsub or bq yet -just when implemented

aperez-worklytics (approved):
Some comments; I think OK

java/core/src/main/java/co/worklytics/psoxy/gateway/auth/PublicKeyRef.java
```diff
@@ -0,0 +1,34 @@
+package co.worklytics.psoxy.gateway.auth;
+
+import lombok.Value;
+
+/**
+ * reference to a k
+ */
+@Value
+public class PublicKeyRef {
```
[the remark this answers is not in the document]
eschultink: mmm, yeah i haven't used that feature at all. but makes sense.

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/S3Output.java
```diff
@@ -0,0 +1,89 @@
...
+
+    final String bucket;
+
+    final String pathPrefix;
+
+    @Inject
+    Provider<AmazonS3> s3ClientProvider;
+
+    @AssistedInject
+    public S3Output(@Assisted OutputLocation location) {
```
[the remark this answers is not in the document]
eschultink: I don't follow what you mean.

afiak, `@AssistedInject` annotated constructor will be used by the `@AssistedFactory`, which will be put on an interface that has same parameters as the annotated constructor. eg

```java
@AssistedFactory
public interface S3OutputFactory extends OutputFactory<S3Output> {

    S3Output create(@Assisted OutputLocation location);

```


dagger will generate that method body, to call the the constructor with that `location` parameter; and then ALSO use field-injection for all properties of the class (`S3Output`) that use field-injection. 

So I believe cleanest approach is a constructor with ONLY the parameters to the class, not the dependencies.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/BatchMergeHandler.java
```diff
@@ -0,0 +1,103 @@
...
+    public void handleBatch(Stream<ProcessedContent> batch) {
+        // Implementation for handling a batch of ProcessedContent
+        // This could involve aggregating, transforming, or writing the content to an output
+        // For example, you might write each item in the batch to a single output location
+        // or perform some aggregation logic before writing.
+
+        // combine into single processes content, call output.
+
+        // create a gzipped stream of the batch content
+        try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
jlorper, line 55: set-up a nice buffer size upfront? Like 1Mb ?

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/InboundWebhookHandler.java
```diff
@@ -0,0 +1,197 @@
...
+                }
+            })
+            .findFirst().isPresent();
+
+        if (!signatureValid ) {
+            return Optional.of("JWT verification failed due to invalid signature");
+        }
+        if (signedJWT.getJWTClaimsSet() == null) {
+            return Optional.of("Auth token invalid because JWT claims set is null");
+        }
```
jlorper, line 171: check this error first, no need to verify if we know is invalid.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/InboundWebhookHandler.java
```diff
@@ -0,0 +1,197 @@
...
+    @SneakyThrows
+    public Optional<String> validate(SignedJWT signedJWT, Collection<RSAPublicKey> publicKeys) {
+
+        boolean signatureValid = publicKeys.stream()
+            .filter(k -> {
+                JWSVerifier verifier = new RSASSAVerifier(k);
+                try {
+                    return signedJWT.verify(verifier);
+                } catch (JOSEException e) {
+                    throw new RuntimeException(e);
```
jlorper, line 161: you may have 2+ keys, right? while rotating, I wouldn't err here, just in case the first one doesn't work, should try with next one

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/output/CompressedOutputWrapper.java
```diff
@@ -1,43 +1,55 @@
...
 
+    @SneakyThrows
     @Override
-    public void write(HttpEventRequest request, ProcessedContent content) throws IOException {
+    public void write(String key, ProcessedContent content) {
 
-        byte[] compressedContent = gzipContent(content.getContent());
+        if (!Objects.equals(COMPRESSION_TYPE, content.getContentEncoding())) {
+            byte[] compressedContent = gzipContent(content.getContent());
+            content = content.withContentEncoding("gzip").withContent(compressedContent);
```
jlorper, line 39: Use the constant here too? COMPRESSION_TYPE

java/impl/aws/src/main/java/co/worklytics/psoxy/WebhookCollectionModeHandler.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.Optional;
+import java.util.stream.Stream;
+import co.worklytics.psoxy.aws.DaggerAwsContainer;
+
+/**
+ * AWS lambda entrypoint that can handle BOTH:
+ *    - webhook requests from API Gateway / Function URL invocations, AND
+ *    - SQS events (presumed to be batches of webhooks)
+ */
+public class WebhookCollectionModeHandler implements RequestStreamHandler {
```
jlorper, line 35: Rename then to `AWSWebhookCollectionModeHandler` or move it under `co.worklytics.psoxy.aws`

java/impl/aws/src/main/java/co/worklytics/psoxy/WebhookCollectionModeHandler.java
```diff
@@ -0,0 +1,174 @@
...
+        sqsPayloadMapper.configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true);
+    }
+
+
+    @Override
+    public void handleRequest(InputStream input, OutputStream output, Context context) throws IOException {
+        // Read the full input stream into a tree
+        JsonNode rootNode = mapper.readTree(input);
+
+        if (isSQSEvent(rootNode)) {
```
jlorper, line 70: Given that is going to be held in memory in the form of a JsonNode I wouldn't bother then.
I'd do:

```
final byte[] bytes = inputStream.readAllBytes();
Optional<SQSEvent> sqsEvent = readSQSEvent(bytes);
// which is the following wrapper with exception handling
objectMapper.readValue(bytes, SQSEvent.class);

if (sqsEvent.isPresent()) { ... sqs ... }
// same thing for other types, easily you can handle API Gateway V1 too this way w/o getting into JSON schema details
```

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/SQSOutput.java
```diff
@@ -0,0 +1,70 @@
...
+    @Inject
+    Provider<SqsClient> sqsClient;
+
+    @Override
+    public void write(ProcessedContent content) {
+        write(null, content);
+    }
+
+    @Override
+    public void write(@Nullable String key, ProcessedContent content) {
```
jlorper, line 45: key ignored - comment why?

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/InboundWebhookHandler.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @SneakyThrows
+    public HttpEventResponse handle(HttpEventRequest request) {
+        Optional<String> authorizationHeader = getAuthorizationHeader(request);
+
+
+        Optional<SignedJWT> authToken;
+
+        if (authorizationHeader.isEmpty()) {
+            boolean authHeaderRequired = configService.getConfigPropertyAsOptional(WebhookCollectorModeConfigProperty.REQUIRE_AUTHORIZATION_HEADER).map(Boolean::parseBoolean).orElse(true);
```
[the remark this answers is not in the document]
eschultink: ? it's optional, defaulting to `true`

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/InboundWebhookHandler.java
eschultink: may have multiple, depends on approach. I thought returns `false` when signature can't be verified, but yeah - I agree that docs unclear about when `JOSEException` is thrown, so we can treat as equivalent to verification returning `false`.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/output/LogsOutput.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * not sure there's any *real* use for this, but perhaps
+ */
+@Log(topic = "output")
+public class LogsOutput implements Output {
+
+
+    @Override
+    public void write(ProcessedContent content) {
+        log.log(Level.INFO, content.toString());
```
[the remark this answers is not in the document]
eschultink: well, really don't see why webhooks should ever be put in logs as 'WARNING' or something

java/impl/aws/src/main/java/co/worklytics/psoxy/WebhookCollectionModeHandler.java
eschultink: yeah, OK - it's using exceptions to control execution flow, but alright.  Agree it's more mem efficient.

PR #911, opened 2025-06-17
webhook collector CORS

eschultink:
### Fixes
 - add CORS support to webhook collectors, so those can work from browser-based JS

Big doubt is whether to allow configuration of allowed origins ...

Curiously, while I added underneath in the handler (app-level), AWS seems to handle these OPTIONS requests at API Gateway/Function URL level, via settings; so I've done them via terraform.

### Change implications

---
  - https://app.asana.com/0/0/1210561492262910

### DIFF
--- a/infra/modules/aws-webhook-collector/main.tf
+++ b/infra/modules/aws-webhook-collector/main.tf
@@ -27,7 +27,7 @@ locals {
   # helper to clarify conditionals throughout
   use_api_gateway = var.api_gateway_v2 != null
 
-  authorization_type = "NONE"                                # TODO: make this configurable, if auth BEYOND function-level checks of Authorization header is desired
+  authorization_type = "IAM"
   http_methods       = var.http_methods                      # Use http_methods directly without adding OPTIONS
 }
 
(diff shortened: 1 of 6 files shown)


infra/modules/aws-webhook-collector/main.tf
```diff
@@ -122,6 +122,16 @@ resource "aws_lambda_function_url" "lambda_url" {
   function_name      = module.gate_instance.function_name
   authorization_type = local.authorization_type
 
+  # NOTE: CORS doesn't require that we include 'OPTIONS' method, as it is added automatically
+  # message is You can use GET, PUT, POST, DELETE, HEAD, PATCH, and the wildcard character (*).",
+  cors {
+    allow_credentials = true
+    allow_headers     = ["*"]
+    allow_methods     = [for m in local.http_methods : m if !(m == "OPTIONS")]
+    allow_origins     = ["*"] # q: support configuration of this??
```
jlorper, line 131: yes, they should add the origins of their local apps

davidfq (approved):

PR #925, opened 2025-07-01
S202 : side output error feedback

eschultink:
### Features
 - return catchable exception if output write fail
 - catch that exception in webhook, side output use cases to give more meaningful feed back to client ( fatal error, non-fatal warning respectively)

### Change implications


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -329,12 +340,9 @@ ProcessedContent sanitize(HttpEventRequest request, RequestUrls requestUrls, Pro
     }
 
 
-    void writeSideOutput(HttpEventRequest request, com.google.api.client.http.HttpResponse sourceApiResponse, ProcessedContent sanitizedContent) {
-        try {
+    void writeSideOutput(HttpEventRequest request, com.google.api.client.http.HttpResponse sourceApiResponse,
+                         ProcessedContent sanitizedContent) throws Output.WriteFailure {
             apiDataSideOutputSanitized.write(request, sanitizedContent);
```
jlorper, line 345: If this is now a one-liner I would just inline the call. It took me a while to realize that the output involved was `apiDataSideOutputSanitized` . Also the response is not used

jlorper (approved):

PR #927, opened 2025-07-02
async processing of API data requests 

eschultink:
### Description

creates option for API Data request to be handled async by the proxy.  Assumes side output is configured, as otherwise there's not point.

Implementation is complete ONLY for AWS; will require SQS queue, equivalent to how we're doing webhook collection.

Data flow is:

1. client sends request to proxy, with header to flag for async
2. proxy parses requests, validates, enqueues to SQS and responds with 202 Accepted.
3. SQS triggers another request (invocation) to proxy, which this time does the actually API request, sanitized response, writes to outputs
4. client can not pick response object(s) up from outputs

Remaining terraform tasks:
  - modify `api-data-connection` module to create SQS queue if `supportAsync` flag or something is enabled.
  - modify `api-data-connection` module to set the hybrid handler in this case, which can deal with SQS or apigateway/cloud function url invocations
  - PoC for slack-analytics use-case

### Change implications

- **Dependencies added/changed?**  NO; will depend on SQS, like webhook case.
- **Something important to note in future release notes?** No
- **Breaking changes?** No, but care should be taken with new handler**


java/impl/aws/src/main/java/co/worklytics/psoxy/AwsApiDataModeHybridHandler.java
```diff
@@ -0,0 +1,133 @@
...
+import java.io.OutputStream;
+
+/**
+ * AWS lambda entrypoint that can handle API data requests via any of:
+ *   - cloud function URL invocations
+ *   - API Gateway events
+ *   - SQS events ( asynchronous processing of API data requests, that originally came from one of the above)
+ */
+@Log
+public class AwsApiDataModeHybridHandler implements RequestStreamHandler {
```
eschultink, line 30: new hybrid handler that will be entry point for API data connections, if they support async processing

java/core/src/main/java/co/worklytics/psoxy/gateway/output/ApiDataOutputUtils.java
```diff
@@ -0,0 +1,267 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.*;
+import java.util.stream.Collectors;
+
+@AllArgsConstructor(onConstructor_ = {@Inject})
+public class ApiDataOutputUtils {
```
eschultink, line 24: eh, split this stuff out from `OutputUtils`, as all quite specific to "API Data" use-case

java/core/src/main/java/co/worklytics/psoxy/ControlHeader.java
```diff
@@ -19,16 +19,23 @@ public enum ControlHeader {
...
     PSEUDONYM_IMPLEMENTATION("Pseudonym-Implementation"),
 
     /**
-     * this header - sent with any value - means the request is a health check, not actually
-     * intended to be forwarded to source
+     * client does NOT want a response; proxy should ACK request with 204 No Content, continue API call
+     * in background, then write response to side output (assuming it's enabled)
      */
-    HEALTH_CHECK("Health-Check"),
+    PROCESS_ASYNC("Process-Async"),
```
eschultink, line 38: flag to be used if client wants the API data request to be processed async.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/output/OutputUtils.java
```diff
@@ -27,20 +21,6 @@
 @Singleton
 public class OutputUtils {
```
eschultink, line 22: 🤔  this is almost more of a factory now; all the metadata stuff moved out, bc it's pretty generic.

java/core/src/main/java/co/worklytics/psoxy/gateway/output/ApiDataOutputUtils.java
```diff
@@ -0,0 +1,267 @@
...
+            .contentCharset(sourceApiResponse.getContentCharset());
+
+        try (InputStream stream = sourceApiResponse.getContent()) {
+            builder.content(stream.readAllBytes());
+        }
+
+        Map<String, String> metadata = this.buildRawMetadata(sourceApiRequest);
+
+
+        //not sure this will work; are we certain to be able to consume HttpContent after request has been sent?
```
aperez-worklytics, line 130: I think not, we'll need to *store* the content in some kind of buffered stream to be reused, otherwise the 2nd time content is read it will fail

java/core/src/main/java/co/worklytics/psoxy/gateway/output/ApiDataOutputUtils.java
```diff
@@ -0,0 +1,267 @@
...
+        ;
+    }
+
+    Provider<UUID> uuidProvider;
+    Base64.Encoder base64encoder;
+
+
+    public String buildRawOutputKey(HttpRequest requestToSourceApi) {
+        Map<String, String> metadata = buildRawMetadata(requestToSourceApi);
+        return buildRawOutputKey(metadata);
```
aperez-worklytics, line 64: I'd put a limit on keys for 1024 characters. We can do 992 characters or regular request value and it exceeds, put the rest into MD5 hashed (32 more characters)

java/impl/aws/src/main/java/co/worklytics/psoxy/AwsApiDataModeHybridHandler.java
```diff
@@ -0,0 +1,133 @@
...
+
+        // whatever the event is, convert to HttpEventRequest and invoke the request handler
+        if (lambdaEventUtils.isSQSEvent(rootNode)) {
+
+            JsonNode records = rootNode.get("Records");
+            if (records == null || records.isEmpty()) {
+                throw new IllegalArgumentException("SQS event has no records");
+            }
+            if (records.size() > 1) {
+                // well, tbh - why have this limitation? we certainly could batch in future if there's a use case
```
aperez-worklytics, line 73: agree

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -266,20 +311,25 @@ public HttpEventResponse handle(HttpEventRequest request) {
             // return response
             builder.statusCode(sourceApiResponse.getStatusCode());
 
+            // TODO: in slack-analytics use-case, this is NDJSON
```
aperez-worklytics, line 314: In addition to that, is a ndjson compressed with `application/gzip`, so we'll need need to uncompressed first

java/impl/aws/src/main/java/co/worklytics/psoxy/AwsApiDataModeHybridHandler.java
eschultink: that was initially specific to the webhook case; this handler is for the API data connector case. 

but I've refactored to handle arbitrary number of records.

java/core/src/main/java/co/worklytics/psoxy/gateway/output/ApiDataOutputUtils.java
eschultink: i've restructured this so it's just `yyyyMMdd/{uuid}`.  makes it easy to inspect that things are "working" by looking in the bucket, still concise.  Avoids need to worry about sanitization of the path.  Also don't need to worry about unuseful repetition of host + path components, which aren't adding any value.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -267,20 +309,18 @@ public HttpEventResponse handle(HttpEventRequest request) {
             builder.statusCode(sourceApiResponse.getStatusCode());
 
             // TODO: if side output cases of the original, we *could* use the potentially compressed stream directly, instead of reading to a string?
-            ProcessedContent original = this.asProcessedContent(sourceApiResponse);
-            apiDataSideOutput.write(request, original);
+            ProcessedContent original = apiDataOutputUtils.responseAsRawProcessedContent(requestToSourceApi, sourceApiResponse);
+            apiDataSideOutput.writeRaw(original, processingContext);
```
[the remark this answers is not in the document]
eschultink: it's a config property. use-case for this is customers who want copies of the original data, because it's a rate-limited API that they've also been using elsewhere. Side-output of original can be used to fulfill that use-case, build in effective data-lake of everything.

PR #935, opened 2025-07-17
improve gmail examples

eschultink:
### Features
 - URL_SAFE_ENCODING by default for gmail
 - an example of `messages` endpoint, even though not really any transform applicable

### Change implications


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestBuilder.java
```diff
@@ -9,6 +9,7 @@
 import com.google.api.client.json.webtoken.JsonWebSignature;
 import com.google.api.client.json.webtoken.JsonWebToken;
 import com.google.common.annotations.VisibleForTesting;
+import io.opencensus.trace.Link;
```
aperez-worklytics, line 12: used?

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -331,6 +344,404 @@ private Object sanitize(Endpoint endpoint, Object jsonResponse) {
     }
 
 
+    private static boolean transformApplies(Transform transform, Object document) {
```
aperez-worklytics, line 347: All of these new code is from `SanitizerUtils`; so not sure if want to move it or it was a bad paste -but code is still there

aperez-worklytics (changes requested):
Some code duplicated; missing commits?

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/ClientCredentialsGrantTokenRequestBuilder.java
eschultink: ah, no; i think came from newrelic branch where this was originally implemented.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
eschultink: yeah, you're correct; I think combination of merge + my not having fully split out all the tests. Done now.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/transforms/Transform.java
```diff
@@ -217,6 +217,13 @@ public FilterTokenByRegex clone() {
     @Getter
     public static class PseudonymizeEmailHeader extends Transform {
 
+        /**
+         * how to encode to the resulting pseudonym
+         */
+        @JsonInclude(JsonInclude.Include.NON_DEFAULT) //doesn't work for enums ...
+        @Builder.Default
+        PseudonymEncoder.Implementations encoding = PseudonymEncoder.Implementations.JSON;
```
jlorper, line 225: I guess by default should be URL_SAFE_TOKEN

jlorper (approved):
A couple of suggestions

java/gateway-core/src/main/java/com/avaulta/gateway/rules/transforms/Transform.java
eschultink: yeah, technically breaking change to behavior - but might as well. no one but us should be dependent on the format.

PR #936, opened 2025-07-17
accept non-ASCII-printable chars in CSV headers

eschultink:
### Features
  - prev, we err'd on non-ASCII-printables. we've seen a case of these in the wild, so going to warn instead

### Change implications

---
  - https://app.asana.com/0/0/1210825949508132


jlorper (approved):
OK. The problem with this is that it translates the issue to our processing and we need to check for weird headers on Worklytics side.
Maybe we could leverage the column renaming feature (`columnsToRename`), to at least contain that to proxy, so "USER**[NBSP]**NAME" -> "USER NAME" . Suggest that in the error log too?

PR #947, opened 2025-07-28
Avoid adding body when is blank/empty/null and mimetypes

aperez-worklytics:
Seen for GCP; happening since v0.5.4.

When body is converted from `CloudFunctionRequest`, is an empty array instead of null. That is causing to consider that the request has a body and the client is failing if the request is a GET and it has a body:

```
java.lang.IllegalArgumentException: GET with non-zero content length is not supported
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:167)
	at com.google.api.client.util.Preconditions.checkArgument(Preconditions.java:67)
...
```

In other hand, when a POST is being done it was always failing due mismatch on check the content type.

### Fixes
[GCP: Empty body is included when doing GET requests](https://app.asana.com/1/27145998307022/project/1210814744431401/task/1210903872034896)
[Content type is not recognized](https://app.asana.com/1/27145998307022/project/1210814744431401/task/1210916133460397)

 - []()

### Change implications

     obviously a no-op? **no**
     change **no**

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
@@ -876,14 +876,14 @@ public class ApiDataRequestHandler {
     @SneakyThrows
     ByteArrayContent reverseRequestBodyTokenization(@NonNull String contentType, String body) {
         // JSON case: use ObjectMapper to parse request body and map decode ON every string value
-        if (contentType.equals(ContentType.APPLICATION_JSON.getMimeType())) {
+        if (contentType.contains(ContentType.APPLICATION_JSON.getMimeType())) {
             JsonNode jsonNode = objectMapper.readTree(body);
             JsonNode transformedNode = applyStringTransformToTree(jsonNode, this::decode);
             String decodedRequestBody = objectMapper.writeValueAsString(transformedNode);
 
             return new ByteArrayContent(contentType,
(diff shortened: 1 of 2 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -876,17 +876,16 @@ private boolean isUtf8CompatibleEncoding(String encoding) {
     @SneakyThrows
     ByteArrayContent reverseRequestBodyTokenization(@NonNull String contentType, String body) {
         // JSON case: use ObjectMapper to parse request body and map decode ON every string value
-        if (contentType.equals(ContentType.APPLICATION_JSON.toString())) {
```
aperez-worklytics, line 879: This is returning `application/json; charset=UTF-8`; needed to use `getMimeType()` to get the right value on the enum object

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -876,17 +876,16 @@ private boolean isUtf8CompatibleEncoding(String encoding) {
...
+        if (contentType.equals(ContentType.APPLICATION_JSON.getMimeType())) {
```
eschultink, line 879: should we be more relaxed in both cases? use `contains()` instead of `equals()`?

eschultink (approved):

PR #949, opened 2025-07-31
ChaGPT Enterprise via Compliance API support

aperez-worklytics:
Support for ChatGPT Compliance API:
- Adding rules on YAML, generated from model https://github.com/Worklytics/evalengin/pull/7740
- Adding a basic test to test the endpoints and response sanitization
- Adding connector spec
- Adding basic instructions and readme

 - []()

### Features
[Psoxy rules](https://app.asana.com/1/27145998307022/task/1210928401031887)

### Change implications

     obviously a no-op?
     change

---
  - https://app.asana.com/0/0/1210928401031887


infra/examples-dev/aws/misc-data-source-variables.tf
```diff
@@ -2,6 +2,12 @@
 # (not coupled to hosting environment; so split into separate file to ease keeping them in sync
 #  across AWS/GCP examples.  DRY!!)
 
+variable "chat_gpt_compliance_example_workspace_id" {
+  type        = string
+  description = "Workspace id to se for example calls"
```
jlorper, line 7:
```suggestion
  description = "Workspace id to set for example calls"
```

docs/sources/chatgpt/README.md
```diff
@@ -0,0 +1,21 @@
+# ChatGPT
+
+## **ALPHA** Compliance API
```
eschultink, line 3: `Enterprise via Compliance API`

also add caveats that 1) this may not work for all ChatGPT subscriptions (at min, requires Enterprise), and 2) may be be subject to rate limits that are too low to transfer all relevant data for your organization.

eschultink (approved):
Looks good.

if we want connector id to be `chatgpt-enterprise-compliance` - OK. but as far as in docs/etc, i think call it "ChatGPT Enterprise" everywhere.

Compliance API is the mechanism, but it will only work for "Enterprise" subs - so we need to emphasize everywhere that it's only for enterprise. And have caveats about the rate limits.

aperez-worklytics (commented):
Ok, I've renamed it @eschultink

PR #952, opened 2025-08-04
S205 : align examples

eschultink:
### Features
  - GCP example variable schema for webhook slightly different than AWS; align to AWS

### Change implications

### DIFF
--- a/infra/examples-dev/gcp/main.tf
+++ b/infra/examples-dev/gcp/main.tf
@@ -98,21 +98,26 @@ module "psoxy" {
   secret_replica_locations          = var.secret_replica_locations
   api_connectors                    = local.api_connectors
   bulk_connectors                   = local.bulk_connectors
-  webhook_collectors                = var.webhook_collectors
-  non_production_connectors         = var.non_production_connectors
-  custom_api_connector_rules        = var.custom_api_connector_rules
-  general_environment_variables     = var.general_environment_variables
-  pseudonymize_app_ids              = var.pseudonymize_app_ids
-  email_canonicalization            = var.email_canonicalization
-  bulk_input_expiration_days        = var.bulk_input_expiration_days
-  bulk_sanitized_expiration_days    = var.bulk_sanitized_expiration_days
-  custom_bulk_connector_rules       = var.custom_bulk_connector_rules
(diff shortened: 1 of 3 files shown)


infra/examples-dev/gcp/main.tf
```diff
@@ -98,21 +98,26 @@ module "psoxy" {
...
-  custom_artifacts_bucket_name      = var.custom_artifacts_bucket_name
-  custom_side_outputs               = var.custom_side_outputs
-  todos_as_local_files              = var.todos_as_local_files
-  todo_step                         = local.max_auth_todo_step
+  webhook_collectors = { for k, v in var.webhook_collectors : k => merge(
+    v,
+    {
+      example_payload = file(v.example_payload_file)
+    }
+  ) }
```
jlorper, line 106: example_payload_file is optional. Does "file" play nicely with nulls? Maybe you need a filter or a `coalesce(try(v.example_payload_file), null)`

```suggestion
  webhook_collectors = { for k, v in var.webhook_collectors if v.example_payload_file != null: k => merge(
    v,
    {
      example_payload = file(v.example_payload_file)
    }
  ) }
```

infra/modules/gcp-host/main.tf
```diff
@@ -271,8 +271,8 @@ module "webhook_collector" {
   rules_file                         = each.value.rules_file
   webhook_batch_invoker_sa_email     = module.psoxy.webhook_batch_invoker_sa_email
   batch_processing_frequency_minutes = try(each.value.batch_processing_frequency_minutes, 5)
-  example_identity                   = each.value.example_identity
-  example_payload                    = each.value.example_payload
+  example_identity                   = try(each.value.example_identity, null)
+  example_payload                    = try(each.value.example_payload, null)
```
jlorper, line 275: is this try right here? should it be coalesce?

jlorper (approved):
A couple of doubts, I'm bit rusty on terraform, so not sure if accurate.
Approving so you can move on if needed

infra/examples-dev/gcp/main.tf
eschultink: coalesce will fail if all null, but yeah, should be try with blank as fallback I think.

PR #966, opened 2025-08-22
GCP fixes

eschultink:
### Fixes
 - some config properties for async support break the usual case if missing

### Features
 - less generic name for Pub/Sub queue

### Change implications

### DIFF
--- a/java/impl/gcp/src/main/java/co/worklytics/psoxy/GcpApiDataRequestHandler.java
+++ b/java/impl/gcp/src/main/java/co/worklytics/psoxy/GcpApiDataRequestHandler.java
@@ -74,7 +74,7 @@ public class GcpApiDataRequestHandler {
     @SneakyThrows
     public void service(HttpRequest request, HttpResponse response) {
 
-        boolean pubsubTriggerConfigured = apiModeConfig.get().getPubSubTopic() != null;
+        boolean pubsubTriggerConfigured = apiModeConfig.get().getAsyncPubSubQueue() != null;
 
         // if a pubsub trigger is configured, check if request is invocation via PubSub
         // as of 2025-08-19, User-Agent appears as 'APIs-Google; (+https://developers.google.com/webmasters/APIs-Google.html)'
@@ -162,7 +162,10 @@ public class GcpApiDataRequestHandler {
             log.log(Level.WARNING, "Unauthorized : no authorization header included");
             throw new GcpWebhookCollectionHandler.AuthorizationException("Unauthorized : no authorization header included");
         }
-        GoogleIdTokenVerifier internalServiceVerifier = googleIdTokenVerifierFactory.getVerifierForAudience(apiModeConfig.get().getServiceUrl());
+        GoogleIdTokenVerifier internalServiceVerifier = 
+            apiModeConfig.get().getServiceUrl()
+                .map(googleIdTokenVerifierFactory::getVerifierForAudience)
+                .orElseThrow(() -> new IllegalStateException("Service URL not configured"));
 
         GoogleIdToken idToken = null;
         try {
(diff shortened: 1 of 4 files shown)


java/impl/gcp/src/main/java/co/worklytics/psoxy/GcpApiDataRequestHandler.java
```diff
@@ -74,7 +74,7 @@ public GcpApiDataRequestHandler(
     @SneakyThrows
     public void service(HttpRequest request, HttpResponse response) {
 
-        boolean pubsubTriggerConfigured = apiModeConfig.get().getPubSubTopic() != null;
+        boolean pubsubTriggerConfigured = apiModeConfig.get().getAsyncPubSubQueue() != null;
```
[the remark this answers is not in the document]
aperez-worklytics: agree on this @eschultink

java/core/src/main/java/co/worklytics/psoxy/gateway/output/ApiDataOutputUtils.java
```diff
@@ -112,8 +115,11 @@ public ProcessedContent responseAsRawProcessedContent(HttpRequest sourceApiReque
             .contentType(sourceApiResponse.getContentType())
             .contentCharset(sourceApiResponse.getContentCharset());
 
-        try (InputStream stream = sourceApiResponse.getContent()) {
-            builder.content(stream.readAllBytes());
+        // sourceApiResponse will not have content for 'HEAD' requests (or potentially even GET, in some cases - 204 No Content, 304 Not Modified, etc.)
+        if (sourceApiResponse.getContent() != null) {
```
aperez-worklytics, line 119: OK;
I think is missing to support 304 as a successful response:

https://github.com/Worklytics/psoxy/blob/926bdf16be39a6b128dc6fa613ca66569dd39d70/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java#L698

As currently in this case, handler will return an error due response is < 300

aperez-worklytics (changes requested):

java/core/src/main/java/co/worklytics/psoxy/gateway/output/ApiDataOutputUtils.java
eschultink: the 304 this in in general; the `isSuccessFamily` thing - we have to look a how that is used. For our use-case, I think correct.

the better question is that if we're using `Get-If-Modified-Since`, which could result in the 304 - then we should consider if we want some other method than `isSuccessFamily`.

afaik, proxy passes non-200s back to caller anyways.

jlorper (approved):

java/core/src/main/java/co/worklytics/psoxy/gateway/output/ApiDataOutputUtils.java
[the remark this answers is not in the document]
eschultink: no.

it's not about the request or the content. This is taking the content from the response, if any.

aperez-worklytics: Yes, but when  *non success* it includes the api error header in the response:

https://github.com/Worklytics/psoxy/blob/926bdf16be39a6b128dc6fa613ca66569dd39d70/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java#L480

We are not currently using 304 (it was a test that I was doing yesterday for a thing that we are not going to need ATM)  and its headers (like `Get-If-Modified-Since`) will currently need to be part of the rules as by default they are omitted when doing the call

PR #973, opened 2025-09-03
Confluence support

aperez-worklytics:
- Support for Confluence Cloud
- Updated docs
- Updated README

 - []()

### Features
[Psoxy connector](https://app.asana.com/1/27145998307022/task/1211148487418666)

### Change implications

     obviously a no-op? 
     change


docs/sources/atlassian/confluence/confluence.yaml
```diff
@@ -0,0 +1,653 @@
+---
+endpoints:
+  - pathTemplate: "/ex/confluence/{cloudId}/wiki/api/{version}/attachments/{pageId}/versions"
```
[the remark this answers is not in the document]
eschultink: I don't feel strongly. hardcoding the explicit version we want is most correct, but I agree with your point.

args against parameterizing version would be:
  - major future version bump could very well require additional changes too, while they're likely to leave major versions live for a long time. so not sure making the rules "upgrade-proof" is super helpful. 
  - potential changes from version to version could alter fields in manner that exposes PII/sensitive fields, although very unlikely given that we're enforcing expected schema. 
 
 
So yeah - OK to parameterize, but don't think we need to make that the consistent practice always.  Let's leave parameterized for now, just in hopes that if there is major version bump at some point, these core endpoints have some probability of not being changed in a way that necessarily breaks things for us.

docs/sources/README.md
```diff
@@ -12,6 +12,7 @@ To add a source, add its Connector ID to the `enabled_connectors` list in your `
 | `azure-ad`                 | Azure Active Directory              | API  | DEPRECATED   |
 | `badge`                    | Badge                               | Bulk | GA           |
 | `chatgpt`                  | ChatGPT                     | API  | BETA         |
+| `confluence`               | Confluence Cloud                    | API  | BETA         |
```
eschultink, line 15: also, make sure you add to `SUMMARY.md`, so shows up in links in gitbook

and symlinke from `docs/sources/` --> `confluence/`, as we have for other sources.

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -801,6 +802,195 @@ all the operations for the connector:
...
+        REFRESH_ENDPOINT : "https://auth.atlassian.com/oauth/token"
+        USE_SHARED_TOKEN : "TRUE"
+      }
+      settings_to_provide = {
+      }
+      reserved_concurrent_executions : null
+      enable_side_output : false
+      example_api_calls_user_to_impersonate : null
+      example_api_calls : [
+        "/oauth/token/accessible-resources", # obtain Confluence Cloud ID from here
```
eschultink, line 852: OK, but possibly not helpful to list first, just bc it's going to be uninteresting test call that is going to be used by default

docs/sources/atlassian/confluence/confluence.yaml
```diff
@@ -0,0 +1,653 @@
...
+      - !<pseudonymize>
+        jsonPaths:
+          - "$.results[*].accountId"
+          - "$.results[*].email"
+        encoding: "URL_SAFE_TOKEN"
+      - !<redact>
+        jsonPaths:
+          - "$.results[*].displayName"
+          - "$.results[*].profilePicture"
+          - "$.results[*].publicName"
```
eschultink, line 362: in theory, we shouldn't need these explicit redactions here - redundant with json-schema filter.

infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -801,6 +802,195 @@ all the operations for the connector:
...
+        "/ex/confluence/${local.confluence_example_cloud_id}/wiki/rest/api/group",
+        "/ex/confluence/${local.confluence_example_cloud_id}/wiki/rest/api/content/search?cql=lastmodified>=${formatdate("YYYY-MM-DD", timeadd(var.example_api_calls_sample_date, "-720h"))}%20AND%20lastmodified<=${formatdate("YYYY-MM-DD", var.example_api_calls_sample_date)}&limit=30&expand=body.atlas_doc_format,ancestors,version,history,history.previousVersion&includeArchivedSpaces=true",
+        "/ex/confluence/${local.confluence_example_cloud_id}/wiki/api/v2/spaces",
+        "/ex/confluence/${local.confluence_example_cloud_id}/wiki/api/v2/attachments/{attachmentId}/versions",
+        "/ex/confluence/${local.confluence_example_cloud_id}/wiki/api/v2/blogposts/{blogpostId}/versions",
+        "/ex/confluence/${local.confluence_example_cloud_id}/wiki/api/v2/pages/{pageId}/versions",
+        "/ex/confluence/${local.confluence_example_cloud_id}/wiki/api/v2/footer-comments/{commentId}/versions",
+        "/ex/confluence/${local.confluence_example_cloud_id}/wiki/api/v2/inline-comments/{commentId}/versions",
+        "/ex/confluence/${local.confluence_example_cloud_id}/wiki/api/v2/tasks",
```
eschultink, line 861: i think if we could add the `groupMembersById` case, that would be good .. it's arg the most important one, and perhaps is the ONLY thing that's going to confirm that `users` scope has been granted?? (eg, none of the above dependent on that??)

eschultink (approved):
a few thoughts/suggestions - no blockers

aperez-worklytics (commented):
Thanks. Some changes done @eschultink 
- Revisit rules: minor changes in pagination and properties that were wrongly mapped
- Updated permissions: drop couple of ones aren't required
- TextDigest for all "doc" object
- Update docs, with links to permission-endpoint to check why we have them.

PR #974, opened 2025-09-03
Stream-based sanitizer interface

eschultink:
wip draft for sake of awareness.

provides little value except in NDJSON response cases, which to date are very few cases (slack analytics API, maybe glean audit log in future). in regular JSON cases, we're going to parse entire response into mem anyways.

### Features
- work towards a fully Stream-based sanitizer interface; in theory, this will improve mem issues and avoid possible OOM in some cases.

### Change implications

 - breaking changes? **no**

---
  - https://app.asana.com/0/0/1210711380911437


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/output/OutputUtils.java
```diff
@@ -105,29 +92,6 @@ OutputLocation locationForWebhookQueue() {
                         "No side output configured for webhook queue"));
     }
 
-    public ProcessedContent decompressIfNeeded(ProcessedContent original) throws IOException {
-        if (Objects.equals(original.getContentType(), "application/gzip")) {
```
eschultink, line 109: so this check covers the Slack Analytics API case, but only that. 

my confusion is how we're handling `gzip` in other cases. best guess is there must be some *other* layer at which we're decompressing. Either we do it explicitly somewhere else, or Google HTTP client is automatically seeing the `Content-Encoding` header and dealing with it.  A concern is be that if indeed Google HTTP client is decompressing gzip but not stripping the Content-Encoding header, we're going to double-decompress the content

PR #984, opened 2025-09-10
options for custom API connectors, API connectors from templates

eschultink:
### Features
  - support creating custom API connectors
  - support creating additional connectors using specs as templates
  - support passing a `rules_file` attribute to aws-host/gcp-host api connector specs

### TODO
 - [ ] some connectors require variables that may need to vary by instance (eg, `salesforce_domain`); solve this?

### Change Implications

- **Dependencies added/changed?** **no**
- **Something important to note in future release notes?** **no**
- **Breaking changes?** **no**


eschultink:
actually, I'm going to split this in two.

PR #985, opened 2025-09-10
Support embedded JSON on pseudonymize

aperez-worklytics:
Created as draft, to discuss.

Supporting pseudonymize embedded JSON instead of the full string from a field. 

Case is for #973, for getting mentions pseudonymization (which are part of an embedded json payload)

 - []()

 - []()

### Change implications

     obviously a no-op?
     change


java/gateway-core/src/main/java/com/avaulta/gateway/rules/transforms/Transform.java
```diff
@@ -274,6 +274,14 @@ public static class Pseudonymize extends Transform implements PseudonymizationTr
         @Builder.Default
         Boolean includeReversible = false;
 
+         @JsonInclude(JsonInclude.Include.NON_DEFAULT)
+         @Builder.Default
+         Boolean isJsonEscaped = false;
+
+         @JsonInclude(JsonInclude.Include.NON_DEFAULT)
+         @Builder.Default
+         String jsonPathToProcessWhenEscaped = null;
```
eschultink, line 283: `jsonPathWithinValue`?

```
/**
  * if provided, value will be parsed as JSON, and the result of evaluating the path will be pseudonymized
  */
```

eschultink:
what's the use-case?

I have some doubts about making it specific to pseudonymize. seems generally applicable: interpret the value as JSON, the apply transform (pseudonymize) to value matching the json path.

So rather than a special case of pseudonymize, is this a `parseFromJsonAndTransform` which takes a transform to apply on the json value it applies to. 


```yaml
    transforms:
      - !<parseFromJsonAndTransform>
        jsonPaths:
          - "$..avatar_url"
          - "$..gravatar_id"
        transforms
          - !<pseudonymize>
            jsonPaths:
              - "$.embeddeField"
      - !<pseudonymize>
        jsonPaths:
          - "$..id"
          - "$..node_id"
          - "$..email"
        encoding: "JSON"
```

PR #986, opened 2025-09-10
support custom api connectors

eschultink:
### Features
 - variable in examples to add custom_api_connectors
 - support passing a rules_file in as part of connector specs

value of this is:
  1. avoid needing to modify customer's `main.tf` to add custom api connectors
  2. not splitting file from the connector spec when we do so

### Change implications

 - breaking changes? **no**


jlorper (approved):
when defining the actual block we may use terraform merge, so no need to copy over everything (not sure if it would work), something like (pseudocode):

```
custom_api_connectors = {
merge( "salesforce-spec", { "name" : "salesforce2", ... }),
merge( "salesforce-spec", { "name" : "salesforce3", ... }),
}
```

PR #987, opened 2025-09-10
support adding multiple connectors to same source_kind, using templates

eschultink:
### Features
 - support multiple connectors of same source kind, leveraging the specs as "templates"
 
## TODO:
 - [ ] solution for cases such as `salesforce_domain`, which if you want mutliple salesforce instances, you'd need to pass in multiple times to create the "templates"; as it stands, multiple salesforce connector usages via templates will end up with the same salesforce domain. 

### Change implications

### DIFF
--- a/infra/examples-dev/aws/variables.tf
+++ b/infra/examples-dev/aws/variables.tf
@@ -219,6 +219,14 @@ variable "enabled_connectors" {
   description = "list of ids of connectors to enabled; see modules/worklytics-connector-specs"
 }
 
+variable "enabled_connectors_from_templates" {
+  type = map(object({
+    template_id = string
+  }))
+  description = "map of connector id --> id of connector template to enable"
+  default     = {}
+}
+
 variable "non_production_connectors" {
   type        = list(string)
   description = "connector ids in this list will be in development mode (not for production use)"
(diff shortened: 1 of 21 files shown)


infra/examples-dev/aws/variables.tf
```diff
@@ -219,6 +219,14 @@ variable "enabled_connectors" {
   description = "list of ids of connectors to enabled; see modules/worklytics-connector-specs"
 }
 
+variable "enabled_connectors_from_templates" {
+  type = map(object({
+    template_id = string
+  }))
```
eschultink, line 225: 1. alter connector "specs" to be more like true templates, and have each evaluated in isolation with a "scope" that's the content of this object??
  - problem with that is that I believe Terraform erases/hides object properties not explicitly in variables definition of the module; so we can't have arbitrary content.
  - might also make module structure pretty complicated, with connectors being invoked repeatedly
2. have a map(string) that is "merged" over top whatever the spec for the template has by default?
    - not sure this is enough
    - undermines value, bc in several cases you won't re-use that much of the template and will have to re-write a bunch of it.
3. explicit set of properties that can be overridden; `target_host`, `environment_variables`, `settings_to_provide`, and `rules_file` probably enough ... 
     - `example_api_requests` will remain problematic.

PR #997, opened 2025-09-17
fix source api error logging

eschultink:
### Fixes
- bc type now `byte[]`, the implicit toString() means it's not readable in logs

### Change implications


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -56,39 +88,6 @@
...
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
-import java.util.UUID;
-import java.util.function.Function;
-import java.util.function.Supplier;
-import java.util.logging.Level;
-import java.util.regex.Pattern;
-import java.util.stream.Collectors;
```
eschultink, line 91: ugh - not sure why cursor likes to organize imports so aggressively

PR #1004, opened 2025-09-22
better encryption key rotation feedback

eschultink:
### Features
 - return meaningful error code when token couldn't be reversed, which, assuming it actually came from the proxy, suggests a key rotation has happened
 - .gitignore addition to drop some XML files ide is now creating 

### Change implications

### DIFF
--- a/java/gateway-core/src/main/java/com/avaulta/gateway/tokens/impl/AESReversibleTokenizationStrategy.java
+++ b/java/gateway-core/src/main/java/com/avaulta/gateway/tokens/impl/AESReversibleTokenizationStrategy.java
@@ -4,14 +4,14 @@ import com.avaulta.gateway.tokens.DeterministicTokenizationStrategy;
 import com.avaulta.gateway.tokens.ReversibleTokenizationStrategy;
 import lombok.*;
 
-import javax.crypto.Cipher;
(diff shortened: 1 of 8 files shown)


java/gateway-core/src/main/java/com/avaulta/gateway/tokens/impl/AESReversibleTokenizationStrategy.java
```diff
@@ -115,13 +114,20 @@ public String getOriginalDatum(@NonNull byte[] reversibleToken) {
...
+        try {
+            cipher.init(Cipher.DECRYPT_MODE, getKey(), cipherSuite.getParameterSpecGenerator().apply(reversibleToken));
 
-        byte[] plain = cipher.doFinal(cryptoText);
-        return new String(plain, StandardCharsets.UTF_8);
+            byte[] plain = cipher.doFinal(cryptoText);
+            return new String(plain, StandardCharsets.UTF_8);
+        } catch (IllegalBlockSizeException | InvalidAlgorithmParameterException |
+                 BadPaddingException | InvalidKeyException e) {
+            throw new RuntimeException(e);
```
jlorper, line 132: why don't you throw TokenInvalid here too?

PR #1006, opened 2025-09-23
eliminate tf runner exotic detection

eschultink:
### Fixes
- existing approach - conditioning `google_service_account_id_token` creation on another data resources, has started failing prior to first `terraform apply` for both tests and a developer's local; unclear why ... but that's a problem.
- solution using `external` to essentially grab variable assignments from top of module hierarchy was hacky

--> finally got this working 1) locally, 2) in tests, and 3) in google cloud shell. dropped use of external, while making the other resources all conditional + refactoring to invoke this higher in the stack SEEMS to be avoiding terraforms complaint about non-determinism.

### TODO:
- [x] try this in gcp cloud shell
- [x] try this in terraform cloud or similar
- [x] init the gcp principal variables in the usual case

### Change implications

 - something important to note in future release notes? **believe it's OK now ...**


infra/modules/gcp-tf-runner/main.tf
```diff
@@ -5,41 +5,35 @@
...
-#
-# however, this still appears to be the email of the authenticated user, not service account that
-# it's impersonating via the provider configuration
-#
-# we could pass in the service account email as a variable; but that pollutes a LOT of interfaces
-# and calls into question why we even use this module at all
+# this seems to be what Cloud Shell relies on??
 data "google_service_account_id_token" "identity" {
-  count = data.google_client_openid_userinfo.me.email == "" ? 1 : 0
+  count = !local.tf_runner_passed_in && try(data.google_client_openid_userinfo.me[0].email, "") == "" ? 1 : 0
```
eschultink, line 25: this works in cloud shell, even though that's not *really* a service account, afiak ... weird. But this is what makes cloud shell work *without* impersonation.

PR #1009, opened 2025-09-26
solve tf cycle

eschultink:
### Fixes
 - avoid terraform resource cycle
 - test-all script not dynamic, more succinct as far 

### Change implications

---
  - https://app.asana.com/0/0/1211452435681588


infra/modules/gcp-webhook-collector/test_script.tftpl
```diff
@@ -1,17 +1,24 @@
...
 ${command_cli_call} -u "${collector_endpoint_url}/.well-known/openid-configuration"
+EXIT_CODE=$((EXIT_CODE + $?))
 
 ${command_cli_call} -u "${collector_endpoint_url}/.well-known/jwks.json"
+EXIT_CODE=$((EXIT_CODE + $?))
 
 ${command_cli_call} -u "${collector_endpoint_url}${collection_path}" --method POST \
-%{ if signing_key_id != null ~}
---signing-key "gcp-kms:${signing_key_id}" \
+%{ if signing_key_arn != null ~}
```
aperez-worklytics, line 13: Tested on my local, and this should be 
```suggestion
%{ if signing_key_id != null ~}
```

infra/modules/gcp-webhook-collector/test_script.tftpl
```diff
@@ -1,17 +1,24 @@
...
+--signing-key "gcp-kms:${signing_key_arn}" \
```
aperez-worklytics, line 14:
```suggestion
--signing-key "gcp-kms:${signing_key_id}" \
```

PR #1012, opened 2025-10-06
recover avoid msft churn

eschultink:
### Fixes
 - msft churn - use `time_rotating` instead of `time_static` wrapping a function call, as truncation done in the expression isn't sufficient enough to stop Terraform from thinking there's a change
 - 

### Change implications

- dependencies added/changed? **yes, depends on 1.7.x**
- something important to note in future release notes? **you MUST upgrade terraform to 1.7.x, as minimum**


infra/modules/worklytics-connectors-msft-365/main.tf
```diff
@@ -1,22 +1,20 @@
...
   environment_id_display_name_qualifier = length(var.environment_id) > 0 ? " ${var.environment_id} " : ""
 }
 
-# beginning of current month - used for example API calls to provide recent but fairly stable dates
-# this will only change monthly, preventing churning while keeping examples relevant
-resource "time_static" "month_start" {
-  rfc3339 = formatdate("YYYY-MM-01'T'00:00:00Z", timestamp())
+# a timestamp used to seed example test API calls, so that they're likely to return interesting data, but the terraform test scripts don't churn too much
+resource "time_rotating" "example_timestamp" {
+  rotation_days = 365
```
jlorper, line 12: Maybe too much? 180 / 6mo probably enough

jlorper (approved):

PR #1068, opened 2025-12-04
wip : direct vpc egress

eschultink:
HOLD; requires google provider upgrade; waiting on google to expose direct vpc egress in the main google provider, instead of just google-beta.

seems supported now : 

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions2_function#argument-reference

### Features
 - use direct vpc egress instead of serverless connector

### Change implications

  - This implementation requires the google-beta provider until compatibility issues with the stable Google provider are resolved.
- something important to note in future release notes?  **any serverless connector that we provisioned for you will be destroyed, as no longer needed; if customer is providing subnet, must be \26 at min; \28 no longer sufficient/correct**


eschultink:
closing in favor of #1155

PR #1111, opened 2026-02-11
Leave settings as default route

aperez-worklytics:
There is a 401 while publishing packages from the action. It could be related to an invalid token, but I think the issue is that the settings are not in the same path as the java action is configured and the maven tool run in the script later. 

See my review comment

### Fixes
[401 while publishing psoxy packages](https://app.asana.com/1/27145998307022/project/1201039336231823/task/1213233589869316)

 - []()

 ### Logistics
 - []()

### Change implications

     obviously a no-op? **no**
     change **no**

### DIFF
--- a/.github/workflows/publish-release-artifacts.yaml
+++ b/.github/workflows/publish-release-artifacts.yaml
@@ -20,7 +20,7 @@ jobs:
         run: |
           REF_NAME="${{ github.ref_name }}"
           echo "Running on ref: $REF_NAME"
-          
+
           if [[ "$REF_NAME" == "main" ]]; then
             echo "✅ Validated running on main branch."
           elif [[ "$REF_NAME" =~ ^v ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-release-artifacts.yaml
```diff
@@ -41,7 +41,6 @@ jobs:
           java-version: '21'
           distribution: 'zulu'
           server-id: github # Value of the distributionManagement/repository/id field of the pom.xml
-          settings-path: ${{ github.workspace }} # Location for the settings.xml file
```
aperez-worklytics, line 44: This is the change. According to the [action docs](https://github.com/actions/setup-java?tab=readme-ov-file#maven-options), this is for setting where the settings file will be overwritten / created. 

`github.workspace` will [have](https://docs.github.com/en/actions/reference/workflows-and-actions/variables) the value of */home/runner/work/psoxy/psoxy*, but in the script maven is invoked without *--settings* overridden to use that path:

https://github.com/Worklytics/psoxy/blob/e41015571d58b56bddbba309ff794e2d789295a9/tools/release/publish-mvn-artifacts.sh#L131

So mvn is reading the settings from the default `.m2/settings.xml`, not */home/runner/work/psoxy/psoxy/settings.xml*

eschultink (approved):

PR #1119, opened 2026-02-28
misc changes for improved compatibility

eschultink:
### Fixes
- add some `ignore_changes`, variables found useful in some customer work into mainline

### Change implications

---
  - https://app.asana.com/0/0/1213472708629719

### DIFF
--- a/infra/modules/aws-psoxy-bulk/main.tf
+++ b/infra/modules/aws-psoxy-bulk/main.tf
@@ -76,6 +76,7 @@ resource "aws_s3_bucket" "input" {
 
   lifecycle {
     ignore_changes = [
+      bucket, # avoid recreation of the bucket
       tags
     ]
   }
--- a/infra/modules/aws-psoxy-lambda/main.tf
+++ b/infra/modules/aws-psoxy-lambda/main.tf
@@ -150,6 +150,7 @@ resource "aws_iam_role" "iam_for_lambda" {
 
   lifecycle {
     ignore_changes = [
+      name, # just for backwards compatibility with legacy deployment
       tags
     ]
   }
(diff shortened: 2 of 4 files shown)


infra/modules/aws-psoxy-bulk/main.tf
```diff
@@ -76,6 +76,7 @@ resource "aws_s3_bucket" "input" {
 
   lifecycle {
     ignore_changes = [
+      bucket, # avoid recreation of the bucket
```
eschultink, line 79: *ever* destroying the buckets is bad, bc will require reconfigurations; so avoid it in all cases.

infra/modules/aws-psoxy-lambda/main.tf
```diff
@@ -150,6 +150,7 @@ resource "aws_iam_role" "iam_for_lambda" {
 
   lifecycle {
     ignore_changes = [
+      name, # just for backwards compatibility with legacy deployment
```
eschultink, line 153: changing lambda names would force function re-deploy. not necessarily a huge problem, but would necessitate some reconfiguration as well as add noise in tf plan as to changes.

infra/modules/private-key-aws-parameter-user-managed/variables.tf
```diff
@@ -18,3 +18,8 @@ variable "kms_key_id" {
   description = "KMS key ID or ARN to use for encrypting secrets. If not provided, secrets will be encrypted by SSM with its keys (controlled by AWS)."
   default     = null
 }
+
+variable "ssm_path" {
+  type        = string
+  description = "path/prefix in SSM Parameter Store at which to create parameters to hold keys"
+}
```
eschultink, line 25: this missing is something of a bug for users using this module.

aperez-worklytics (approved):

PR #1128, opened 2026-03-09
S220 health check fix

jlorper:
### Fixes
- [psoxy healtcheck does not show salt hash if using AWS Secret Store](https://app.asana.com/1/27145998307022/project/1213457449259792/task/1213569868602897)
- Reading SALT from configService, not secretStore for health check / piiSaltSha

### Change implications

 - something important to note in future release notes? no
 - breaking changes? no

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandler.java
@@ -3,7 +3,13 @@ package co.worklytics.psoxy.gateway.impl;
 import co.worklytics.psoxy.ControlHeader;
 import co.worklytics.psoxy.HashUtils;
 import co.worklytics.psoxy.HealthCheckResult;
-import co.worklytics.psoxy.gateway.*;
+import co.worklytics.psoxy.gateway.ApiModeConfigProperty;
+import co.worklytics.psoxy.gateway.ConfigService;
+import co.worklytics.psoxy.gateway.HttpEventRequest;
+import co.worklytics.psoxy.gateway.HttpEventResponse;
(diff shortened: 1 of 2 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandler.java
```diff
@@ -218,7 +224,7 @@ private HttpEventResponse handle(HttpEventRequest request) {
      */
     public String piiSaltHash() {
         if (piiSaltHash == null) {
-            piiSaltHash = config.getConfigPropertyAsOptional(ProxyConfigProperty.PSOXY_SALT)
+            piiSaltHash = secretStore.getConfigPropertyAsOptional(ProxyConfigProperty.PSOXY_SALT)
```
eschultink, line 227: possibly not correct in the cli case ...

eschultink:
why do you think this bug exists? it works for me:

image

eschultink (commented):
where's the bug?

eschultink:
don't see it missing in gcp either ...

image

jlorper:
If using AWS_SECRETS_MANAGER it fails because configService defaults to ParameterStore.
I think in GCP is properly handled. But check actual code of the HealthCheck, is already using the secretStore to check the SALT first.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/HealthCheckRequestHandler.java
jlorper, line 126: @eschultink see here, reading from secretStore

eschultink: if this is `true` for them then, should be OK, right?

eschultink (approved):

PR #1141, opened 2026-03-17
Cursor rules update

aperez-worklytics:
Update Cursor rules including reversible pseudonyms for email and id on team members
Also, adding a basic test for existing rules.

### Fixes
[Update psoxy rules](https://app.asana.com/1/27145998307022/task/1213688733955168)

### Change implications

     obviously a no-op? **no**
     change **no**

---
  - https://app.asana.com/0/0/1213688733955168


docs/sources/cursor/example-api-responses/original/team-members.json
```diff
@@ -1,24 +1,28 @@
...
-            "name": "Taylor",
-            "email": "team.lead@company.com",
-            "role": "admin"
-        }
-    ]
-}
\ No newline at end of file
+  "teamMembers": [
+    {
+      "name": "Alex",
```
jlorper, line 4: are the id ommited because the test will fail for the integer/string issue?

eschultink: i think we should include `id` for completeness/correctness. this is the proper immutable identifier for the account; `email` is potentially changeable.

eschultink (approved):

PR #1143, opened 2026-03-18
Support regex with multiple matches

aperez-worklytics:
Add support for processing multiple pseudonymizations on `PseudonymizeRegexMatches` instead of a single one.

### Features
[Supporting multiple pseudonymizations on regex](https://app.asana.com/1/27145998307022/project/1201039336231823/task/1213719485461863)

### Change implications

     obviously a no-op?
     change


java/core/src/main/java/co/worklytics/psoxy/impl/SanitizerUtils.java
```diff
@@ -395,27 +395,56 @@ public MapFunction getPseudonymizeRegexMatches(Pseudonymizer pseudonymizer, Tran
...
+                    matcher.appendReplacement(result, Matcher.quoteReplacement(pseudonymizedString));
                 }
+            }
+
+            if (found) {
+                matcher.appendTail(result);
+                return result.toString();
             } else {
-                //if no match, redact it
+                // original behavior: if no real match, redact
```
eschultink, line 448: yeah, this is probably not the right behavior.

eschultink: perhaps mark with TODO to change in next version with breaking changes; and/or we introduce a new transform that behaves consistently, and mark this one as deprecated.

eschultink (approved):
seems correct and reasonable; but what's the use-case?

I'd also mention it in the CHANGELOG; it IS a behavior change that would break things in theory, but I do think reasonable to view the change as a fix. result will be stuff being sanitized that wasn't before, not the reverse - so not something that needs to wait for an full version increment.

aperez-worklytics:
> seems correct and reasonable; but what's the use-case?
> 

Mostly this: https://github.com/Worklytics/psoxy/pull/1143/changes#diff-fb79884f2ef53e88da07cfc05bb4f88961f8085f5142ae1dcbe76d4653ad5826R513

That it is something that I'll need for gitlab, as some actions appears as part of system notes instead of proper event structure.

PR #1146, opened 2026-03-24
Add missing perm on builder

aperez-worklytics:
After #1145, builder-sa requires `storageViewer` role; otherwise it cannot access to the bucket to check the status of the function during deployment

### Fixes
[Missing role for builder-sa in GCP](https://app.asana.com/1/27145998307022/project/1213638267413355/task/1213802944713819)

 - []()

 ### Logistics
 - []()

### Change implications

     obviously a no-op? **no**
     change **no**

### DIFF
--- a/infra/modules/gcp/main.tf
+++ b/infra/modules/gcp/main.tf
@@ -327,6 +327,14 @@ resource "google_service_account" "proxy_builder_sa" {
   project      = var.project_id
 }
 
+resource "google_project_iam_member" "grant_proxy_builder_as_storage_object_viewer" {
+  count = var.provision_project_level_iam ? 1 : 0
+
+  project = var.project_id
+  role    = "roles/storage.objectViewer"
+  member  = "serviceAccount:${google_service_account.proxy_builder_sa.email}"
+}
+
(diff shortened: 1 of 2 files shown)


eschultink:
my bad; yeah, moving to worktrees I don't have my dev envs set-up with all of those; only the main clone ... need to think on best approach ...

infra/modules/gcp/main.tf
```diff
@@ -327,6 +327,14 @@ resource "google_service_account" "proxy_builder_sa" {
   project      = var.project_id
 }
 
+resource "google_storage_bucket_iam_member" "grant_proxy_builder_object_viewer_on_artifacts" {
+  count = local.is_remote_bundle ? 0 : 1
+
+  bucket = google_storage_bucket.artifacts[0].name
+  role   = "roles/storage.objectViewer"
+  member = "serviceAccount:${local.builder_sa_email}"
+}
```
eschultink, line 336: so made this more granular - just on the bucket; was both claude feedback and my own.

aperez-worklytics: makes sense

eschultink (approved):

PR #1160, opened 2026-04-06
use GCP Parameter Manager for non-secret config 

eschultink:
### Features
- use the new GCP Parameter Manager, which is now GA and supported in google tf provider 7.x, for non-secret configuration data

### Change implications

- dependencies added/changed? **yes - gcloud jar with that; depends on more perms for customers utilizing GCP**
- something important to note in future release notes? **yes, breaking change**

---
  - https://app.asana.com/0/0/1210060335913182
  - https://app.asana.com/0/0/1211398767873390


java/core/src/test/java/co/worklytics/psoxy/gateway/impl/InMemorySecretStore.java
aperez-worklytics, line 1: 👍

infra/modules/gcp-pm-rules/variables.tf
```diff
@@ -0,0 +1,40 @@
...
+variable "content" {
+  type        = string
+  description = "Raw rules content string. Mutually exclusive with file_path; one of the two must be provided."
+  default     = null
+}
+
+variable "instance_sa_email" {
+  type        = string
+  description = "The email address of the service account to use for the proxy instance that will access the rules."
+}
```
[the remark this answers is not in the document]
aperez-worklytics: true; @eschultink it seems we can drop this

infra/modules/gcp/main.tf
```diff
@@ -34,6 +34,7 @@ resource "google_project_service" "gcp_infra_api" {
     "compute.googleapis.com", # seems required w newer Google provider versions, for resources we use
     "iam.googleapis.com",     # manage IAM via terraform (as of 2023-04-17, internal dev envs didn't have this; so really needed?)
     "run.googleapis.com",     # required for cloud run functions gen2
+    "parametermanager.googleapis.com",
```
aperez-worklytics, line 37:
```suggestion
    "parametermanager.googleapis.com", # used to store configuration parameters that are shared / too large for function environment
```

aperez-worklytics (approved):
LGTM; and I also think that some Copilot comments are worth to consider

infra/modules/gcp-rules-pm/variables.tf
```diff
@@ -0,0 +1,40 @@
...
+  validation {
+    condition     = var.file_path == null || try(endswith(var.file_path, ".yaml"), false)
+    error_message = "Rules should be plain .yaml file."
+  }
+}
+
+variable "content" {
+  type        = string
+  description = "Raw rules content string. Mutually exclusive with file_path; one of the two must be provided."
+  default     = null
```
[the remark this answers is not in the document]
eschultink: tf doesn't support validations referencing multiple variables.

infra/modules/gcp-host/main.tf
```diff
@@ -249,15 +252,26 @@ module "custom_api_connector_rules" {
...
   project_id        = var.gcp_project_id
-  prefix            = "${local.config_parameter_prefix}${upper(replace(each.key, "-", "_"))}_"
+  prefix            = "${local.pm_shared_prefix}${replace(each.key, "-", "/")}/"
   content           = each.value
   instance_sa_email = module.api_connector[each.key].service_account_email
 }
 # END API CONNECTORS
 
+# Grant all API connector SAs the Parameter Reader role at project level
+# (per-parameter IAM is not yet available in the Google provider for PM)
```
eschultink, line 259: mmm, so this may be a deal-breaker; some customers don't allow project-level IAM to be be provisioned by terraform - so we have to be very limited with that. It requires that 1) this has to be conditional on `provision_project_iam` variable, and 2) for those customers, this is N additional grants they must make outside our tf ...

PR #1173, opened 2026-04-23
add salt hash to metadata on bulk files 

eschultink:
### Features
 - add salt hash to metadata on bulk files 

### Change implications

 - something important to note in future release notes? **none**
 - breaking changes? **no**

---
  - https://app.asana.com/0/0/1214248670311472

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
@@ -12,6 +12,7 @@ import java.io.Serial;
 import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -104,6 +105,8 @@ public class StorageHandler {
 
     @Inject
     HashUtils hashUtils;
+
+    private volatile String piiSaltHash;
 
     static void warnIfEncodingDoesNotMatchFilename(@NonNull StorageEventRequest request, @Nullable String contentEncoding) {
         if (request.getSourceObjectPath().endsWith(EXTENSION_GZIP)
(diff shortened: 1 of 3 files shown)


java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
```diff
@@ -233,12 +244,29 @@ public List<ObjectTransform> buildTransforms() {
...
+        if (StringUtils.isNotBlank(hash)) {
+            metadata.put(BulkMetaData.SALT_SHA.getMetaDataKey(), hash);
+        }
+
+        return metadata;
+    }
+
+    String piiSaltHash;
+
+    public String piiSaltHash() {
```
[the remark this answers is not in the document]
aperez-worklytics: Agree

java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
```diff
@@ -233,12 +244,29 @@ public List<ObjectTransform> buildTransforms() {
     public Map<String, String> buildObjectMetadata(String sourceBucket, String sourceKey, ObjectTransform transform) {
         //transform currently unused; in future we probably want to record what transform was
         // applied, to aid traceability of pipelines
-        return Map.of(
+        Map<String, String> metadata = new java.util.HashMap<>(Map.of(
```
aperez-worklytics: I'd go for `LinkedHashMap` so we can keep the same order of the properties

aperez-worklytics (approved):
No blockers; two comments I think are worth to apply

java/core/src/main/java/co/worklytics/psoxy/storage/StorageHandler.java
```diff
@@ -233,12 +244,29 @@ public List<ObjectTransform> buildTransforms() {
...
+        return metadata;
+    }
+
+    String piiSaltHash;
+
+    public String piiSaltHash() {
+        if (piiSaltHash == null) {
+            piiSaltHash = secretStore.getConfigPropertyAsOptional(ProxyConfigProperty.PSOXY_SALT)
+                .map(salt -> hashUtils.hash(salt, ProxyConstants.SALT_FOR_SALT)).orElse("");
+        }
```
jlorper: I think this is fine for AWS lambda and GCP 1st generation, but 2nd gen functions can handle multiple request concurrently, so this should be guarded agains that:

```suggestion
        if (piiSaltHash == null) {
            synchronized(StorageHandler.class) { // or a custom lock for this purpose
            piiSaltHash = secretStore.getConfigPropertyAsOptional(ProxyConfigProperty.PSOXY_SALT)
                .map(salt -> hashUtils.hash(salt, ProxyConstants.SALT_FOR_SALT)).orElse("");
            }
        }
```
if we ever enable gen2 we need to be careful with this pattern

jlorper (approved):

PR #1174, opened 2026-04-27
Propagate error info to caller

jlorper:
Try to propagate some info to the caller on the FAILED_TO_BUILD_URL error case.

### Change implications

     obviously a no-op? no
     change -> no

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/ProcessedDataMetadataFields.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/ProcessedDataMetadataFields.java
@@ -25,11 +25,6 @@ public enum ProcessedDataMetadataFields {
     ERROR("Error"),
 
     /**
-     * a specific error code while processing the request, complements ERROR (optional)
-     */
-    ERROR_CODE("Error-Code"),
-
-    /**
      * a warning code while processing the request
      */
     WARNING("Warning"),
--- a/java/core/src/main/java/co/worklytics/psoxy/ProcessedDataMetadataFields.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/ProcessedDataMetadataFields.java
@@ -25,6 +25,11 @@ public enum ProcessedDataMetadataFields {
     ERROR("Error"),
 
     /**
+     * a specific error code while processing the request, complements ERROR (optional)
+     */
+    ERROR_CODE("Error-Code"),
+
+    /**
      * a warning code while processing the request
      */
     WARNING("Warning"),
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
@@ -236,11 +236,9 @@ public class ApiDataRequestHandler {
             // InvalidTokenException extends RuntimeException
             if (e instanceof ReversibleTokenizationStrategy.InvalidTokenException ite) {
                 return HttpEventResponse.builder()
-                    // should this be a 500? Maybe 422 Unprocessable Content?
-                    .statusCode(HttpStatus.SC_INTERNAL_SERVER_ERROR)
+                    .statusCode(HttpStatus.SC_UNPROCESSABLE_ENTITY)
                     .header(ProcessedDataMetadataFields.ERROR.getHttpHeader(),
-                        ErrorCauses.FAILED_TO_BUILD_URL.name())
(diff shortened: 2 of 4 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
jlorper, line 231: this catching Throwable doesn't give us much info on what could be happening.
Provide InvalidTokenException with some error codes (probably we should do something like this so we can have more info on our end on what can be happening in order to help the customer)

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -241,6 +233,17 @@ public HttpEventResponse handle(HttpEventRequest requestToProxy,
             // our canonicalization code for this to go wrong
             log.log(Level.WARNING, "Error parsing  / building request URL", e);
 
+            // InvalidTokenException extends RuntimeException
+            if (e instanceof ReversibleTokenizationStrategy.InvalidTokenException ite) {
+                return HttpEventResponse.builder()
+                    // should this be a 500? Maybe 422 Unprocessable Content?
```
jlorper: 500 will make us retry, and same call would yield same results.
I think 422 makes more sense in this case and we can handle globally on Worklytics' side

java/gateway-core/src/main/java/com/avaulta/gateway/tokens/ReversibleTokenizationStrategy.java
```diff
@@ -41,11 +43,32 @@ default byte[] getReversibleToken(String originalDatum) {
...
     /**
      * Indicates that the token could not be reversed, likely because invalid
-     *
      */
     class InvalidTokenException extends RuntimeException {
-        public InvalidTokenException(String message, Throwable cause) {
-            super(message, cause);
+
+        @Getter
+        public enum ErrorCode {
```
jlorper, line 50: maybe not the best approach, but for certain cases (rules exceptions, this, and access errors) could be helpful to exactly know what's going on rather than a generic error

java/core/src/main/java/co/worklytics/psoxy/ProcessedDataMetadataFields.java
```diff
@@ -24,6 +24,11 @@ public enum ProcessedDataMetadataFields {
      */
     ERROR("Error"),
 
+    /**
+     * a specific error code while processing the request, complements ERROR (optional)
+     */
+    ERROR_CODE("Error-Code"),
```
eschultink: i think Error is a value from an enum ... so is specific

java/gateway-core/src/main/java/com/avaulta/gateway/tokens/ReversibleTokenizationStrategy.java
```diff
@@ -41,11 +43,32 @@ default byte[] getReversibleToken(String originalDatum) {
...
+            ALGORITHM_PARAMETER_ERROR("ITE01", "Failed to decrypt token; some algorithm parameter, such as iv, is wrong"),
```
eschultink: no need to be so clever here; just return whole enum names ... don't care about a few extra chars in error case

eschultink (changes requested):
i think just use `X-Psoxy-Error` that exists; no need to create something else that basically duplicates that.

and return whole error enum name or exception name. No need to add extra levels of indirection to save a few chars, but require us to grep the code to determine the actual error

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -241,6 +233,17 @@ public HttpEventResponse handle(HttpEventRequest requestToProxy,
...
+                    .statusCode(HttpStatus.SC_INTERNAL_SERVER_ERROR)
+                    .header(ProcessedDataMetadataFields.ERROR.getHttpHeader(),
+                        ErrorCauses.FAILED_TO_BUILD_URL.name())
```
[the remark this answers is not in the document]
eschultink: yeah, I agree not a 500 ... whatever 4xx seems most semantically correct for the case.

java/core/src/main/java/co/worklytics/psoxy/ProcessedDataMetadataFields.java
eschultink: so if Error is a strict code -- I think eliminate this, just use `Error`; but if it's freeform message text, OK.

PR #1178, opened 2026-04-29
Gong api fixes

aperez-worklytics:
Couple of issues spotted with Gong:
1) APi endpoint should have `instance.api.gong.io`; that `api` part was missing when adding the URL. Now it will be provided handle if missing or included.
2) Auth strategy was wrong; it should be `oauth2_refresh_token` instead of `basic_auth`

Docs updated.

### Fixes
[bug: gong subdomain issues](https://app.asana.com/1/27145998307022/project/1213797956438900/task/1214359291881725)

 - []()

 ### Logistics
 - []()

### Change implications

     obviously a no-op?
     change


jlorper (approved):
We could pass the entire host as a variable, instead of just asking for the subdomain.
`gong_instance_subdomain = customer.api.gong.io`

PR #1180, opened 2026-04-29
refactor API connector options

eschultink:
### Features
- improve how we pass connector-specific settings into worklytics-connectors module, so examples more future-proof

### Change implications

 - something important to note in future release notes? **could flag stuff is deprecated**

---
  - https://app.asana.com/0/0/1214358474410692


aperez-worklytics (approved):
ok; what about `google-workspace-variables` and `msft-365-variables` ?

PR #1183, opened 2026-04-30
avoid tf cycle in upgrades

eschultink:
@aperez-worklytics lmk if this works for you ...

### Fixes
- avoid reported cycle in upgrade

### Change implications


jlorper (approved):

eschultink:
doesn't seem to solve the cycle, and is also inelegant soln breaking encapsulation of the modules.

eschultink:
going to close; have solved cycle issue separately, and will break out TODO refactor into distinct PR targeting that in next release.

PR #1191, opened 2026-05-05
todo files created at top

eschultink:
ON HOLD - moving all the way to top `main.tf` is potentially breaking change for those who forked/cloned examples earlier; needs to go to next major release.

### Features
 - move all `local_file` resources to top/root - improves compatibility with Terraform Cloud and the like; as well as avoids some customers static analyzers complaining about the local_file stuff, even when it's disabled
 - limit potential of cycles with todo_step/next_todo_step being computed internally in modules
 - structure/refactor in a way that will be more future-proof for module distribution

### Change implications

 - something important to note in future release notes? *yes, see CHANGELOG*

---
  - https://app.asana.com/0/0/1214540303176940


eschultink:
reasonable thing, but https://github.com/Worklytics/psoxy/pull/1437 more versatile soln

PR #1185, opened 2026-05-02
augments PoC

eschultink:
### Features
- PoC of augments feature, to enrich/add workplace metadata - beyond simply sanitizing

### Change implications

 - dependencies added/changed? **no**
 - something important to note in future release notes?
   - No significant changes expected in release notes.
   - breaking changes? **no**

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1211958863440336

PR #1197, opened 2026-05-11
sentence-metadata augment poc

eschultink:
### Features
 - a proof of concept (PoC) for sentence metadata augmentation

NOTE: branches from #1185 

### Change implications

 - something important to note in future release notes? **jar size**

---
  - https://app.asana.com/0/0/1214428923612149


docs/development/sentence-metadata-augment.md
```diff
@@ -0,0 +1,108 @@
...
+The primary use cases include analyzing LLM prompts to understand their structure and semantic elements, extracting insights from text fields in API payloads without exposing sensitive raw text data, and identifying trends in prompt structure such as the use of imperatives, constraints, or hedges.
+
+## Output Schema
+The `+sentenceMetadata` block is structured with a `sentences` array containing one object per detected sentence, and a `doc_summary` object containing net-new aggregations not derivable from a simple flatten of the sentence list.
+
+### Annotated JSON Example
+
+```json
+{
+  "+prompt.sentenceMetadata": {
```
jlorper, line 19: PoC so OK, but we'd need to check how much the payload is going to increase. 
JSON gzipped I guess it should be fine though

java/core/src/main/java/co/worklytics/psoxy/FunctionRuntimeModule.java
```diff
@@ -135,12 +136,19 @@ static ConfigService configService(EnvVarsConfigService envVarsConfigService,
...
-        return CompositeResourceService.builder()
+        ResourceService instanceResourceService = CompositeResourceService.builder()
             .preferred(new LocalFileResourceService(ResourceService.DEFAULT_LOCAL_RESOURCE_PATH))
             .fallback(remoteResourceService)
             .build();
+        ResourceService openNlpResourceService = CompositeResourceService.builder()
+            .preferred(instanceResourceService)
+            .fallback(sharedRemoteResourceService)
+            .build();
+        SentenceMetadataProcessor.configureResourceService(openNlpResourceService);
```
jlorper: don't like this, initializing SentenceMetadataProcessor in ResourceService, just because is static.
totally unrelated with this method's purpose.

jlorper: probably the solution should be something similar to the `Initiable` / config them at runtime before use, rather than using static classes to perform the action

jlorper (approved):
OK for a PoC, but mixing responsibilities on the module, configuring SentenceMetadataProcessor in an unrelated provider. This should be addressed at some point.

docs/development/sentence-metadata-augment.md
eschultink: yeah - i think it's a PoC and we need to test NLP effectiveness (is it helpful?) and then customer preference for NLP vs generative approaches.

PR #1199, opened 2026-05-12
remote resource services

eschultink:
### Features
 - support remote resources stored in gcs/S3; use cases are potentially large rule sets, llm models, nlp models - that can't efficiently be in a shared bundle common to all proxy instances

### Change implications

 - something important to note in future release notes? **IAM changes, if have gcs/s3 artifacts buckets that exist**

---
  - https://app.asana.com/0/0/1214540303176947


java/core/src/main/java/co/worklytics/psoxy/gateway/ResourceService.java
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * Check whether a resource exists at the given path without opening a stream.
+     *
+     * <p>Default implementation attempts to open and immediately close the resource.
+     * Implementations should override for efficiency where possible.</p>
+     *
+     * @param objectPath path to the resource
+     * @return true if the resource exists
+     */
+    default boolean exists(String objectPath) {
```
aperez-worklytics: I'd drop this default impl. Depending on the `getResource`, an stream can be opened as this is considering even the comment says *without opening a stream*

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/LocalFileResourceService.java
```diff
@@ -0,0 +1,54 @@
...
+    @NonNull
+    final String basePath;
+
+    @Override
+    public Optional<InputStream> getResource(String objectPath) {
+        Path resolved = Paths.get(basePath, objectPath);
+        if (!Files.exists(resolved) || !Files.isRegularFile(resolved)) {
+            log.log(Level.FINE, "Local resource not found: {0}", resolved);
+            return Optional.empty();
+        }
```
aperez-worklytics: Use existing method

```suggestion
        if (!exists(objectPath) {
            log.log(Level.FINE, "Local resource not found: {0}", resolved);
            return Optional.empty();
        }
```

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/LocalFileResourceService.java
```diff
@@ -0,0 +1,54 @@
...
+
+        try {
+            log.log(Level.INFO, "Loading resource from local filesystem: {0}", resolved);
+            return Optional.of(new FileInputStream(resolved.toFile()));
+        } catch (Exception e) {
```
aperez-worklytics: Worth to just handle `IOException` instead of `Exception` ?

java/core/src/main/java/co/worklytics/psoxy/gateway/ResourceService.java
aperez-worklytics: Also, seems to be only called from impleemntations, not other stuff? Worth to drop and just each implementation could have its own private method for this ?

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/S3ResourceService.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Path prefix within the bucket (may be empty string for root).
+     */
+    @NonNull
+    final String pathPrefix;
+
+    @Override
+    public Optional<InputStream> getResource(String objectPath) {
+        String key = resolveKey(objectPath);
```
aperez-worklytics, line 41: In `LocalFileResourceService` and `GcsResourceService` we are checking if the *resource* exists or not. Here not, just relying on the exception found... I'd prefer to just check `exists(objectPath)` instead

java/impl/gcp/src/main/java/co/worklytics/psoxy/GcsResourceService.java
```diff
@@ -0,0 +1,84 @@
...
+    @NonNull
+    final String pathPrefix;
+
+    @Override
+    public Optional<InputStream> getResource(String objectPath) {
+        String key = resolveKey(objectPath);
+        BlobId blobId = BlobId.of(bucketName, key);
+
+        try {
+            if (storage.get(blobId) == null) {
```
aperez-worklytics: Same logic as  `exists`

java/core/src/main/java/co/worklytics/psoxy/ConfigRulesModule.java
```diff
@@ -76,11 +86,32 @@ static RuleSet rules(Logger log,
...
 
     }
 
+    /**
+     * Attempt to load rules from the InstanceResourceService (local FS or remote S3/GCS bucket).
+     */
+    static Optional<RuleSet> getRulesFromResource(Logger log, RulesUtils rulesUtils, ResourceService resourceService) {
+        try {
+            Optional<InputStream> rulesStream = resourceService.getResource(RULES_RESOURCE_PATH);
+            if (rulesStream.isEmpty()) {
```
aperez-worklytics, line 102:
```suggestion
            if (resourceService.exists(RULES_RESOURCE_PATH)) {
```

to avoid potentials streams to be opened and not closed (depending on implementation)

docs/configuration/remote-resources.md
```diff
@@ -0,0 +1,122 @@
...
+psoxy will check for a `RULES` object at `{INSTANCE_RESOURCE_PATH}/RULES` in the remote bucket.
+
+A hardcoded local filesystem path (`/var/psoxy/resources`) is also checked before the remote
+bucket, providing a fast-path for containerized or VM-based deployments where resources can be
+mounted locally.
+
+## Terraform Configuration
+
+By default, the Terraform modules in this repository will **automatically** configure the
+`REMOTE_RESOURCE_BUCKET` for you if you pass the `remote_resource_bucket` variable to your host
```
aperez-worklytics: I'd remark that `REMOTE_RESOURCE_BUCKET` should exist and it is not created by our TF script; and also that the TF runner should have access to that bucket as we are going to enable the SA to read content from it

aperez-worklytics (changes requested):
Cool; some changes about `exists` logic to be reused

java/core/src/main/java/co/worklytics/psoxy/ConfigRulesModule.java
```diff
@@ -23,6 +26,12 @@ public class ConfigRulesModule {
 
     public static final String NO_APP_IDS_SUFFIX = "_no-app-ids";
 
+    /**
+     * Well-known resource path for rules loaded from InstanceResourceService (local FS or remote
+     * cloud storage).
+     */
+    public static final String RULES_RESOURCE_PATH = "rules.yaml";
```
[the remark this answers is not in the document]
eschultink: i don't agree ...

java/impl/gcp/src/main/java/co/worklytics/psoxy/GcsResourceService.java
```diff
@@ -0,0 +1,84 @@
...
+                log.log(Level.FINE, "GCS resource not found: gs://{0}/{1}", new Object[]{bucketName, key});
+                return Optional.empty();
+            }
+
+            ReadChannel readChannel = storage.reader(blobId);
+            log.log(Level.INFO, "Loaded resource from GCS: gs://{0}/{1}", new Object[]{bucketName, key});
+            return Optional.of(Channels.newInputStream(readChannel));
```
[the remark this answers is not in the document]
eschultink: tried to refactor this. note i don't want a caching service, due to risk of mem issues if cache stuff like llm models loaded as resources

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/S3ResourceService.java
eschultink: tried to make more consistent

aperez-worklytics (approved):

PR #1219, opened 2026-05-23
Reuse cached OAuth token to avoid redundant same-instance refresh

eschultink:
## Summary

Fixes redundant refresh-token exchanges when multiple requests on the **same proxy instance** need a fresh access token at the same time.

- **Primary case (non-shared tokens):** `USE_SHARED_TOKEN=false` connectors do not use the cross-instance lock. `refreshAccessToken()` is `synchronized`, so threads serialize, but the second thread previously still performed another exchange because `checkIfAlreadyRefreshed` only consulted the shared token store (always empty here). Now `findReusableAccessToken()` checks the in-memory cache first.
- **Secondary case (shared tokens):** cross-instance dedupe is already handled by the distributed lock + secret-store re-read. Checking the in-memory cache first is defense in depth when the store read lags.
- `cachedToken` is documented as instance-shared state and marked `volatile` so unsynchronized reads in `getCredentials()` see updates promptly.

Split out from #1216 for independent review.

## Test plan
- [x] `mvn -pl core -am -Dtest=OAuthRefreshTokenSourceAuthStrategyTest -Dsurefire.failIfNoSpecifiedTests=false test`

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
@@ -444,18 +444,17 @@ public class OAuthRefreshTokenSourceAuthStrategy implements SourceAuthStrategy {
          * @return a non-expired token to reuse, or empty if a refresh exchange is required
          */
         private Optional<AccessToken> findReusableAccessToken() {
-            AccessToken freshToken = sourceAuthStrategy.getCachedToken();
-            if (!sourceAuthStrategy.shouldRefresh(freshToken, clock.instant())) {
-                DevLogUtils.info(envVarsConfigService, log, "Token already refreshed " + freshToken.getExpirationTime());
-                return Optional.of(freshToken);
+            AccessToken cachedToken = sourceAuthStrategy.getCachedToken();
+            if (!sourceAuthStrategy.shouldRefresh(cachedToken, clock.instant())) {
+                return Optional.of(cachedToken);
             }
 
-            freshToken = sourceAuthStrategy.getSharedAccessTokenIfSupported().orElse(null);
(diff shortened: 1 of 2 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -434,12 +434,23 @@ public synchronized AccessToken refreshAccessToken() throws IOException {
...
+         * <p>Checks the in-memory cache first (same-instance refreshes), then the shared token
+         * store when {@link OAuthRefreshTokenSourceAuthStrategy#useSharedToken()} is true
+         * (cross-instance refreshes).
+         *
+         * @return a non-expired token to reuse, or empty if a refresh exchange is required
          */
-        private Optional<AccessToken> checkIfAlreadyRefreshed() {
-            AccessToken freshToken = sourceAuthStrategy.getSharedAccessTokenIfSupported().orElse(null);
+        private Optional<AccessToken> findReusableAccessToken() {
+            AccessToken freshToken = sourceAuthStrategy.getCachedToken();
```
eschultink: poorly named? not necessarily a freshToken?

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/oauth/OAuthRefreshTokenSourceAuthStrategy.java
```diff
@@ -434,12 +434,23 @@ public synchronized AccessToken refreshAccessToken() throws IOException {
...
+            if (!sourceAuthStrategy.shouldRefresh(freshToken, clock.instant())) {
+                DevLogUtils.info(envVarsConfigService, log, "Token already refreshed " + freshToken.getExpirationTime());
```
eschultink: is static use of this log utils method the usual pattern?

aperez-worklytics (approved):

PR #1232, opened 2026-05-27
azure perms prereq

eschultink:
### Features
- add azure perms prereq check, if possible

### Change implications

---
  - https://app.asana.com/0/0/1202867174049854

### DIFF
--- a/tools/check-prereqs.sh
+++ b/tools/check-prereqs.sh
@@ -187,84 +187,49 @@ else
       printf ".\n"
     fi
 
-    POWERSHELL_CMD=""
-    if command -v pwsh &> /dev/null; then
-      POWERSHELL_CMD="pwsh"
-    elif command -v powershell &> /dev/null; then
-      POWERSHELL_CMD="powershell"
+    AZ_GRAPH_ROLES_URL="https://graph.microsoft.com/v1.0/me/transitiveMemberOf/microsoft.graph.directoryRole"
+    AZ_SKIP_ROLE_CHECK=false
+    AZ_PRINCIPAL_TYPE=$(az account show --query user.type -o tsv 2>/dev/null)
+    if [[ "$AZ_PRINCIPAL_TYPE" == "servicePrincipal" ]]; then
+      AZ_SP_APP_ID=$(az account show --query user.name -o tsv 2>/dev/null)
+      AZ_SP_OBJECT_ID=$(az ad sp show --id "$AZ_SP_APP_ID" --query id -o tsv 2>/dev/null)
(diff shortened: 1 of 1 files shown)


tools/check-prereqs.sh
```diff
@@ -165,12 +165,101 @@ printf "\n"
...
+        printf " (tenant ${CODE}${AZ_TENANT_ID}${NC})"
+      fi
+      printf ".\n"
+    fi
+
+    POWERSHELL_CMD=""
+    if command -v pwsh &> /dev/null; then
+      POWERSHELL_CMD="pwsh"
+    elif command -v powershell &> /dev/null; then
+      POWERSHELL_CMD="powershell"
```
aperez-worklytics: Powershell is *required* in case of online meetings for Teams for setup the permissions required between the app the teams policies. But rest of connectors, it is not used

tools/check-prereqs.sh
```diff
@@ -165,12 +165,101 @@ printf "\n"
...
+      esac
+    else
+      printf "\t- ${WARN}PowerShell is not installed; cannot automatically verify Entra directory roles.${NC}\n"
+      printf "\t  ${AZ_ENTRA_ADMIN_ROLES_REASON}\n"
+      printf "\t  Verify role assignments in the Microsoft Entra admin center (see ${AZ_ENTRA_ROLES_DOC}), or install PowerShell (${CODE}pwsh${NC} on macOS/Linux) and re-run this script.\n"
+    fi
+  else
+    printf "\t- ${WARN}Azure CLI is not authenticated.${NC} Run ${CODE}az login --allow-no-subscriptions${NC} to authenticate.\n"
+    printf "\t  Once authenticated, re-run this script to verify Entra directory role requirements (${CODE}Global Administrator${NC} or ${CODE}Application Administrator${NC}).\n"
+  fi
```
aperez-worklytics: 🔴 No need to use curl, we use a simpler approach there.

```
az rest --method GET \
  --url "https://graph.microsoft.com/v1.0/me/transitiveMemberOf/microsoft.graph.directoryRole" \
  --query "value[].{Role:displayName, Description:description}" \
  --output table
```

That will return something like:

```
Role                       Description
-------------------------  -----------------------------------------------------------------------------------------
Global Administrator       Company Administrator role has full access to perform any operation in the company scope.
Application Administrator  Can create and manage all aspects of app registrations and enterprise apps.
```

And then you can check if the roles matches what we expect

aperez-worklytics (changes requested):

aperez-worklytics (approved):

PR #1235, opened 2026-05-28
Rules update and SA fixes

aperez-worklytics:
- Update some rules and made connector async
- Fix an issue when a new GCP connector is added with SA

 - []()

### Features
[Psoxy support](https://app.asana.com/1/27145998307022/task/1214109202786427)

 ### Logistics
 - []()

### Change implications

     obviously a no-op?
     change

---
  - https://app.asana.com/0/0/1214109202786427


infra/modules/gcp-proxy-api/main.tf
```diff
@@ -200,19 +200,14 @@ locals {
   }
 }
 
-data "google_service_account" "function" {
-  account_id = var.service_account_email
```
aperez-worklytics, line 204: That was causing me an error in GCP during plan. It seems like TF uses this during plan phase, so if it doesn't exist (which is the case when a new connector is added) it will complain with an error related to lack of permissions or *it does not exist*

docs/sources/chatgpt-enterprise/example-api-responses/original/sample_log_message.json
```diff
@@ -0,0 +1,51 @@
...
+    "id": "msg_xyz789abc123",
+    "parent_id": null,
+    "created_at": "2026-05-20T12:47:09.501218Z",
+    "author": {
+      "type": "user",
+      "client_type": "unknown"
+    },
+    "content": {
+      "type": "user_instructions",
+      "value": "Can you summarize the Q2 financial results for the EMEA region?"
```
eschultink, line 24: why no textDigest or or similar on this?

docs/sources/chatgpt-enterprise/example-api-responses/sanitized/sample_log_message.json
```diff
@@ -0,0 +1,47 @@
...
+      {
+        "id":"ann_001",
+        "type":"citation",
+        "quote_text":""
+      }
+    ]
+  },
+  "conversation":{
+    "id":"conv_def456ghi789",
+    "title":"",
```
eschultink: take this to `null` or drop entirely? imho that's cleaner than `""`

docs/sources/chatgpt-enterprise/example-api-responses/sanitized/sample_log_message.json
```diff
@@ -0,0 +1,47 @@
...
+  "message":{
+    "id":"msg_xyz789abc123",
+    "parent_id":null,
+    "author":{
+      "type":"user",
+      "client_type":"unknown"
+    },
+    "content":{
+      "type":"user_instructions",
+      "value":"summarize"
```
eschultink: i think we need textDigest - were doing it in the old API.

eschultink (changes requested):
i think we need to do textDigest on `content.value`, to maintain functionality that we currently have.

docs/sources/chatgpt-enterprise/chatgpt-enterprise.yaml
```diff
@@ -171,380 +18,173 @@ endpoints:
...
         jsonPaths:
-          - "$.data[*].user_email"
+          - "$.principal.id"
+          - "$..email"
+          - "$..user_id"
+          - "$..user_email"
         encoding: "URL_SAFE_TOKEN"
-      - !<textDigest>
-        keywords:
+      - !<redactExceptPhrases>
```
jlorper: why the keywords are gone???
this won't work on Worklytics' side as it is right now.
Also seems that the list of keywords is different?? Why that change?

jlorper: nevermind... now seems good, the diff is a mess. I saw stuff like "blog post" or "grammar" being removed and thought you were changing the original keywords... but turns out those where not in the list, so I assume you put the right ones.

docs/sources/chatgpt-enterprise/example-api-responses/sanitized/sample_log_message.json
jlorper: yes, text digest + keywords.

docs/sources/chatgpt-enterprise/example-api-responses/sanitized/sample_log_message.json
jlorper: I guess that's the redactExcept outcome... but it should be textDigest + keywords

docs/sources/chatgpt-enterprise/example-api-responses/sanitized/sample_log_message.json
aperez-worklytics: ok; see https://github.com/Worklytics/evalengin/pull/8498#discussion_r3324519790

docs/sources/chatgpt-enterprise/chatgpt-enterprise.yaml
aperez-worklytics: I put what we have as generic. I think we can do a combination between the old ones and the generic ones. See https://github.com/Worklytics/evalengin/pull/8498#discussion_r3324519790

jlorper (approved):

docs/sources/chatgpt-enterprise/chatgpt-enterprise.yaml
jlorper: we should use just GenericAIToolsSource.AI_KEYWORDS

eschultink (approved):

PR #1237, opened 2026-05-28
fix NPE when content is null

aperez-worklytics:
When content is null, during converting bytes to String it was failing, causing psoxy to throw a NPE and an error to the caller.

### Fixes
[NPE when handling response with null content](https://app.asana.com/1/27145998307022/project/1201039336231823/task/1215192487734411)

 - []()

 ### Logistics
 - []()

### Change implications

     obviously a no-op?
     change

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/ProcessedContent.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/ProcessedContent.java
@@ -60,7 +60,7 @@ public class ProcessedContent implements Serializable {
      */
     public String getContentAsString() {
         if (getContent() == null) {
-            return "";
+            return null;
         } else {
             return new String(getContent(), contentCharset);
         }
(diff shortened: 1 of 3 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/ProcessedContent.java
```diff
@@ -59,6 +59,10 @@ public class ProcessedContent implements Serializable {
      * @return the content as a string, using the specified contentCharset
      */
     public String getContentAsString() {
-        return new String(getContent(), contentCharset);
+        if (getContent() == null) {
+            return null;
+        } else {
+            return new String(getContent(), contentCharset);
```
aperez-worklytics, line 65: `new String(null, contentCharset` produces a NPE:

```
public String(byte[] bytes, Charset charset) {
        this(Objects.requireNonNull(charset), bytes, 0, bytes.length);
    }
```

eschultink (approved):

PR #1251, opened 2026-06-01
avoid bucket write to caller in testing case

eschultink:
### Fixes
- Updated permission logic to grant bulk test input writes to `caller_aws_arns` instead of the `Caller` role, enhancing security and flexibility in access management.

### Change implications

  - No major changes expected in `CHANGELOG.md` or `terraform plan`/`apply`.
  - breaking changes? No breaking changes anticipated.

---
  - https://app.asana.com/0/0/1204918934280708

### DIFF
--- a/infra/examples-dev/aws/kms-cmek.tf
+++ b/infra/examples-dev/aws/kms-cmek.tf
@@ -18,7 +18,7 @@
 #      "Sid": "Allow Test Principals to Encrypt for Input Upload",
 #      "Effect": "Allow",
 #      "Principal": {
-#        "AWS": var.caller_aws_arns
+#        "AWS": coalesce(var.test_aws_principal_arns, [])
 #      },
 #      "Action": [
 #        "kms:Encrypt",
--- a/infra/examples-dev/aws/kms-cmek.tf
+++ b/infra/examples-dev/aws/kms-cmek.tf
@@ -13,15 +13,17 @@
 #locals {
 #  key_arn = aws_kms_key.example_key.arn # alternatively, use ar.project_aws_kms_key_arn
 #
-#  # TODO: can eliminate this if test tool doesn't assume role when uploading to bucket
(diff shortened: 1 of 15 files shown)


davidfq (approved):

infra/modules/aws-proxy-bulk/main.tf
```diff
@@ -300,11 +300,11 @@ resource "aws_ssm_parameter" "rules" {
...
 }
 
-resource "aws_iam_policy" "testing" {
+resource "aws_iam_policy" "testing_input_write" {
   count = var.provision_iam_policy_for_testing ? 1 : 0
 
-  name_prefix = "${local.iam_policy_prefix}Testing"
-  description = "Allow to write to input bucket, read from sanitized bucket to test Lambda's behavior"
+  name_prefix = "${local.iam_policy_prefix}TestingInputWrite"
+  description = "Allow to write to input bucket for testing"
```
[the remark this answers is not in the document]
eschultink: yeah - i think this is valid feedback; create such a variable, default to the user running tf. ensure IAM policy *potentially* works cross account (expecting common case to be a root account user writing into buckets in sub account)

infra/examples-dev/aws/kms-cmek.tf
```diff
@@ -18,7 +18,7 @@
 #      "Sid": "Allow Test Principals to Encrypt for Input Upload",
 #      "Effect": "Allow",
 #      "Principal": {
-#        "AWS": var.caller_aws_arns
+#        "AWS": coalesce(var.test_aws_principal_arns, [])
```
eschultink: does AWS event permit empty list??

aperez-worklytics: From https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html I understand that not, you could have one or many, but least it should be one

aperez-worklytics (approved):

PR #1262, opened 2026-06-04
handler `Publisher` shutdown

jlorper:
### Fixes
[Handle publisher shutdown](https://app.asana.com/1/27145998307022/project/1215080811397652/task/1215405974976566)

### Change implications

 - something important to note in future release notes? NO
     obviously a no-op?
     change


java/impl/gcp/src/main/java/co/worklytics/psoxy/ApiDataRequestViaPubSub.java
```diff
@@ -33,14 +35,25 @@ public class ApiDataRequestViaPubSub implements AsyncApiDataRequestHandler {
...
     ObjectMapper objectMapper;
     @Inject
     EnvVarsConfigService envVarsConfigService;
 
+    @SneakyThrows
     @AssistedInject
     public ApiDataRequestViaPubSub(@Assisted @NonNull String topicName) {
         this.topicName = topicName;
+        this.publisher = Publisher.newBuilder(TopicName.parse(topicName)).build();
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
```
jlorper, line 50: I believe this is a Singleton, so doing this makes sense.

java/impl/gcp/src/main/java/co/worklytics/psoxy/ApiDataRequestViaPubSub.java
```diff
@@ -33,14 +35,25 @@ public class ApiDataRequestViaPubSub implements AsyncApiDataRequestHandler {
...
     @Inject
     ObjectMapper objectMapper;
     @Inject
     EnvVarsConfigService envVarsConfigService;
 
+    @SneakyThrows
     @AssistedInject
     public ApiDataRequestViaPubSub(@Assisted @NonNull String topicName) {
         this.topicName = topicName;
+        this.publisher = Publisher.newBuilder(TopicName.parse(topicName)).build();
```
eschultink, line 49: imho, question if this should be done via Dagger/CI; in effect you're constructing a dependency of this class w/i its constructor - issues with that is 1) complicating testing (no easy way to mock Publisher or something), 2) constructor is heavier / more brittle.

That said, not so straightforward to add into dagger ... as in effect it's. a parameterized constructor --> requiring AssistInject/AssistedFactory ... so lot of stuff to wire just to gain that.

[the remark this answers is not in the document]
eschultink: this doesn't bother me

eschultink (approved):
i think it's fine; would be most correct to rely on Dagger to provide instance of `Publisher`

PR #1278, opened 2026-06-09
Adding property config for redirect

aperez-worklytics:
Instead of relying into the async case, just supporting have this as a env var, so we can setup without doing major changes.

 - []()

### Features
[Setting for handling redirect](https://app.asana.com/1/27145998307022/project/1213797956438900/task/1215510682641177)

 ### Logistics
 - []()

### Change implications

     obviously a no-op? **no**
     change


eschultink (approved):
OK; agree it's clearer.

Mixed feelings on putting it on spec via `environment_variables`; that's flexible, as easy to keep backwards compatible w/o a strict schema for the specs.  But hurts readability/clarity as to what the defaults are.

PR #1279, opened 2026-06-09
Deps maintenance

jlorper:
### Fixes
- Packages maintenance to eliminate flagged vulnerabilities
- [aircompressor Snappy and LZ4 Java-based decompressor implementation can leak information from reused output buffer ](https://www.cve.org/CVERecord?id=CVE-2025-67721)
- [Apache OpenNLP ExtensionLoader Vulnerable to Arbitrary Class Instantiation via Model Manifest](https://www.cve.org/CVERecord?id=CVE-2026-42027)
- [Apache OpenNLP DictionaryEntryPersistor Vulnerable to XML External Entity (XXE) via Unsanitized Dictionary Parsing](https://www.cve.org/CVERecord?id=CVE-2026-40682)
- [Apache OpenNLP AbstractModelReader has an OOM Denial of Service via Unbounded Array Allocation](https://www.cve.org/CVERecord?id=CVE-2026-42440)

### Change implications

 - dependencies added/changed? **yes, updates to versions of aircompressor, parquet-floor and opennlp that remove security vulnerabilites**
     obviously a no-op? **no**
     change - **no**

### DIFF
--- a/java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/SentenceMetadataProcessor.java
+++ b/java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/SentenceMetadataProcessor.java
@@ -11,6 +11,7 @@ import com.avaulta.gateway.rules.augments.SentenceMetadataResult.Verb;
 import opennlp.tools.chunker.ChunkerME;
 import opennlp.tools.chunker.ChunkerModel;
 import opennlp.tools.postag.POSModel;
+import opennlp.tools.postag.POSTagFormat;
 import opennlp.tools.postag.POSTaggerME;
 import opennlp.tools.sentdetect.SentenceDetectorME;
 import opennlp.tools.sentdetect.SentenceModel;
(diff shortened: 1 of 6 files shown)


java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/SentenceMetadataProcessor.java
```diff
@@ -289,7 +290,7 @@ private void initializeModels() {
                  InputStream posModelStream = loadModel("en-pos-maxent.bin");
                  InputStream chunkerModelStream = loadModel("en-chunker.bin")) {
                 SentenceDetectorME detector = new SentenceDetectorME(new SentenceModel(sentenceModelStream));
-                POSTaggerME tagger = new POSTaggerME(new POSModel(posModelStream));
+                POSTaggerME tagger = new POSTaggerME(new POSModel(posModelStream), POSTagFormat.PENN);
```
eschultink, line 293: 🤔 not sure relation here ...

eschultink (approved):

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/SentenceMetadataProcessor.java
jlorper: was needed, otherwise a test failed on CI... I don't get why in local it was working, maybe something with the environment is picking up a different format for the tags.
Instead of NN it seems it was using NOUN (Universal Dependencies), so enforcing to Penn made the test work.

PR #1298, opened 2026-06-16
Handling transient cases when reading a secret

aperez-worklytics:
Trying to fix issue that happening in AWS when there are requests after a few and then the instance starts returning CONNECTION_SETUP. 

I think the issue could come about how the `ParameterStore` /  `SecretsManagerSecretStore` in AWS. Its client is marked as `Singleton` so the instance is the same across all the lifetime of the function. As a difference with GCP, where even `SecretManagerConfigService` is marked as singleton, the GCP client is created on each request:

https://github.com/Worklytics/psoxy/blob/82542de64aaeb7cd4fc23e42e42d29155ecf17e4/java/impl/gcp/src/main/java/co/worklytics/psoxy/SecretManagerConfigService.java#L93

What I think it is happening in AWS is that after a while (it should match less han an hour of running time) is that the internal token authentication between PS / SMS it is expired, waiting some requests until this is obtained a new fresh one. That matches what I see in the caller side where after some requests returning CONNECTION_SETUP during 1-3 seconds, then the function is working well after that.

So what this PR is doing is mostly:
- trying to enforce reload on cache
- detect access denied vs other error, considering it as transient, after couple of retries it will return 503 if still not working

Ideas are welcomed here

### Fixes
[Suddenly CONNECTION_SETUP error is killing the reader](https://app.asana.com/1/27145998307022/project/1203776123599466/task/1215641372342099)

 - []()

 ### Logistics
 - []()

### Change implications

     obviously a no-op?
     change

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CachingConfigServiceDecorator.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CachingConfigServiceDecorator.java
@@ -9,21 +9,39 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Ticker;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
 import co.worklytics.psoxy.gateway.ConfigService;
 import co.worklytics.psoxy.gateway.SecretStore;
 import co.worklytics.psoxy.gateway.WritableConfigService;
-import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
+import lombok.extern.java.Log;
+import org.jspecify.annotations.NonNull;
+
+import java.util.logging.Level;
 
 
-@RequiredArgsConstructor
+@Log
 public class CachingConfigServiceDecorator implements WritableConfigService, SecretStore {
 
     final ConfigService delegate;
     final Duration defaultTtl;
+    final Ticker ticker;
+
+    public CachingConfigServiceDecorator(ConfigService delegate, Duration defaultTtl) {
+        this(delegate, defaultTtl, Ticker.systemTicker());
+    }
+
+    @VisibleForTesting
+    CachingConfigServiceDecorator(ConfigService delegate, Duration defaultTtl, Ticker ticker) {
+        this.delegate = delegate;
+        this.defaultTtl = defaultTtl;
+        this.ticker = ticker;
+    }
 
     private volatile LoadingCache<ConfigProperty, String> cache;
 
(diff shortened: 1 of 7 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CachingConfigServiceDecorator.java
```diff
@@ -9,21 +9,51 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 import co.worklytics.psoxy.gateway.ConfigService;
 import co.worklytics.psoxy.gateway.SecretStore;
+import co.worklytics.psoxy.gateway.TransientConfigException;
 import co.worklytics.psoxy.gateway.WritableConfigService;
-import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
+import lombok.extern.java.Log;
+import org.jspecify.annotations.NonNull;
```
eschultink: i don't think this is the `NonNull` we us anywhere else.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CachingConfigServiceDecorator.java
```diff
@@ -39,12 +69,58 @@ LoadingCache<ConfigProperty, String> getCache() {
                 if (this.cache == null) {
                     this.cache = CacheBuilder.newBuilder()
                         .maximumSize(100)
-                        .expireAfterWrite(defaultTtl.getSeconds(), TimeUnit.SECONDS)
+                        .ticker(ticker)
+                        .refreshAfterWrite(defaultTtl.getSeconds(), TimeUnit.SECONDS)
```
eschultink, line 74: agree this is better.

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CachingConfigServiceDecorator.java
```diff
@@ -39,12 +69,58 @@ LoadingCache<ConfigProperty, String> getCache() {
...
+                                    }
+                                    try {
+                                        return delegate.getConfigPropertyAsOptional(key).orElse(NEGATIVE_VALUE);
+                                    } catch (TransientConfigException e) {
+                                        lastException = e;
+                                        log.log(Level.WARNING, "Transient failure on attempt {0}/{1} for config property {2}",
+                                            new Object[]{attempt + 1, MAX_TRANSIENT_RETRIES, key.name()});
+                                    }
+                                }
+                                throw Objects.requireNonNull(lastException);
```
eschultink: pointelss, right?

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/CachingConfigServiceDecorator.java
```diff
@@ -39,12 +69,58 @@ LoadingCache<ConfigProperty, String> getCache() {
...
+                        .refreshAfterWrite(defaultTtl.getSeconds(), TimeUnit.SECONDS)
                         .recordStats()
                         .build(new CacheLoader<ConfigProperty, String>() {  //req for java8-backwards compatibility
                             @Override
-                            public String load(ConfigProperty key) {
-                                return delegate.getConfigPropertyAsOptional(key).orElse(NEGATIVE_VALUE);
+                            public String load(@NonNull ConfigProperty key) {
+                                TransientConfigException lastException = null;
+                                for (int attempt = 0; attempt < MAX_TRANSIENT_RETRIES; attempt++) {
+                                    if (attempt > 0) {
```
eschultink: don't really get why don't get rid of this condition, move the sleep to end of loop instead of beginning.

eschultink (changes requested):
OK, seems reasonable ... appreciate the tests - as yes, I'm worried about edge cases here but OK.

minor changes requested.

aperez-worklytics (commented):
Ok, feedback applied

eschultink (approved):

PR #1336, opened 2026-06-30
fix possible null body side outputs

eschultink:
### Fixes
- incorrect handling of side-output writes when the response body is null, for GCS, S3, and gzip-wrapped outputs.

### Change implications

   - NOTE in `CHANGELOG.md` anything that will show up in `terraform plan`/`apply` that isn't obviously a no-op? **No** — runtime behavior fix only
   - breaking changes? **No** — bug fix; no API or module contract changes expected

### DIFF
--- a/java/impl/aws/src/main/java/co/worklytics/psoxy/aws/S3Output.java
+++ b/java/impl/aws/src/main/java/co/worklytics/psoxy/aws/S3Output.java
@@ -49,7 +49,7 @@ public class S3Output implements Output {
 
     @Override
     public void write(String key, ProcessedContent content) throws WriteFailure {
-        byte[] body = content.getContent() != null ? content.getContent() : new byte[0];
+        byte[] body = content.getContent();
 
         if (key == null) {
             key = DigestUtils.md5Hex(body);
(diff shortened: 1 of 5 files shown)


java/impl/aws/src/main/java/co/worklytics/psoxy/aws/S3Output.java
```diff
@@ -49,9 +49,10 @@ public S3Output(@Assisted OutputLocation location) {
 
     @Override
     public void write(String key, ProcessedContent content) throws WriteFailure {
+        byte[] body = content.getContent() != null ? content.getContent() : new byte[0];
```
aperez-worklytics: As we have the same, probably I'd do that in `ProcessedContent` with `getContent()` -so instead of using a simple get method, we could move that logic

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/S3Output.java
```diff
@@ -83,7 +84,8 @@ public void write(String key, ProcessedContent content) throws WriteFailure {
     public void write(ProcessedContent content) throws WriteFailure {
         // Generate a canonical key based on the content's hash
         // random UUID better??
-        String key = DigestUtils.sha256Hex(content.getContent());
+        byte[] body = content.getContent() != null ? content.getContent() : new byte[0];
```
aperez-worklytics: Yeah, same thing more than twice...

aperez-worklytics (approved):
Approved to not be a blocker, but Id consider my suggestion

PR #1338, opened 2026-07-01
GCP bulk mem control

eschultink:
### Features
- Default GCP bulk connector deployments to 1024MB memory with automatic CPU pairing so bulk workloads get a sensible baseline without manual sizing.
- allow configuration of bulk mem + CPU

### Change implications

   - NOTE in `CHANGELOG.md`: existing GCP bulk connector deployments may show memory (and possibly CPU) updates on `terraform plan`/`apply` when adopting the new defaults.
   - breaking changes? **no** (default-only change; may trigger in-place resource updates, not a module API break)

---
  - https://app.asana.com/0/0/1216130122059351

### DIFF
--- a/infra/modules/gcp-proxy-bulk/main.tf
+++ b/infra/modules/gcp-proxy-bulk/main.tf
@@ -19,8 +19,6 @@ locals {
     local.resolved_memory_mb <= 24576 ? "6" :
     "8"
   )
-
-  resolved_available_cpu = coalesce(var.available_cpu, local.auto_available_cpu)
 }
 
 # computed
@@ -219,24 +217,11 @@ removed {
   }
 }
 
-resource "terraform_data" "bulk_service_config" {
-  input = {
-    available_memory_mb = local.resolved_memory_mb
-    available_cpu       = local.resolved_available_cpu
-  }
-}
-
 resource "google_cloudfunctions2_function" "function" {
   name        = local.function_name
   description = "Psoxy instance to process ${var.source_kind} files"
   project     = var.gcp_project.project_id
   location    = var.region
-
-  lifecycle {
-    replace_triggered_by = [
-      terraform_data.bulk_service_config
-    ]
-  }
 
   build_config {
     runtime         = "java25"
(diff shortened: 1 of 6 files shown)


infra/modules/gcp-proxy-bulk/main.tf
```diff
@@ -201,12 +219,25 @@ removed {
...
   name        = local.function_name
   description = "Psoxy instance to process ${var.source_kind} files"
   project     = var.gcp_project.project_id
   location    = var.region
 
+  lifecycle {
+    replace_triggered_by = [
+      terraform_data.bulk_service_config
+    ]
+  }
```
eschultink: this really necessary? is changing available_cpu on `service_config` enough?

infra/modules/gcp-proxy-bulk/main.tf
```diff
@@ -225,7 +256,8 @@ resource "google_cloudfunctions2_function" "function" {
   # NOTE: bulk connectors are STILL triggered through HTTPS invocations, so these have default HTTPS endpoint URls
 
   service_config {
-    available_memory      = "${coalesce(var.available_memory_mb, 1024)}M"
+    available_memory      = "${local.resolved_memory_mb}M"
+    available_cpu         = local.resolved_available_cpu
```
eschultink: original problem is that if you change `available_memory` only, google provider tries to update the function rather than replacing it, but this fails if it's too much mem for the reserved CPU.  Explicit replace solves it; not sure if explicitly setting and bumping available_cpu will solve it.

infra/modules/gcp-proxy-bulk/variables.tf
```diff
@@ -97,10 +97,18 @@ variable "bucket_write_role_id" {
...
+  # google_cloudfunctions2_function.service_config.available_memory (e.g. "1024M"), consistent with available_cpu.
   description = "Memory (in MB), available to the function. Default value is 1024. Possible values include 128, 256, 512, 1024, 2048, 4096; above that requires multiple CPUs, beyond scope of our built-in configurations."
   default     = 1024
 }
 
+variable "available_cpu" {
+  type        = string
+  description = "vCPUs for the function (e.g. '0.333', '0.5', '1'). When null, a minimum CPU for the configured memory is chosen automatically per GCP limits. If you set both available_cpu and available_memory_mb, both values are applied as given (not validated against GCP limits)."
+  default     = null
+}
```
eschultink: i dont' see value in exposing explicit configurable CPU atm; just expose mem and grab min CPU for each amount of mem.

aperez-worklytics (approved):

PR #1356, opened 2026-07-24
OneDrive connector

aperez-worklytics:
### Features

- Adding support for OneDrive connector (rules, sample files, deep linking, docs, etc). Generated based on https://github.com/Worklytics/evalengin/pull/8810
- [Psoxy update](https://app.asana.com/1/27145998307022/task/1216794006228765)

### Change implications

     obviously a no-op? **no**
     change

---
  - https://app.asana.com/0/0/1216794006228765

### DIFF
--- a/docs/sources/microsoft-365/msft-onedrive/msft-onedrive.yaml
+++ b/docs/sources/microsoft-365/msft-onedrive/msft-onedrive.yaml
@@ -1,0 +1,571 @@
+---
+endpoints:
+  - pathTemplate: "/v1.0/drives/{driveId}/activities"
+    allowedQueryParams:
+      - "$expand"
+      - "$skiptoken"
+    transforms:
+      - !<pseudonymize>
+        jsonPaths:
+          - "$..user.email"
+        encoding: "URL_SAFE_TOKEN"
+    responseSchema:
+      type: "object"
+      properties:
+        '@microsoft.graph.hasMoreData':
+          type: "boolean"
+        '@odata.deltaLink':
+          type: "string"
+        '@odata.nextLink':
+          type: "string"
+        value:
+          type: "array"
+          items:
+            $ref: "#/definitions/ItemActivity"
+  - pathTemplate: "/v1.0/drives/{driveId}/items/{itemId}/activities"
+    allowedQueryParams:
+      - "$expand"
+      - "$skiptoken"
+    transforms:
+      - !<pseudonymize>
+        jsonPaths:
+          - "$..user.email"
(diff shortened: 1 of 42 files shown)


docs/sources/microsoft-365/msft-onedrive/msft-onedrive.yaml
```diff
@@ -0,0 +1,571 @@
...
+          type: "array"
+          items:
+            $ref: "#/definitions/DriveItem"
+  - pathTemplate: "/v1.0/groups"
+    allowedQueryParams:
+      - "$top"
+      - "$select"
+      - "$skiptoken"
+      - "$orderBy"
+      - "$count"
```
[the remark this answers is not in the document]
eschultink: i prefer to make the checking case insensitive. will deal with that on a separate branch.

eschultink (approved):
i agree with the hardwraps in md; and casing in example urls (We should match MSFT docs)

docs/sources/microsoft-365/msft-onedrive/msft-onedrive.yaml
```diff
@@ -0,0 +1,571 @@
...
+        value:
+          type: "array"
+          items:
+            $ref: "#/definitions/Drive"
+  - pathTemplate: "/v1.0/users"
+    allowedQueryParams:
+      - "$top"
+      - "$select"
+      - "$skiptoken"
+      - "$orderby"
```
aperez-worklytics, line 126: Now matching the value that appears in https://learn.microsoft.com/en-us/graph/query-parameters?tabs=http#odata-system-query-options

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -351,8 +352,15 @@ String effectiveRegex(Endpoint endpoint) {
 
     boolean allowedQueryParams(Endpoint endpoint, List<Pair<String, String>> queryParams) {
         boolean matchesAllowed = endpoint.getAllowedQueryParamsOptional()
-                .map(allowedParams -> allowedParams.containsAll(
-                        queryParams.stream().map(Pair::getKey).collect(Collectors.toList())))
+                .map(allowedParams -> {
+                    Set<String> allowedParamsLowerCase = allowedParams.stream()
```
aperez-worklytics, line 356: Adding this change to support ignoring case when checking query parameters

aperez-worklytics (commented):
Some improvements in docs and adding support for ignoring case when matching query parameters

PR #1361, opened 2026-07-30
request headers to override path/query 

eschultink:
Adds support for request headers that override the proxied path and query string, enabling fallback routing or request shaping without changing the underlying connector configuration.

Use case: GCP does some decoding of path / query string values we don't want; as well as weird routing with //.  have had similar issues with AWS lambda/api gateway, although could more reliably work around those; still worth having feature everywhere in case needed.

### Features
 - `x-Psoxy-TargetPath` header - overrides path that request will be sent to in source
 - `x-Psoxy-TargetQuery` header - overrides querystring that will be sent with request to source

### Change implications

   - NOTE in `CHANGELOG.md` anything that will show up in `terraform plan`/`apply` that isn't obviously a no-op?
     - Unlikely unless new env vars or Terraform inputs were added for the override headers (not evident from the commit message alone).
     - **No** — this appears to be additive behavior (new override headers), not a change to existing defaults.

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/TargetOverrideRequestResolver.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/TargetOverrideRequestResolver.java
@@ -74,11 +74,12 @@ public class TargetOverrideRequestResolver {
     }
 
     String validateTargetPath(String path) {
-        if (StringUtils.isEmpty(path)) {
+        if (StringUtils.isBlank(path)) {
             throw new IllegalArgumentException(
                 ControlHeader.TARGET_PATH.getHttpHeader() + " must not be empty");
         }
-        validateUntrustedHeaderValue(ControlHeader.TARGET_PATH.getHttpHeader(), path);
+        // Path must not include query/fragment delimiters — those belong in TargetQuery / nowhere.
+        validateUntrustedHeaderValue(ControlHeader.TARGET_PATH.getHttpHeader(), path, false);
         if (!path.startsWith("/")) {
             throw new IllegalArgumentException(
                 ControlHeader.TARGET_PATH.getHttpHeader() + " must start with '/'");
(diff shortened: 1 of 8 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/TargetOverrideRequestResolver.java
```diff
@@ -0,0 +1,187 @@
...
+                effectiveQuery.orElse(""),
+                ControlHeader.TARGET_QUERY.getHttpHeader(),
+                request.getQuery().orElse("")));
+        }
+
+        return new HttpEventRequestWithPathQueryOverrides(request, effectivePath, effectiveQuery);
+    }
+
+    String validateTargetPath(String path) {
+        if (StringUtils.isEmpty(path)) {
```
aperez-worklytics:
```suggestion
        if (StringUtils.isBlank(path)) {
```

java/core/src/main/java/co/worklytics/psoxy/gateway/TargetOverrideRequestResolver.java
```diff
@@ -0,0 +1,187 @@
...
+        // empty string is a valid override meaning "no query"
+        if (StringUtils.isNotEmpty(normalized)) {
+            validateUntrustedHeaderValue(ControlHeader.TARGET_QUERY.getHttpHeader(), normalized);
+        }
+        return normalized;
+    }
+
+    void validateUntrustedHeaderValue(String headerName, String value) {
+        if (value.length() > MAX_HEADER_VALUE_LENGTH) {
+            throw new IllegalArgumentException(
```
aperez-worklytics, line 110: ok with the exception, but that should return a 400 to the client

aperez-worklytics (approved):

PR #1371, opened 2026-08-04
Onedrive: Rules update for including reversible

aperez-worklytics:
### Fixes
 - `no_app_ids` didnt't mark ids for user as reversible in rules.

### Change implications

     obviously a no-op?
     change


eschultink (approved):
arg should use the pathParam schemas to REQUIRE reversibles (eg, not except ids in the clear for these).

I think something like:

```yaml
pathParameterSchemas:
  userId:
    type: "string"
    format: "reversible-pseudonym"
```

PR #1376, opened 2026-08-13
S231 : more alb

eschultink:
### Features
- use ALB host in GCP TODOs when an external load balancer is enabled
- use ALB host in proxy endpoint outputs when an external load balancer is enabled

### Change implications

  - NOTE in `CHANGELOG.md` anything that will show up in `terraform plan`/`apply` that isn't obviously a no-op?
    - **yes** — GCP TODO and proxy endpoint outputs may change when an external load balancer is enabled (ALB host instead of prior values)
    - **no** — output value changes only when external load balancer is enabled; behavior is more accurate for that configuration


infra/modules/gcp-proxy-api/main.tf
```diff
@@ -581,14 +581,20 @@ output "cloud_function_name" {
 }
 
 output "cloud_function_url" {
-  value = local.cloud_function_url
+  description = "Direct Cloud Function / Cloud Run URI (*.run.app). Pub/Sub push and internal callers use this; internet clients should use endpoint_url when an external ALB is enabled."
```
eschultink, line 584: as `aud` param?

and how about *can* use it, as an alternative to adding external ALB route as a Custom Audience in the cloud function.

infra/modules/gcp-proxy-api/main.tf
```diff
@@ -581,14 +581,20 @@ output "cloud_function_name" {
...
 }
 
 output "proxy_endpoint_url" {
-  description = "Public URL used for tests / TODOs (per-connector path on external_lb_base_url when set, otherwise Cloud Function URI)."
+  description = "Public URL used for tests / TODOs (per-connector path on external_lb_base_url when set, otherwise Cloud Function URI). No trailing slash."
   value       = local.proxy_endpoint_url
 }
 
+output "endpoint_url" {
+  description = "Public proxy base URL with trailing slash (ALB https://<host>/<function-name>/ when external_lb_base_url is set, otherwise the Cloud Function URI). Use this in Worklytics connection TODOs and terraform outputs."
```
eschultink, line 594:
```suggestion
  description = "Public proxy base URL with trailing slash (ALB https://<host>/<function-name>/ when external_lb_base_url is set, otherwise the Cloud Function URI)."
```

infra/modules/gcp-proxy-api/main.tf
```diff
@@ -581,14 +581,20 @@ output "cloud_function_name" {
...
+  value       = local.cloud_function_url
 }
 
 output "proxy_endpoint_url" {
-  description = "Public URL used for tests / TODOs (per-connector path on external_lb_base_url when set, otherwise Cloud Function URI)."
+  description = "Public URL used for tests / TODOs (per-connector path on external_lb_base_url when set, otherwise Cloud Function URI). No trailing slash."
   value       = local.proxy_endpoint_url
 }
 
+output "endpoint_url" {
```
eschultink, line 593: what's the purpose of this!?!? why bother, totally duplicatuve w `proxy_endpoint_url`; use that, and let ppl add `/` if/when they want

KISS/DRY!!!

infra/modules/gcp-host/main.tf
```diff
@@ -360,18 +356,36 @@ module "external_api_alb" {
...
+${local.alb_dns_setup_body}
+EOT
+}
+
+resource "local_file" "todo_alb_dns_setup" {
+  count = var.todos_as_local_files && local.alb_dns_todo != null ? 1 : 0
+
+  filename = "TODO ${var.todo_step} - configure DNS for API connector load balancer.md"
+  content  = local.alb_dns_todo
+}
```
eschultink, line 391: avoid this; move to our top-level example - conditionally creating a TODO file for this IF todo output came out of gcp-host. This let's customers who ban/block `local_file` entirely still use the modules as-is (having disabled via count conditions not acceptable) - we have to move to this pattern long-term generally.

eschultink (commented):
OK; so looks like TODO files actually correct? and I was looking at stuff generated in 0.6.8 or something that were a prlbme.

jlorper (approved):

PR #1378, opened 2026-08-14
Add Zoom meeting-summary finder; drop unused /settings example

eschultink:
## Summary
- Adds `tools/psoxy-test/find-zoom-meeting-summary.js` and documents it in the testing guide and Zoom example API calls. `GET /v2/meetings/{id}/meeting_summary` needs a past instance UUID with an AI Companion summary, not a scheduled meeting number.
- Removes `/v2/users/{USER_ID}/settings` from Zoom `example_api_calls`. That path is not allowed by proxy rules, so generated `test-zoom.sh` should not advertise it.

## Test plan
- [x] From a Zoom deploy directory, run `find-zoom-meeting-summary.js` with the same `-u` / `-f` / `--allow-insecure-tls` (or `-r`) flags as `test-zoom.sh` and confirm it walks users → meetings → instances.
- [x] Confirm generated `test-zoom.sh` help no longer lists `GET '/v2/users/{USER_ID}/settings'` after a Terraform refresh of the test scripts.
- [x] Confirm `GET /v2/users/{id}/settings` is still blocked by Zoom rules (not in `zoom.yaml`).

Made with [Cursor](https://cursor.com)


infra/modules/worklytics-connector-specs/main.tf
```diff
@@ -922,7 +922,6 @@ EOT
       example_api_calls : [
         "/v2/users",
         "/v2/users/{USER_ID}/meetings",
-        "/v2/users/{USER_ID}/settings",
```
aperez-worklytics, line 925: 👍

tools/psoxy-test/find-zoom-meeting-summary.js
```diff
@@ -0,0 +1,283 @@
...
+          userId,
+          meetingId,
+          meetingStartTime: meeting.start_time || meeting.startTime,
+          instanceUuid: uuid,
+          instanceStartTime: instance.start_time,
+          pastMeeting: details.data,
+          meetingSummaryStatus: summary.status,
+          meetingSummary: summary.data,
+          testCall: `./test-zoom.sh GET '/v2/meetings/${encodedUuid}/meeting_summary'`,
+        }, null, 2));
```
aperez-worklytics, line 266: I'd only log when 200 to avoid noise until the summary is found; in theory that should work on first attempt but who knows...

docs/sources/zoom/example-api-calls.md
```diff
@@ -56,3 +56,26 @@ node tools/psoxy-test/cli-call.js -u [your_psoxy_url]/v2/past_meetings/[zoom_mee
...
+```
+
+Once you have an instance UUID (`[zoom_meeting_uuid]`), you can also call:
+
+```shell
+node tools/psoxy-test/cli-call.js -u [your_psoxy_url]/v2/past_meetings/[zoom_meeting_uuid]
+node tools/psoxy-test/cli-call.js -u [your_psoxy_url]/v2/meetings/[zoom_meeting_uuid]/meeting_summary
+```
+
+If the UUID starts with `/` or contains `//`, [double-encode](https://developers.zoom.us/docs/api/rest/using-zoom-apis/#meeting-id-and-uuid) it in the path. Smart Recording summaries in the Zoom UI are a different feature; this endpoint only returns AI Companion **meeting** summaries.
```
aperez-worklytics: Bit confused here; that's done as part of the `find-zoom-meeting-summary.js` execution. It could lead to say like the script is going to return a `zoom_meeting_uuid` and you can also use that value manually.

I'd just include something like *If you don't want to use the script and provide an instance UUID, you can check with ...*

aperez-worklytics (approved):

PR #1401, opened 2026-08-27
MSFT rules update

aperez-worklytics:
IN DRAFT UNTIL FURTHER TESTING

Parent PR is #1402 

Java rules for MSFT dropped and using the ones generated from https://github.com/Worklytics/evalengin/pull/9018. For that:
- YAML files have been updated / replaced
- Added more tests to ensure all query parameters we need are covered by rules through `InvocationExamples`
- Updated spec in psoxy-test for MSFT; also supporting pagination and links between calls 

 - []()

### Features
[Psoxy update](https://app.asana.com/1/27145998307022/task/1217962080141990)

 ### Logistics
> paste links to issues/tasks in project management
 - []()

### Change implications

 - dependencies added/changed? **yes (explain) / no**
 - something important to note in future release notes?
   - NOTE in `CHANGELOG.md` anything that will show up in `terraform plan`/`apply` that isn't
     obviously a no-op?
   - breaking changes? if in module/example that is NOT marked `alpha`, requires major version
     change

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1217962080141990

PR #1402, opened 2026-08-28
OneDrive: support token in path for deltas and fix pseudonymizing delta link when listing user drives

aperez-worklytics:
From https://github.com/Worklytics/evalengin/pull/9014
- Support delta token in path, apart from query parameter
- Pseudonymizing delta link when listing user drives

 - []()

### Features
[Psoxy update](https://app.asana.com/1/27145998307022/task/1217917147563451)

 ### Logistics
 - []()

### Change implications

     obviously a no-op?
     change

---
  - https://app.asana.com/0/0/1217917147563451

### DIFF
--- a/docs/sources/microsoft-365/msft-onedrive/msft-onedrive.yaml
+++ b/docs/sources/microsoft-365/msft-onedrive/msft-onedrive.yaml
@@ -44,7 +44,7 @@ endpoints:
         type: "array"
         items:
           $ref: "#/definitions/ItemActivity"
-- pathRegex: "^/v1\\.0/drives/[^/]+/root/delta(\\(token=[^)]*\\))?(\\?.*)?$"
+- pathTemplate: "/v1.0/drives/{driveId}/root/delta"
   allowedQueryParams:
   - "token"
   - "$select"
@@ -55,6 +55,34 @@ endpoints:
     jsonPaths:
     - "$..user.email"
     encoding: "URL_SAFE_TOKEN"
+  responseSchema:
+    type: "object"
+    properties:
+      '@microsoft.graph.hasMoreData':
+        type: "boolean"
+      '@odata.deltaLink':
+        type: "string"
+      '@odata.nextLink':
+        type: "string"
+      value:
+        type: "array"
+        items:
+          $ref: "#/definitions/DriveItem"
+# OData function-call continuation shape for /root/delta, eg "/root/delta(token=abc123)"
+- pathTemplate: "/v1.0/drives/{driveId}/root/{deltaExpression}"
+  allowedQueryParams:
+  - "token"
+  - "$select"
+  - "$expand"
+  - "$top"
+  transforms:
+  - !<pseudonymize>
+    jsonPaths:
+    - "$..user.email"
+    encoding: "URL_SAFE_TOKEN"
+  pathParameterSchemas:
+    deltaExpression:
+      pattern: "^delta\\(token=[^)]+\\)$"
   responseSchema:
     type: "object"
     properties:
--- a/docs/sources/microsoft-365/msft-onedrive/msft-onedrive.yaml
+++ b/docs/sources/microsoft-365/msft-onedrive/msft-onedrive.yaml
@@ -44,7 +44,7 @@ endpoints:
         type: "array"
         items:
           $ref: "#/definitions/ItemActivity"
-- pathTemplate: "/v1.0/drives/{driveId}/root/delta"
+- pathRegex: "^/v1\\.0/drives/[^/]+/root/delta(\\(token=[^)]*\\))?(\\?.*)?$"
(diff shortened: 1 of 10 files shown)


aperez-worklytics:
Mark draft, need to check couple of things

docs/sources/microsoft-365/msft-onedrive/msft-onedrive.yaml
```diff
@@ -44,7 +44,7 @@ endpoints:
         type: "array"
         items:
           $ref: "#/definitions/ItemActivity"
-- pathTemplate: "/v1.0/drives/{driveId}/root/delta"
+- pathRegex: "^/v1\\.0/drives/[^/]+/root/delta(\\(token=[^)]*\\))?(\\?.*)?$"
```
eschultink: i certainly think this stuff is cleaner with a `pathTemplate` than a regex if that's possible. 

I'm skeptical about your `token=` segment too; shouldn't that be AFTER an optional `?` ?

docs/sources/microsoft-365/msft-onedrive/msft-onedrive_no-app-ids.yaml
```diff
@@ -46,7 +46,7 @@ endpoints:
           type: "array"
           items:
             $ref: "#/definitions/ItemActivity"
-  - pathTemplate: "/v1.0/drives/{driveId}/root/delta"
+  - pathRegex: "^/v1\\.0/drives/[^/]+/root/delta(\\(token=[^)]*\\))?(\\?.*)?$"
```
eschultink: again; prefer template over regex as more readable and less error prone; and i'm skeptical whether  `?` is missing from this one.

I don't see why this can't be a plain regex + token as an allowed query param:

Image

java/core/src/test/java/co/worklytics/psoxy/rules/msft/OneDrive_NoAppIds_Tests.java
```diff
@@ -41,14 +41,22 @@ public Stream<InvocationExample> getExamples() {
...
+            InvocationExample.of(baseEndpoint + "/groups/" + groupId + "/drives", "list_groups_drives.json"),
+            InvocationExample.of(baseEndpoint + "/groups/" + groupId + "/drives?$select=id,driveType,system&$skiptoken=abcXYZ123&$top=999&$orderby=id&$expand=root", "list_groups_drives.json"),
             // /v1.0/drives/{driveId}/root/delta - no query params, and with all allowed query params
             InvocationExample.of(baseEndpoint + "/drives/" + driveId + "/root/delta", "get_drive_delta.json"),
+            // real pagination continuation shape confirmed via live testing against the Graph API
+            // (a "?token=..." query param, matching MSFT's own docs example)
             InvocationExample.of(baseEndpoint + "/drives/" + driveId + "/root/delta?token=abcXYZ123&", "get_drive_delta.json"),
+            // the OTHER real continuation shape, from this fixture's own captured @odata.nextLink
+            // (an OData function-call path segment -- rules must accept both)
+            InvocationExample.of(baseEndpoint + "/drives/" + driveId + "/root/delta(token=1230919asd190410jlka)", "get_drive_delta.json"),
```
eschultink, line 59: mmmm ... OK; so I can see why this would need to be a regex; f'ing msft.

still possibly as a template the following would work?

pathTemplate: "/v1.0/drives/{driveId}/root/{deltaExpression}"

then could add paramSpec for `deltaExpression` that enforces pattern just against that portion - which may be simpler. something like `^delta\(token=\D+\)$`?

but nice-to-have imho.


might require a second endpoint spec for the other case:

pathTemplate: "/v1.0/drives/{driveId}/root/delta" as before

I think duplication preferable to single regex that is hard to read (and thus hard to be certain that it's obviously correct)

eschultink (approved):

aperez-worklytics (commented):
Split rules endpoint

java/core/src/test/java/co/worklytics/psoxy/rules/msft/OneDrive_NoAppIds_Tests.java
aperez-worklytics: The funny thing is that while doing onedrive I've never faced the delta in path, that's something spotted while doing tests on https://github.com/Worklytics/psoxy/pull/1401. And it is not clear why in the docs it says that delta is part of the path meanwhile it says that it could be a parameter https://learn.microsoft.com/en-us/graph/api/driveitem-delta?view=graph-rest-1.0&tabs=http#example-3-retrieving-the-current-deltalink, which it makes sense (besides that this what I see in the graph tests I have) because it is how it works ALL endpoints in graph.

jlorper (approved):

PR #1411, opened 2026-09-03
fix v0.7.0

eschultink:
### Misc
- Stop treating source token-exchange failures as a request parse error.

### Change implications

- breaking changes? **no**

### DIFF
--- a/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
+++ b/java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
@@ -423,16 +423,21 @@ public class ApiDataRequestHandler {
             if (isSocketTimeoutException(e)) {
                 return buildConnectTimeoutErrorResponse(builder, e);
             }
-            
-            builder.statusCode(HttpStatus.SC_INTERNAL_SERVER_ERROR);
-            builder.body("Failed to parse request; review logs");
+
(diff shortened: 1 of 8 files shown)


java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -1172,6 +1177,39 @@ private boolean isSocketTimeoutException(Throwable throwable) {
...
     }
 
+    /**
+     * True when outbound source-auth failed while obtaining an access token (e.g. Google service
+     * account JWT exchange 401), as opposed to a malformed inbound request.
+     */
+    @VisibleForTesting
+    static boolean isSourceAuthTokenFailure(IOException e) {
+        Throwable cause = e;
+        while (cause != null) {
```
jlorper: ok, but this core class shouldn't know about particular implementation auth errors.
We should have per-implementation knowledge (aws/gcp) to know when this happens.
Interface + choose implementation + check

jlorper (approved):

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
eschultink: yeah, agreed ... probably needs to be a method of `SourceAuthStrategy`

aperez-worklytics (approved):

PR #1423, opened 2026-09-11
MSFT Teams: update onlineMeetings setup documentation

aperez-worklytics:
> Adds the exact PowerShell commands needed to set up the Teams application access policy, fixes a field-name mismatch that was silently dropping these instructions from the generated setup doc, and extracts the msft-teams instructions into a template file for easier maintenance.

 - Field-name mismatch (`external_todo` vs `external_token_todo`) in `infra/modules/worklytics-connector-specs/msft-365.tf` that silently dropped the PowerShell setup instructions from the generated `TODO - setup msft-teams.md` file during `terraform apply`

 - Document the exact `New-CsApplicationAccessPolicy` / `Grant-CsApplicationAccessPolicy` PowerShell commands (not just a link to Microsoft's docs) in the `msft-teams` connector README, so customers can copy-paste them directly
 - Extract the msft-teams setup instructions out of an inline heredoc into `infra/modules/worklytics-connector-specs/docs/msft-teams/instructions.tftpl`, following the same pattern already used for `claude`/`cursor`/etc., for easier maintenance

 ### Logistics
 - [Asana task](https://app.asana.com/1/27145998307022/project/1218001620445764/task/1218371531744442)

### Change implications

 - dependencies added/changed? no
   - The generated `TODO - setup msft-teams.md` will now include the PowerShell section (previously silently dropped due to the field-name bug) - output content differs, but this is a bug fix, not a breaking change
   - not in a module/example marked `alpha`, but this is a bug fix restoring intended behavior, not a breaking change

### Test plan
- [x] `terraform validate` passes in `infra/modules/worklytics-connector-specs` and `infra/modules/worklytics-connectors-msft-365`
- [x] Verified via `terraform console` that `templatefile(...)` on the new `.tftpl` renders byte-identical content to the old heredoc (including the `%%entraid.client_id%%` marker, still substituted downstream in `worklytics-connectors-msft-365/main.tf`)
- [ ] `terraform apply` in a real msft-365 deployment to confirm the generated `TODO - setup msft-teams.md` now includes the PowerShell section with the real client ID substituted in

### DIFF
--- a/infra/modules/worklytics-connector-specs/docs/msft-teams/instructions.tftpl
+++ b/infra/modules/worklytics-connector-specs/docs/msft-teams/instructions.tftpl
@@ -1,0 +1,29 @@
+To enable the connector, you need to allow permissions on the application created for reading OnlineMeetings. You will need Powershell for this.
+
+Please follow the steps below:
+1. Ensure the user you are going to use for running the commands has the "Teams Administrator" role. You can add the role in the
+[Microsoft 365 Admin Center](https://learn.microsoft.com/en-us/microsoft-365/admin/add-users/assign-admin-roles?view=o365-worldwide#assign-a-user-to-an-admin-role-from-active-users)
+
+**NOTE**: About the role, can be assigned through Entra Id portal in Azure portal OR in Entra Admin center https://admin.microsoft.com/AdminPortal/Home. It is possible that even login with an admin account in Entra Admin Center the Teams role is not available to assign to any user; if so, please do it through Azure Portal (Entra Id -> Users -> Assign roles)
+
+2. Install [PowerShell Teams](https://learn.microsoft.com/en-us/microsoftteams/teams-powershell-install)  You can use `pwsh` in the terminal
+    enter to PowerShell.
+3. Then, run the following command. It will open a browser window for login to Microsoft Teams. After login, close the browser and return to the terminal.
+   Please choose the user who has the "Teams Administrator" role.
+```shell
+Connect-MicrosoftTeams
+```
+
+4. Follow steps on [Configure application access to online meetings or virtual events](https://learn.microsoft.com/en-us/graph/cloud-communication-online-meeting-application-access-policy):
+  - Add a policy for the application created for the connector, providing its `application id` (client ID)
+```shell
+New-CsApplicationAccessPolicy -Identity Teams-Policy-For-Worklytics -AppIds "%%entraid.client_id%%" -Description "Policy for MSFT Teams used for Worklytics Psoxy connector"
+```
+  - Grant the policy to the whole tenant (NOT to any specific application or user)
+```shell
+Grant-CsApplicationAccessPolicy -PolicyName Teams-Policy-For-Worklytics -Global
+```
+
+**Issues**:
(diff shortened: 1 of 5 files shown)


infra/modules/worklytics-connector-specs/msft-365.tf
```diff
@@ -28,7 +28,7 @@ locals {
       "MailboxSettings.Read"
     ]
     environment_variables : local.msft_365_environment_variables
-    external_todo : null
+    external_token_todo : null
```
aperez-worklytics, line 31: Fixing this, that was broken

infra/modules/worklytics-connector-specs/docs/msft-teams/instructions.tftpl
```diff
@@ -0,0 +1,29 @@
...
+3. Then, run the following command. It will open a browser window for login to Microsoft Teams. After login, close the browser and return to the terminal.
+   Please choose the user who has the "Teams Administrator" role.
+```shell
+Connect-MicrosoftTeams
+```
+
+4. Follow steps on [Configure application access to online meetings or virtual events](https://learn.microsoft.com/en-us/graph/cloud-communication-online-meeting-application-access-policy):
+  - Add a policy for the application created for the connector, providing its `application id` (client ID)
+```shell
+New-CsApplicationAccessPolicy -Identity Teams-Policy-For-Worklytics -AppIds "%%entraid.client_id%%" -Description "Policy for MSFT Teams used for Worklytics Psoxy connector"
```
aperez-worklytics, line 20: Adding here the commands needed to run.

infra/modules/worklytics-connector-specs/msft-365.tf
```diff
@@ -168,37 +168,7 @@ locals {
...
-  - Grant the policy to the whole tenant (NOT to any specific application or user)
-```shell
-Grant-CsApplicationAccessPolicy -PolicyName Teams-Policy-For-Worklytics -Global
-```
-
-**Issues**:
-- If you receive "access denied" is because no admin role for Teams has been detected. Please close and reopen the Powershell terminal after assigning the role.
-- Commands have been tested over a Powershell (7.4.0) terminal in Windows, installed from Microsoft Store and with Teams Module (5.8.0). It might not work on a different environment
-EOT
+      external_token_todo : templatefile("${path.module}/docs/msft-teams/instructions.tftpl", {})
```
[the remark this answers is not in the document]
aperez-worklytics: Good catch — fixed in 188390828: `local_file.todo-with-external-todo` now reads `filename`/`todo` from `module.msft_365_grant_to_shared[0]` instead of `module.msft_365_grants[each.key]` when `provision_entraid_apps` is false (shared-app path).

aperez-worklytics:
Good catch, not a false positive — fixed in 00a4eba6: factored the enriched per-connector todo text (base grant todo + external_token_todo with client-ID substitution) into a single `local.msft_365_todos` map, and now use it for both the `todos` output and the `local_file.todo-with-external-todo` content, so the `todos_as_outputs` flow (and the shared-app path) get the msft-teams PowerShell instructions too.

eschultink (approved):

PR #1431, opened 2026-09-15
gen metadata

eschultink:
### Features

- Add genMetadata augment for custom rules via Vertex and Bedrock.

### Change implications

- dependencies added/changed? **yes** — LangChain4j `1.16.0` (`langchain4j`, `langchain4j-bedrock`); Vertex uses `langchain4j-google-genai` `1.16.0-beta26`
- something important to note in future release notes? **yes** — BETA `!<genMetadata>` augment (custom rules only; cloud Bedrock/Vertex; `enable_gen_metadata`)

Internally deployed end-to-end test of bulk for both AWS + GCP:

https://github.com/Worklytics/proxy-dev/pull/9


docs/configuration/remote-resources.md
eschultink, line 1: restyle file; avoid hardwraps in md so renders well in gitbook/etc.

eschultink: drop rest of this section; too much detail; describes implementation, not stuff customer needs to know for configuration. More of a design doc.

eschultink: change this to "NLP Models for `sentenceMetadata`"

docs/configuration/remote-resources.md
```diff
@@ -135,9 +131,9 @@ aws s3 cp en-sent.bin s3://{REMOTE_RESOURCE_BUCKET}/{SHARED_RESOURCE_PATH}/openn
 gsutil cp en-sent.bin gs://{REMOTE_RESOURCE_BUCKET}/{SHARED_RESOURCE_PATH}/opennlp/en-sent.bin
 ```
 
-### LLM Weights (future)
-Smaller language models that fit in memory can be placed in the shared resource path for
-on-the-fly inference within the proxy.
+### genMetadata (BETA)
+
+The `!<genMetadata>` augment uses **Vertex AI** (GCP) or **Amazon Bedrock** (AWS). `enable_gen_metadata` does not set `REMOTE_RESOURCE_BUCKET` / `SHARED_RESOURCE_PATH` and does not load LLM weights from the artifacts bucket. If those env vars appear on Lambdas or Cloud Functions, they come from host `enable_remote_resources = true` (remote rules / OpenNLP). See [gen-metadata-augment.md](../development/alpha-features/gen-metadata-augment.md).
```
eschultink: too verbose explanation; just say `genMeta` augments - link alpha feature docs; explain implementation details there, if/as needed.

docs/development/alpha-features/gen-metadata-augment.md
```diff
@@ -0,0 +1,203 @@
...
+Wrong cloud on wrong platform → `augment-gen-unavailable`. Auth / quota failures omit the augment and add a warning; the rest of the response still succeeds.
+
+## Two inference modes (driven by `outputSchema`)
+
+Rules always declare `prompt` + `outputSchema`. The runtime **infers mode** from the schema (no separate `mode` field):
+
+| Mode | When | Model output | Proxy normalizes to |
+|------|------|--------------|---------------------|
+| **classify** | Schema is an object with a **single** required string property that has `enum`, or a root string `enum` | Vertex: constrained JSON. Bedrock: free-form JSON/label in text | Same Map / string after parse / label recovery |
+| **extract** | Any richer object / array schema | Constrained JSON matching the schema | Parsed object/array as-is (after schema gate) |
```
eschultink: let's have a better name for this; call it `compute` or something. It's not directly "extracting" any subset of the content.

docs/development/alpha-features/gen-metadata-augment.md
```diff
@@ -0,0 +1,203 @@
...
+## Infrastructure (Terraform)
+
+Set `enable_gen_metadata = true` on the API or bulk connector you want to try, and load **custom rules** that declare `!<genMetadata>` augments.
+
+Host defaults:
+
+- AWS: `gen_metadata_backend = "bedrock"` (validated)
+- GCP: `gen_metadata_backend = "vertex"` (validated)
+
+Optional per-connector override: `gen_metadata_backend` on the connector object.
```
eschultink: drop this section; there is not OTHER implentation; if anything, move this docs under gcp/aws specific documentation

docs/SUMMARY.md
```diff
@@ -56,6 +56,7 @@
   * [GCP External Application Load Balancer (ALB) + Cloud Armor](development/gcp-external-alb.md)
   * [Alpha Features](development/alpha-features/README.md)
     * [Webhook Collectors](development/alpha-features/webhook-collectors.md)
+    * [genMetadata Augment](development/alpha-features/gen-metadata-augment.md)
```
eschultink, line 58: move up - alphabetize menus like this.

infra/examples-dev/aws/variables.tf
```diff
@@ -525,4 +531,16 @@ variable "todos_as_local_files" {
...
+variable "enable_remote_resources" {
+  type        = bool
+  description = "**beta** Load rules / OpenNLP models from the artifacts S3 bucket at runtime. Default false; not required for genMetadata (Bedrock). Set true only if rules are too large for SSM / env, or you use sentenceMetadata."
+  default     = false
+}
+
+variable "gen_metadata_backend" {
+  type        = string
+  description = "Default genMetadata backend for enable_gen_metadata connectors: \"bedrock\" only."
+  default     = "bedrock"
```
eschultink: validation - do we have any implementation OTHER than bedrock atm? I guess could have local/embedded option in future.

infra/modules/aws-host/gen_metadata.tf
```diff
@@ -0,0 +1,33 @@
...
+    error_message = "aws-host genMetadata backend must be \"bedrock\" (Vertex is GCP-only)."
+  }
+}
+
+output "remote_resource_gen_metadata_todo" {
+  description = "TODO (markdown) for Bedrock genMetadata setup when any connector has enable_gen_metadata. Null when unused."
+  value = local.gen_metadata_uses_bedrock ? trimspace(<<-EOT
+	## Bedrock genMetadata setup
+
+	Connectors with `enable_gen_metadata` use Amazon Bedrock. Default model (Java): `us.amazon.nova-2-lite-v1:0` (US cross-region inference profile). Override with env `GEN_METADATA_MODEL` if needed (e.g. `eu.amazon.nova-2-lite-v1:0` or a Claude id).
```
eschultink: don't explain this in TODO.

infra/modules/aws-host/gen_metadata.tf
```diff
@@ -0,0 +1,33 @@
...
+
+	### Usually automatic (Terraform)
+
+	- Lambda/connector IAM already allows `bedrock:InvokeModel` / `bedrock:Converse` on foundation models and inference profiles.
+	- Prefer deploying the proxy in a **US** region when using the default `us.*` profile.
```
eschultink: this is unclear/confusing.

infra/modules/aws-host/gen_metadata.tf
```diff
@@ -0,0 +1,33 @@
...
+
+	### Manual / account steps (not fully automatable in Terraform)
```
eschultink: just call this the TODO steps. 

obv not automatable in terraform, or we would have done it.

infra/modules/aws-host/gen_metadata.tf
```diff
@@ -0,0 +1,33 @@
...
+
+	1. **Amazon Nova (default):** Use a CRIS inference profile id (`us.amazon.nova-2-lite-v1:0`), not the bare foundation-model id (`amazon.nova-2-lite-v1:0`) — on-demand invoke of the bare id returns 400. With Bedrock’s simplified model access, AWS first-party Nova is generally available without a console “enable model” step. If invokes fail with access errors, confirm the account/region is not blocked by an SCP and that Bedrock is usable in that region ([model access](https://docs.aws.amazon.com/bedrock/latest/userguide/model-access.html)).
+	2. **Anthropic Claude (only if you set `GEN_METADATA_MODEL` to a Claude id):** First-time Anthropic use in the account still requires a one-time use-case form (Bedrock console playground or `PutUseCaseForModelAccess`) — Terraform does not submit that form for you.
+
+	See [gen-metadata-augment.md](https://github.com/worklytics/psoxy/blob/main/docs/development/alpha-features/gen-metadata-augment.md).
```
eschultink: why?? also make links relative to actual gitbook hosted doc site - not github repo.

infra/modules/aws-host/gen_metadata.tf
```diff
@@ -0,0 +1,33 @@
...
+	Connectors with `enable_gen_metadata` use Amazon Bedrock. Default model (Java): `us.amazon.nova-2-lite-v1:0` (US cross-region inference profile). Override with env `GEN_METADATA_MODEL` if needed (e.g. `eu.amazon.nova-2-lite-v1:0` or a Claude id).
+
+	### Usually automatic (Terraform)
+
+	- Lambda/connector IAM already allows `bedrock:InvokeModel` / `bedrock:Converse` on foundation models and inference profiles.
+	- Prefer deploying the proxy in a **US** region when using the default `us.*` profile.
+
+	### Manual / account steps (not fully automatable in Terraform)
+
+	1. **Amazon Nova (default):** Use a CRIS inference profile id (`us.amazon.nova-2-lite-v1:0`), not the bare foundation-model id (`amazon.nova-2-lite-v1:0`) — on-demand invoke of the bare id returns 400. With Bedrock’s simplified model access, AWS first-party Nova is generally available without a console “enable model” step. If invokes fail with access errors, confirm the account/region is not blocked by an SCP and that Bedrock is usable in that region ([model access](https://docs.aws.amazon.com/bedrock/latest/userguide/model-access.html)).
```
eschultink: this is unclear what I need to do; KISS, just tell customer WHAT and HOW they need to activate/confgiure in their AWS account, step-by-step - in addition to what TF has done. Don't explain what TF has done for them already; don't explain why. Don't mix in troubleshooting into those steps

I'm OK leaving the SCP part in TODO. but separate note at the end.

infra/modules/aws-host/gen_metadata.tf
```diff
@@ -0,0 +1,33 @@
...
+	2. **Anthropic Claude (only if you set `GEN_METADATA_MODEL` to a Claude id):** First-time Anthropic use in the account still requires a one-time use-case form (Bedrock console playground or `PutUseCaseForModelAccess`) — Terraform does not submit that form for you.
```
eschultink: drop this; move to gen-metadata-augment.md docs or whatever the claude set-up stuff; using model sother than aws nova seems prohibitvely expensive so advanced/edge case.

infra/modules/aws-host/main.tf
```diff
@@ -86,6 +86,36 @@ locals {
 
   # proxy caller role requires direct lambda access if API Gateway v2 is not used and there are API connectors
   caller_requires_direct_lambda_access = !local.use_api_gateway_v2 && length(module.api_connector) > 0
+
+  # Effective genMetadata backend per connector (null if genMetadata disabled).
+  # AWS is cloud-only: bedrock.
```
eschultink: at least temp; mark as such.

infra/modules/aws-host/main.tf
```diff
@@ -250,6 +281,12 @@ module "api_connector" {
   timeout_seconds               = coalesce(try(each.value.timeout_seconds, null), 180)
   allowed_data_access_ip_blocks = var.allowed_data_access_ip_blocks
 
+  extra_lambda_role_iam_statements = (
```
eschultink: very generic var name.

also where is it used? 

is this intermediate local even needed?

I think conditional concat a list of all potential statements - conditionally putting empty or the actual list - then combine all. shouldn't go one-by-one.

infra/modules/aws-host/main.tf
```diff
@@ -339,6 +380,12 @@ module "bulk_connector" {
 
 
 
+  extra_lambda_role_iam_statements = (
+    local.connector_gen_metadata_backend[each.key] == "bedrock"
+    ? local.bedrock_invoke_iam_statements
+    : []
+  )
```
eschultink, line 373: very generic var name.

also where is it used? 

is this intermediate local even needed?

I think conditional concat a list of all potential statements - conditionally putting empty or the actual list - then combine all. shouldn't go one-by-one.

infra/modules/gcp-host/gen_metadata.tf
```diff
@@ -0,0 +1,43 @@
...
+      if backend == "vertex" && contains(keys(var.api_connectors), k)
+    },
+    {
+      for k, backend in local.connector_gen_metadata_backend : k => module.bulk_connector[k].instance_sa_email
+      if backend == "vertex" && contains(keys(var.bulk_connectors), k)
+    },
+  )
+}
+
+resource "google_project_service" "aiplatform" {
```
eschultink: why nnot just list of services to enable, iterate over that - rather than conditional for each service?

infra/modules/gcp-host/gen_metadata.tf
eschultink, line 1: gen_metadata.tf - rename to `gen-metadata.tf` for style consistency - IF not merge this stuff into `main.tf`. Splitting .tf files by feature/functionality seems OK to me, but i believe it's a departure from canonical terraform style.

infra/modules/gcp-host/gen_metadata.tf
```diff
@@ -0,0 +1,43 @@
...
+
+check "gen_metadata_backend_gcp_only" {
+  assert {
+    condition = alltrue([
+      for k, backend in local.connector_gen_metadata_backend :
+      backend == null || backend == "vertex"
+    ])
+    error_message = "gcp-host genMetadata backend must be \"vertex\" (Bedrock is AWS-only)."
+  }
+}
```
eschultink: drop this; I don't like such validations in modules;  makes them unnecessarily brittle; at most, do a test in CI

infra/modules/gcp-host/main.tf
```diff
@@ -87,6 +89,21 @@ locals {
...
+      try(v.enable_gen_metadata, false)
+      ? lower(coalesce(try(v.gen_metadata_backend, null), var.gen_metadata_backend, "vertex"))
+      : null
+    )
+  }
+
+  gen_metadata_uses_vertex = length([
+    for k, backend in local.connector_gen_metadata_backend : k
+    if backend == "vertex"
+  ]) > 0
```
eschultink, line 97: so this is problematic in combination with the separate `gen-metadata.tf` file; that file doesn't standalone, which undermines merit in departing from canonical tf style by splitting that out.  

Also comment on why this is necessary; it seems like it's intended to provide extensibility for local/other remote backends in the future other than vertex; I'm inclined to say YAGNI and drop this functionality for now.

infra/modules/gcp-host/variables.tf
```diff
@@ -193,6 +193,8 @@ variable "api_connectors" {
     oauth_scopes_needed     = optional(list(string), [])
     environment_variables   = optional(map(string), {})
     enable_async_processing = optional(bool, false)
+    enable_gen_metadata     = optional(bool, false)
+    gen_metadata_backend    = optional(string) # "vertex" (GCP); host default applies when null
```
eschultink: so just make it optional(string, "vertex") for clarity; that's the effective default.

infra/modules/gcp-host/variables.tf
```diff
@@ -499,3 +503,14 @@ variable "api_connector_external_lb_host" {
     error_message = "api_connector_external_lb_host must be null or a non-empty hostname/IP."
   }
 }
+
+variable "gen_metadata_backend" {
```
eschultink: this is confusing imho to have defined in TWO places - both on `gcp-host` module and as an option for each of the specific connector instances. I would say DROP this (and any equiv in aws-host) for now.  If we need this functionality in the future as a "default" host for any default when true, just call it `default_gen_metadata_backend` so clear which takes precedence.

infra/modules/gcp-host/variables.tf
```diff
@@ -499,3 +503,14 @@ variable "api_connector_external_lb_host" {
...
+  type        = string
+  description = "Default genMetadata backend for connectors with enable_gen_metadata when not set per connector. On GCP: \"vertex\" only."
+  default     = "vertex"
+
+  validation {
+    condition     = var.gen_metadata_backend == "vertex"
+    error_message = "gen_metadata_backend must be \"vertex\" on gcp-host (Bedrock is AWS-only)."
+  }
```
eschultink: while reasonable, validation in modules makes the tf brittle

java/core/src/main/java/co/worklytics/psoxy/gateway/impl/ApiDataRequestHandler.java
```diff
@@ -687,11 +691,11 @@ ProcessedContent sanitize(HttpEventRequest requestToProxy, RequestUrls requestUr
         metadata.put(ProcessedDataMetadataFields.PII_SALT_SHA256.getMetadataKey(),
                 healthCheckRequestHandler.piiSaltHash());
 
-        // q: add instance id to the metadata??
         return ProcessedContent.builder()
                 .contentType(originalContent.getContentType())
                 .contentCharset(originalContent.getContentCharset())
                 .metadata(metadata)
+                .sanitizationWarnings(sanitizerForRequest.getLastSanitizationWarnings())
```
eschultink: seems too bespoked to our use-case; add as `ProcessedDataMetadataFields` instead.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/BedrockGenMetadataConfig.java
eschultink, line 1: I think move this and other bedrock-specific files down into impl.aws package; and similarly vertex-specific files down into impl.gcp

There's just NO immediate use-case for cross-platform calls like this - versus merit in keeping each deployed JAR as small as possible.  If this presents a challenge for cmd-line mode or whatever - so be it;  have that throw NotSupportException for now (ideally choosing standard java/apache equiv) - unless solution is to have cmd-line module depend on BOTH gcp/aws packages itself, which actually doesn't seem unreasably for me (cmd-line JAR would be huge, but that's OK).

java/core/src/main/java/co/worklytics/psoxy/impl/gen/BedrockGenMetadataConfig.java
```diff
@@ -0,0 +1,51 @@
...
+     * a cross-region inference profile (e.g. {@code us.amazon.nova-2-lite-v1:0}). Bare Nova ids are
+     * rewritten to the US profile. Ids that already have a geo/global prefix are left unchanged.
+     */
+    static String normalizeModelId(String modelId) {
+        if (StringUtils.isBlank(modelId)) {
+            return modelId;
+        }
+        String id = modelId.trim();
+        if (id.startsWith("amazon.nova")) {
+            return "us." + id;
```
eschultink: avoid magic strings; make constants for these, at min at top for clarityf.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/BedrockGenMetadataConfig.java
```diff
@@ -0,0 +1,51 @@
...
+        return new BedrockGenMetadataConfig(
+            normalizeModelId(modelId), timeoutSeconds, DEFAULT_MAX_ATTEMPTS);
+    }
+
+    /**
+     * Nova foundation-model ids ({@code amazon.nova-…}) cannot be invoked on-demand; Bedrock requires
+     * a cross-region inference profile (e.g. {@code us.amazon.nova-2-lite-v1:0}). Bare Nova ids are
+     * rewritten to the US profile. Ids that already have a geo/global prefix are left unchanged.
+     */
+    static String normalizeModelId(String modelId) {
```
eschultink, line 44: why should this be static; we should avoid statics.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataConfig.java
```diff
@@ -0,0 +1,131 @@
...
+            .orElse(null);
+        Optional<Backend> parsed = Backend.fromConfigValue(rawBackend);
+        if (rawBackend != null && parsed.isEmpty()) {
+            return new Unsupported(rawBackend);
+        }
+        Backend backend = parsed.orElse(Backend.BEDROCK);
+        return switch (backend) {
+            case BEDROCK -> BedrockGenMetadataConfig.from(configService);
+            case VERTEX -> VertexGenMetadataConfig.from(configService);
+        };
```
eschultink: this looks like something that Dagger should be able to do for us; gcp/aws modules would respectively provide GenMetadataConfig, with the platform-specific implementations; you could *check* this or do similar switch there, if as needed (if those later support local backends or something.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataConfig.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public boolean isSupportedCloudBackend() {
```
eschultink, line 70: out of the scope of this; let Dagger handle this stuff in more pluggable/extensible way.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataPromptBudget.java
```diff
@@ -0,0 +1,35 @@
...
+ * Caps the <em>dynamic</em> genMetadata corpus (serialized source match) in estimated tokens.
+ *
+ * <p>The static prefix — system text, task {@code prompt}, and schema/labels — is not counted.
+ * That prefix is stable per augment rule and is the candidate for future provider prompt caching.
+ *
+ * <p>Estimate is {@value #CHARS_PER_TOKEN_ESTIMATE} chars/token, not a model tokenizer.
+ */
+@Singleton
+@NoArgsConstructor(onConstructor_ = @Inject)
+public class GenMetadataPromptBudget {
```
eschultink, line 20: 1) why do we need to "estimate" tokens? let's not do that - just tokenize.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataPromptBudget.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /** Rough chars-per-token for English / JSON-escaped input. */
+    static final int CHARS_PER_TOKEN_ESTIMATE = 4;
+
+    /**
+     * Truncate serialized source so it fits {@code maxInputTokens} estimated tokens.
+     */
+    public String fitDynamicInput(String inputData, int maxInputTokens) {
```
eschultink, line 28: this is still; just have a TextTokensUtils or something class, with a `truncateToMaxTokens(String input, int maxTokens)`

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataPromptBuilder.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+@UtilityClass
+class GenMetadataPromptBuilder {
+
+    static final String SYSTEM_CLASSIFY =
+        "You are a data-processing component in a privacy proxy. "
+            + "Respond with exactly one JSON object whose single property is the classification label. "
+            + "Example shape: {\"category\":\"<label>\"}. "
+            + "Use only an allowed label as the property value. "
+            + "No markdown fences, no prose before or after the JSON.";
```
eschultink, line 30: remove this case for now.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataPromptBuilder.java
```diff
@@ -0,0 +1,107 @@
...
+
+    static final String SYSTEM_CLASSIFY_STRING =
+        "You are a data-processing component in a privacy proxy. "
+            + "Respond with exactly one allowed classification label as a JSON string. "
```
eschultink, line 34: just 'string', not JSON string.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataPromptBuilder.java
```diff
@@ -0,0 +1,107 @@
...
+            + "Example: \"Research and Ideation\". "
+            + "Do not wrap the label in an object or add any other keys. "
```
eschultink, line 35: i think we can drop this line?

do we need the example?

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataPromptBuilder.java
```diff
@@ -0,0 +1,107 @@
...
+import dev.langchain4j.data.message.UserMessage;
+import lombok.experimental.UtilityClass;
+
+import java.util.List;
+
+/**
+ * Builds chat messages for genMetadata cloud inference (Bedrock / Vertex).
+ */
+@UtilityClass
+class GenMetadataPromptBuilder {
```
eschultink, line 21: too many static methods here; we use dagger and should follow canonical ui.

don't use Dagger UTilityCalss - experimental, and I'm not sure it has a role in oure codebase yet.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataPromptBuilder.java
```diff
@@ -0,0 +1,107 @@
...
+            + "not a JSON Schema definition. "
+            + "Never include schema keywords such as type, properties, required, or enum. "
+            + "No markdown fences, no prose before or after the JSON.";
+
+    static List<ChatMessage> toMessages(String taskPrompt, JsonSchemaFilter outputSchema,
+                                        String inputData, ObjectMapper objectMapper) {
+        if (GenMetadataSchemaSupport.mode(outputSchema) == GenMetadataSchemaSupport.Mode.CLASSIFY) {
+            GenMetadataSchemaSupport.ClassifyShape shape =
+                GenMetadataSchemaSupport.classifyShape(outputSchema).orElseThrow();
+            String system = shape.isRootString() ? SYSTEM_CLASSIFY_STRING : SYSTEM_CLASSIFY;
```
eschultink, line 50: this kinda discovery seems a little weird.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataResponseFormats.java
```diff
@@ -0,0 +1,116 @@
...
+                ? JsonEnumSchema.builder()
+                    .enumValues(shape.getEnumValues())
+                    .build()
+                : JsonObjectSchema.builder()
+                    .addProperty(shape.getPropertyName(), JsonEnumSchema.builder()
+                        .enumValues(shape.getEnumValues())
+                        .build())
+                    .required(shape.getPropertyName())
+                    .additionalProperties(false)
+                    .build();
```
eschultink, line 51: this is an unreadable abuse of ternary .. refactor.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataResponseFormats.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Builds LangChain4j {@link ResponseFormat} constraints from genMetadata {@link JsonSchemaFilter}.
+ */
+@UtilityClass
+public class GenMetadataResponseFormats {
```
eschultink, line 29: no lombok Utility class (add that to our AGENTS.md / style guide if needed).

make this a proper class and inject where needed; avoid so many `statics`; dagger-implemented DI is our pattern!

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataTokenUsageAccumulator.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+@Singleton
+public class GenMetadataTokenUsageAccumulator {
+
+    private final AtomicLong inputTokens = new AtomicLong();
+    private final AtomicLong outputTokens = new AtomicLong();
+    private final AtomicLong calls = new AtomicLong();
+
+    @Inject
+    public GenMetadataTokenUsageAccumulator() {
```
eschultink: lombok `@NoArgsConstructor`? more clear.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataTokenUsageAccumulator.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Record one completed LLM call. Null counts are treated as zero (provider omitted usage).
+     */
+    public void record(Integer inputTokenCount, Integer outputTokenCount) {
+        inputTokens.addAndGet(inputTokenCount != null ? inputTokenCount : 0);
```
eschultink: not a very clear why to this; may `Optional.ofNullable(inputTokenCount).ifPresent(inputTokens::add)`?

certainly `addAndGet` is weird; you don't actually use the value.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/LangChain4jGenMetadataBackend.java
```diff
@@ -0,0 +1,317 @@
...
+@Singleton
+public class LangChain4jGenMetadataBackend implements GenMetadataBackend {
+
+    static final int CLOUD_MAX_CONCURRENT = 4;
+
+    static final class ModelHandle {
+        final ChatModel chatModel;
+        final Exception failure;
+
+        private ModelHandle(ChatModel chatModel, Exception failure) {
```
eschultink, line 49: lombok `AllArgsConstructor`

ensure our AGENTS.md strongly encourages lombok/Dagger use as standard approach.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/LangChain4jGenMetadataBackend.java
```diff
@@ -0,0 +1,317 @@
...
+                                         GenMetadataPromptBudget promptBudget,
+                                         GenMetadataChatModelFactory chatModelFactory,
+                                         GenMetadataTokenUsageAccumulator tokenUsageAccumulator) {
+        this.config = config;
+        this.objectMapper = objectMapper;
+        this.promptBudget = promptBudget;
+        this.chatModelFactory = chatModelFactory;
+        this.tokenUsageAccumulator = tokenUsageAccumulator != null
+            ? tokenUsageAccumulator
+            : new GenMetadataTokenUsageAccumulator();
```
eschultink: abuse of ternary; at min, invert; if not convert to if-else.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/LangChain4jGenMetadataBackend.java
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    public LangChain4jGenMetadataBackend(GenMetadataConfig config, ObjectMapper objectMapper,
```
eschultink: very odd that this is public

java/core/src/main/java/co/worklytics/psoxy/impl/gen/LangChain4jGenMetadataBackend.java
```diff
@@ -0,0 +1,317 @@
...
+        }
+        return config.getBackend() != null ? config.getBackend().configValue() : "unknown";
+    }
+
+    private String thinkingLogSuffix() {
+        if (config instanceof VertexGenMetadataConfig vertex) {
+            return " thinkingLevel=" + vertex.getThinkingLevel();
+        }
+        return "";
+    }
```
eschultink, line 220: avoid coupling to Vertex pkg.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/LangChain4jGenMetadataBackend.java
```diff
@@ -0,0 +1,317 @@
...
+
+    static boolean isAuthOrQuotaFailure(Throwable t) {
+        for (Throwable c = t; c != null; c = c.getCause()) {
+            String name = c.getClass().getName();
+            String msg = c.getMessage() != null ? c.getMessage().toLowerCase() : "";
+            if (name.contains("AccessDenied")
+                || name.contains("Authorization")
+                || name.contains("PermissionDenied")
+                || name.contains("ResourceExhausted")
+                || name.contains("Throttling")
```
eschultink: pointless given below.

all checks shoudl prob be case-insitive.

cleanest to split to isAuthFailure, isQuotaFailure - and explicitly all OR accross those wehre needed.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/LangChain4jGenMetadataBackend.java
```diff
@@ -0,0 +1,317 @@
...
+        if (usage == null && response.metadata() != null) {
+            usage = response.metadata().tokenUsage();
+        }
+        Integer input = usage != null ? usage.inputTokenCount() : null;
+        Integer output = usage != null ? usage.outputTokenCount() : null;
+        tokenUsageAccumulator.record(input, output);
+    }
+
+    static boolean isAuthOrQuotaFailure(Throwable t) {
+        for (Throwable c = t; c != null; c = c.getCause()) {
```
eschultink, line 280: hacky way to recurse, looks like iteration, but isn't .... needs a comment at min

java/core/src/main/java/co/worklytics/psoxy/impl/gen/LangChain4jGenMetadataBackend.java
```diff
@@ -0,0 +1,317 @@
...
+        var usage = response.tokenUsage();
+        if (usage == null && response.metadata() != null) {
+            usage = response.metadata().tokenUsage();
+        }
+        Integer input = usage != null ? usage.inputTokenCount() : null;
+        Integer output = usage != null ? usage.outputTokenCount() : null;
+        tokenUsageAccumulator.record(input, output);
+    }
+
+    static boolean isAuthOrQuotaFailure(Throwable t) {
```
eschultink: shouldn't be static; document what method does.

java/core/src/main/java/co/worklytics/psoxy/impl/AugmentProcessor.java
```diff
@@ -186,34 +225,85 @@ private void applyAugmentAtConcretePath(Augment augment, Object document, String
...
-            // For PoC, outputSchema validation is deferred
+    private static AugmentProcessingException toAugmentProcessingException(
+            GenMetadataAugmentException e) {
+        Warning warning = switch (e.getCode()) {
+            case UNAVAILABLE -> Warning.AUGMENT_GEN_UNAVAILABLE;
+            case INFERENCE_FAILED -> Warning.AUGMENT_GEN_INFERENCE_FAILED;
+        };
+        return e.getCause() != null
+            ? new AugmentProcessingException(warning, e.getMessage(), e.getCause())
+            : new AugmentProcessingException(warning, e.getMessage());
```
eschultink: feels like AugmentProcessingException could do this internally - and just take `Throwable e` as th argument - map message/cause internally.

java/core/src/main/java/co/worklytics/psoxy/impl/RESTApiSanitizerImpl.java
```diff
@@ -142,6 +143,14 @@ public RESTApiSanitizerImpl(@Assisted RESTRules rules,
     @Inject
     AugmentProcessor augmentProcessor;
 
+    private final ThreadLocal<List<String>> lastAugmentWarnings = new ThreadLocal<>();
+
+    @Override
+    public List<String> getLastSanitizationWarnings() {
+        List<String> warnings = lastAugmentWarnings.get();
+        return warnings == null ? List.of() : warnings;
```
eschultink: wtf; this is just return `Optional.ofNullable(lastAugmentWarnings.get()).orElse(List.of())`

java/core/src/main/java/co/worklytics/psoxy/PsoxyModule.java
```diff
@@ -38,22 +38,44 @@
...
 import dagger.Provides;
+import dagger.multibindings.Multibinds;
 import lombok.extern.java.Log;
 
 /**
  * provides implementations for platform-independent dependencies of 'core' module
  */
 @Log
-@Module
+@Module(includes = PsoxyModule.Bindings.class)
```
eschultink, line 60: not consistent with how we've done similar DI; take this approach here ONLY if it is truly the best approach (eg, this is something new).

java/core/src/main/java/co/worklytics/psoxy/PsoxyModule.java
```diff
@@ -382,4 +404,24 @@ Base64KeyClient base64KeyClient() {
...
+    }
+
+    @Provides
+    @Singleton
+    static GenMetadataProcessor genMetadataProcessor(GenMetadataBackend genMetadataBackend,
+                                                     ObjectMapper objectMapper,
+                                                     GenMetadataConfig genMetadataConfig,
+                                                     JsonSchemaValidationUtils jsonSchemaValidationUtils) {
+        return new GenMetadataProcessor(genMetadataBackend, objectMapper,
+            genMetadataConfig.getMaxAttempts(), jsonSchemaValidationUtils);
```
eschultink, line 415: I prefer to directly set `GenMetadataConfig` on GenMetadataProcessor; eg injectable constructor with `GenMetadataConfig` there, let GenMetadataProcessor worry about mapping what it cares about in `GenMetadataConfig`

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/Augment.java
```diff
@@ -236,4 +239,76 @@ static Set<String> signalWords(List<String> configured, List<String> defaults) {
...
+     * Requires {@link #outputSchema} and {@link #prompt}; model/backend selection is deployment config.
+     */
+    @SuperBuilder(toBuilder = true)
+    @NoArgsConstructor
+    @AllArgsConstructor
+    @Getter
+    @EqualsAndHashCode(callSuper = true)
+    public static class GenMetadata extends Augment {
+
+        public static final int DEFAULT_MAX_INPUT_TOKENS = 256;
```
eschultink, line 256: add comment describing, consist with others.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/Augment.java
```diff
@@ -236,4 +239,76 @@ static Set<String> signalWords(List<String> configured, List<String> defaults) {
...
+
+        /**
+         * Cap on the <em>dynamic</em> source corpus (serialized jsonPath match), in estimated
+         * tokens. The static task {@link #prompt} and schema/labels are not counted.
+         * Default {@value #DEFAULT_MAX_INPUT_TOKENS}. Typical classify: 100; longer extract: 500+.
+         */
+        @JsonInclude(JsonInclude.Include.NON_NULL)
+        Integer maxInputTokens;
+
+        public int effectiveMaxTokens() {
```
eschultink: i don't think this method is good approach; still let's people access getMaxTokens and get the wrong idea. And naming inconsistent; really a getter, just not such named.

custom get/set for maxTokens that inits with default for null; throws IllegalArgumentException if < 1, etc.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/Augment.java
```diff
@@ -236,4 +239,76 @@ static Set<String> signalWords(List<String> configured, List<String> defaults) {
...
+            return maxTokens != null && maxTokens > 0
```
eschultink: throw Error or IllegalArgument exception on maxTokens < 1.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/Augment.java
```diff
@@ -236,4 +239,76 @@ static Set<String> signalWords(List<String> configured, List<String> defaults) {
...
+                ? maxTokens
+                : DEFAULT_MAX_TOKENS;
+        }
+
+        public int effectiveMaxInputTokens() {
```
eschultink: similar to maxTokens cast.

for clarity, we should probably `maxOutputTokens`, `maxInputTokens` - more readiable. avoids ambiguity.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataAugmentException.java
```diff
@@ -0,0 +1,29 @@
...
+        super(message);
+        this.code = code;
+    }
+
+    public GenMetadataAugmentException(Code code, String message, Throwable cause) {
+        super(message, cause);
+        this.code = code;
+    }
+
+    public Code getCode() {
```
eschultink: lombok Getter

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataProcessor.java
```diff
@@ -0,0 +1,358 @@
...
+
+    public GenMetadataProcessor(GenMetadataBackend backend, ObjectMapper objectMapper,
+                                int maxAttempts, JsonSchemaValidationUtils jsonSchemaValidationUtils) {
+        this.backend = backend != null ? backend : new UnavailableGenMetadataBackend();
+        this.objectMapper = objectMapper != null ? objectMapper : new ObjectMapper();
+        this.jsonSchemaValidationUtils = jsonSchemaValidationUtils != null
+            ? jsonSchemaValidationUtils
+            : new JsonSchemaValidationUtils();
+        this.maxAttempts = maxAttempts > 0 ? maxAttempts : DEFAULT_MAX_ATTEMPTS;
+    }
```
eschultink, line 38: ijectable; let injection provider worry about the `new` here

properly architected DI code should have no `new` outside provider methods in modules, right?

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataProcessor.java
```diff
@@ -0,0 +1,358 @@
...
+        if (parsed instanceof Map<?, ?> map) {
+            log.info("genMetadata parsed output keys: " + map.keySet()
+                + "; value=" + truncateForLog(serializeForLog(parsed)));
+        } else {
+            log.info("genMetadata parsed output: " + truncateForLog(serializeForLog(parsed)));
+        }
+        return parsed;
+    }
+
+    private boolean validatesOutputSchema(Object parsed, JsonSchemaFilter outputSchema) {
```
eschultink: name of this method is unclear. it doesn't "validate" outputSchema, I believe this method is `isValuePerOutputSchema(outputSchema, parsed)`; this shouldn't directly relate to `GenMetadataProcessor` at all - so if needed this method should be in another lib/utils class somewhere.

But very similar stuff is used in API data sanitization - so I'm suprprised nohting similar exists.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataProcessor.java
```diff
@@ -0,0 +1,358 @@
...
+        try {
+            String json = objectMapper.writeValueAsString(parsed);
+            return jsonSchemaValidationUtils.validateJsonBySchema(json, outputSchema);
+        } catch (Exception e) {
```
eschultink: cathcing generic Exception is code smell; catch the validation failure explicity that we expect to happen in the invalide case, if we do; otherwise it's probably and error and SHOULD blow things up - right?  quiet warning + false is prob not what we want.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataProcessor.java
```diff
@@ -0,0 +1,358 @@
...
+                }
+                return objectMapper.writeValueAsString(text);
+            }
+            Object prepared = input instanceof Map<?, ?> map
+                ? prepareMapForLlm(map)
+                : stripAugmentKeysDeep(input);
+            return objectMapper.writeValueAsString(prepared);
+        } catch (Exception e) {
+            log.log(Level.WARNING, "Failed to serialize genMetadata input", e);
+            return null;
```
eschultink, line 133: see general concernes about catching generic Exception; avoid.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataProcessor.java
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+
+    /**
+     * Build LLM corpus from a matched JSON object: strip {@code +}-prefixed augment keys and put
+     * {@code title}/{@code body} first so {@code maxInputTokens} truncation of the dynamic corpus
```
eschultink: no no no 🤯 

this is hardcoding a SPECIFIC example in this somehow.  do not do that.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataProcessor.java
```diff
@@ -0,0 +1,358 @@
...
+     */
+    private LinkedHashMap<String, Object> prepareMapForLlm(Map<?, ?> map) {
+        LinkedHashMap<String, Object> ordered = new LinkedHashMap<>();
+        putPreferredTextField(ordered, map, "title");
+        putPreferredTextField(ordered, map, "body");
+        for (Map.Entry<?, ?> entry : map.entrySet()) {
+            if (!(entry.getKey() instanceof String key)) {
+                continue;
+            }
+            if (key.startsWith("+") || "title".equals(key) || "body".equals(key)) {
```
eschultink: `+` OK I think you're trying to avoid including fields that were prev introduced by our augments? don't love it, but if that's what's inteneded, refer to augment field prefix but its constant.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataProcessor.java
```diff
@@ -0,0 +1,358 @@
...
+    private void putPreferredTextField(LinkedHashMap<String, Object> ordered, Map<?, ?> map,
+                                       String key) {
+        if (!map.containsKey(key)) {
+            return;
+        }
+        ordered.put(key, stripAugmentKeysDeep(map.get(key)));
+    }
+
+    /**
+     * Recursively drop {@code +}-prefixed keys so prior augment output is not sent to the model.
```
eschultink: yeah, ok; 1) use a constant for this, 2) don't repeat same literal everywhere.

eschultink: separately - why do we care/need to drop it? 1) YAGNI - don't anticipate a case where this is needed; multiple LLM passes on same thing will be tooexpesnvie anyway; 2) for prev non-LLM augment, why enforce that can't be passed to LLM augment calls later?

IU'm inclining to just include augmented stuff, if present. We also have to configure that it's working on cotnent that has prev undergone proxy; no really do unexpected drops.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataProcessor.java
```diff
@@ -0,0 +1,358 @@
...
+            List<Object> cleaned = new ArrayList<>(list.size());
+            for (Object item : list) {
+                cleaned.add(stripAugmentKeysDeep(item));
+            }
+            return cleaned;
+        }
+        return value;
+    }
+    Object parseModelJson(Object raw) {
+        return parseModelJson(raw, null);
```
eschultink: doen't seem helpful

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataSchemaSupport.java
```diff
@@ -0,0 +1,188 @@
...
+        /** Property name for object classify; {@code null} when schema is a root string enum. */
+        String propertyName;
+        List<String> enumValues;
+
+        public boolean isRootString() {
+            return StringUtils.isBlank(propertyName);
+        }
+    }
+
+    private GenMetadataSchemaSupport() {
```
eschultink: lombok - NoArgsConstructor

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataSchemaSupport.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    public static Optional<String> findEnumLabelInText(String text, ClassifyShape shape) {
+        if (StringUtils.isBlank(text) || shape == null) {
+            return Optional.empty();
+        }
+        String haystack = text.trim();
+        String best = null;
+        for (String allowed : shape.getEnumValues()) {
+            if (StringUtils.isBlank(allowed)) {
```
eschultink, line 144: why would this happen ...

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataSchemaSupport.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    static String unwrapQuoted(String trimmed) {
+        if (trimmed.startsWith("\"") && trimmed.endsWith("\"") && trimmed.length() >= 2) {
+            return trimmed.substring(1, trimmed.length() - 1).trim();
+        }
+        return trimmed;
+    }
+
+    private static boolean containsIgnoreCase(String haystack, String needle) {
```
eschultink: so wtf - just call StringUtils.containsIgnoreCase directly - don't add our own wrapping.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/augments/GenMetadataThinkingLevels.java
```diff
@@ -0,0 +1,48 @@
...
+    public static final String LOW = "LOW";
+    public static final String MEDIUM = "MEDIUM";
+    public static final String HIGH = "HIGH";
+
+    /** Default when omitted from rules / null / blank. */
+    public static final String DEFAULT = MINIMAL;
+
+    private static final Set<String> ALLOWED = Set.of(MINIMAL, LOW, MEDIUM, HIGH);
+
+    private GenMetadataThinkingLevels() {
```
eschultink: lombok NoArgsConstructor

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilter.java
```diff
@@ -25,8 +25,6 @@
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties({
     "title",
-    "required", // not relevant to 'filter' use case
-    "additionalProperties", // not currently supported
```
eschultink, line 29: wtf is this inscope???

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaFilter.java
```diff
@@ -59,6 +57,23 @@ public Map<String, JsonSchemaFilter> getProperties() {
...
+    /**
+     * Required property names when validating augment output (predicate use case).
+     */
+    List<String> required;
+
+    /**
+     * Whether additional properties are allowed; used when validating augment output.
+     * Response-schema filtering still ignores this (always strips undeclared properties).
+     */
+    Boolean additionalProperties;
```
eschultink: so it seems like you're done stuff to JsonSchemaFilter, to re-use it for output validation on augments. this is not a good idea.  

Use JsonSchemaFilter for filtering.

Use canonical JsonSchema (from actually 3rd party lib) for validation. If we need our own `JsonSchema` rather than that - OK; create it. Don't make JsonSchemaFilter polymorphic magic that can work either way.  Otherwise stuff like required creates nonsense/ambiguity in filter context.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaValidationUtils.java
```diff
@@ -40,6 +41,28 @@ public class JsonSchemaValidationUtils {
...
+    private final LoadingCache<JsonSchemaFilter, Schema> jsonSchemaFilterCache =
+            CacheBuilder.newBuilder()
+                    .maximumSize(100)
+                    .build(CacheLoader.from(this::getJsonSchemaFromFilter));
+
+    /**
+     * Validates JSON against an augment {@link JsonSchemaFilter} (genMetadata output gate, etc.).
+     */
+    @SneakyThrows
+    public boolean validateJsonBySchema(String jsonString, JsonSchemaFilter schema) {
```
eschultink: no

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaValidationUtils.java
```diff
@@ -40,6 +41,28 @@ public class JsonSchemaValidationUtils {
...
+        List<Error> validationMessages = jsonSchema.validate(jsonString, InputFormat.JSON);
+        if (!validationMessages.isEmpty()) {
+            log.warning("Validation failed for augment output: " + validationMessages
+                + "; output=" + truncateForLog(jsonString));
+        }
+        return validationMessages.isEmpty();
+    }
+
     @SneakyThrows
     public boolean validateJsonBySchema(String jsonString,
```
eschultink, line 45: this is the validation method you should use, if need.

java/gateway-core/src/main/java/com/avaulta/gateway/rules/JsonSchemaValidationUtils.java
```diff
@@ -143,6 +166,20 @@ private Schema getJsonSchema(com.avaulta.gateway.rules.JsonSchema schema) {
                 JACKSON3.valueToTree(rewritePseudonymToPattern(schema)));
     }
 
+    private Schema getJsonSchemaFromFilter(JsonSchemaFilter schema) {
```
eschultink: maybe, but seems hacky; create json-serializable structure of JsonSchema if needed, use that in augment

java/gateway-core/src/main/java/com/avaulta/gateway/rules/RecordRules.java
```diff
@@ -82,6 +94,7 @@ public Optional<JsonSchemaFilter> getOutputSchemaFilterOptional() {
      */
     public RecordRules() {
         this.transforms = Collections.emptyList();
+        this.augments = new ArrayList<>();
```
eschultink, line 98: wtf would this be different approach than one line above?

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/BedrockChatModelProvider.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * <p>Uses the Lambda execution role / default AWS credentials chain; region from the default
+ * AWS region provider (Lambda {@code AWS_REGION}).
+ */
+@Singleton
+public class BedrockChatModelProvider implements GenMetadataChatModelProvider {
+
+    @Inject
+    public BedrockChatModelProvider() {
+    }
```
eschultink: lombok noArgsConstructor.

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/BedrockChatModelProvider.java
```diff
@@ -0,0 +1,44 @@
...
+    public boolean supports(GenMetadataConfig config) {
+        return config instanceof BedrockGenMetadataConfig;
+    }
+
+    @Override
+    public ChatModel create(GenMetadataConfig config, Path modelCacheDir) {
+        return BedrockChatModel.builder()
+            .modelId(config.getModelId())
+            .timeout(Duration.ofSeconds(config.getTimeoutSeconds()))
+            .maxRetries(1)
```
eschultink: should this not come from config??

java/impl/gcp/src/main/java/co/worklytics/psoxy/VertexGeminiChatModelProvider.java
```diff
@@ -0,0 +1,89 @@
...
+ * <p>For {@code location=global}, google-genai uses {@code https://aiplatform.googleapis.com}
+ * (not {@code global-aiplatform.googleapis.com}).
+ */
+@Log
+@Singleton
+public class VertexGeminiChatModelProvider implements GenMetadataChatModelProvider {
+
+    @Inject
+    public VertexGeminiChatModelProvider() {
+    }
```
eschultink: lombok NoargsConstructor

java/impl/gcp/src/main/java/co/worklytics/psoxy/VertexGeminiChatModelProvider.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public boolean supports(GenMetadataConfig config) {
+        return config instanceof VertexGenMetadataConfig;
+    }
+
+    @Override
+    public ChatModel create(GenMetadataConfig config, Path modelCacheDir) {
+        VertexGenMetadataConfig vertex = (VertexGenMetadataConfig) config;
```
eschultink: fine; pershaps illegalArgument/state secrption if it's NOT this.

java/impl/gcp/src/main/java/co/worklytics/psoxy/VertexGeminiChatModelProvider.java
```diff
@@ -0,0 +1,89 @@
...
+            .projectId(project)
+            .location(vertex.getModelRegion())
+            .modelName(vertex.getModelId())
+            .thinkingLevel(vertex.getThinkingLevel())
+            .timeout(Duration.ofSeconds(vertex.getTimeoutSeconds()))
+            .maxRetries(1)
+            .build();
+    }
+
+    String resolveProjectId() {
```
eschultink, line 67: don't we ahve GcpHost or something that provides this?

seems like you're reinventing the wheel here.

java/core/src/main/java/co/worklytics/psoxy/impl/gen/GenMetadataConfig.java
```diff
@@ -0,0 +1,136 @@
...
+     * {@link VertexGenMetadataConfig.ConfigProperty}.
+     */
+    public enum ConfigProperty implements ConfigService.ConfigProperty {
+        /** {@code bedrock} or {@code vertex}. Unset defaults to Bedrock. */
+        GEN_METADATA_BACKEND,
+        GEN_METADATA_MODEL,
+        GEN_METADATA_TIMEOUT_SECONDS,
+        GEN_METADATA_RETRIES,
+        /** Set by Terraform {@code enable_gen_metadata = true}. */
+        ENABLE_GEN_METADATA,
```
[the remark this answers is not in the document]
eschultink: review - but I'm OK with that; have it as a terraform level host variable; if genMetadata transforms defined, then if host module did not setup APIS/enable models as required, can have a GEN_METADATA_BACKEND_UNAVAILABLE error or something - fine.

java/impl/aws/src/main/java/co/worklytics/psoxy/aws/AwsModule.java
```diff
@@ -239,6 +242,12 @@ static KmsClient awsKmsClient(AwsEnvironment awsEnvironment) {
     }
 
 
+    @Provides
+    @Singleton
+    static GenMetadataConfig genMetadataConfig(ConfigService configService) {
+        return BedrockGenMetadataConfig.from(configService);
```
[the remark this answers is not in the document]
eschultink: per above, thing I'm fine with that.

java/impl/gcp/src/main/java/co/worklytics/psoxy/GcpModule.java
```diff
@@ -168,6 +171,12 @@ static KeyManagementServiceClient providesKeyManagementServiceClient() {
         return KeyManagementServiceClient.create();
     }
 
+    @Provides
+    @Singleton
+    static GenMetadataConfig genMetadataConfig(ConfigService configService) {
+        return VertexGenMetadataConfig.from(configService);
```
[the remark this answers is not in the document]
eschultink: per above, think I'm ok with that; just need to ensure we expect/handle possible unavaiability errprs

infra/modules/aws-host/main.tf
```diff
@@ -250,6 +272,10 @@ module "api_connector" {
   timeout_seconds               = coalesce(try(each.value.timeout_seconds, null), 180)
   allowed_data_access_ip_blocks = var.allowed_data_access_ip_blocks
 
+  extra_lambda_role_iam_statements = concat(
+    try(each.value.enable_gen_metadata, false) ? local.bedrock_invoke_iam_statements : [],
+  )
```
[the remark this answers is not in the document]
eschultink: or just set this as `local.bedrock_invoke_iam_statements`, if that is always a list, never null.

java/core/src/main/java/co/worklytics/psoxy/storage/impl/CsvRecordWriter.java
```diff
@@ -37,7 +39,7 @@ public void writeRecord(Map<String, Object> record) throws IOException {
         }
         
         for (String header : headers) {
-            printer.print(record.get(header));
+            printer.print(RecordCellSerialization.forTabularCell(record.get(header), objectMapper));
```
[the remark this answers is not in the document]
eschultink: edge case, but reasonable.

java/core/src/main/java/co/worklytics/psoxy/storage/impl/RecordBulkDataSanitizerImpl.java
```diff
@@ -208,7 +213,19 @@ private RecordRules.Format formatFromSourceObjectPath(String sourceObjectPath) {
...
     /**
-     * Apply transforms, then optional output schema filter.
+     * Apply augments then transforms to a single record.
+     */
+    Map<String, Object> sanitizeRecord(Map<String, Object> document,
+                                       List<Triple<JsonPath, RecordTransform, MapFunction>> compiledTransforms)
+            throws UnmatchedPseudonymization {
+        if (ObjectUtils.isNotEmpty(rules.getAugments())) {
+            augmentProcessor.applyAugments(rules.getAugments(), document);
+        }
```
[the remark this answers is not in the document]
eschultink: yes; need to derive potential columns from the rules. Explicitly output empty when no matches

java/impl/gcp/src/main/java/co/worklytics/psoxy/GcsFileEventHandler.java
```diff
@@ -96,6 +99,15 @@ private void process(String importBucket, String sourceName, StorageHandler.Obje
...
             storageHandler.handle(request, transform, inputStreamSupplier, outputStreamSupplier);
+
+            // Apply final metadata including genMetadata token aggregates (S3 builds metadata after
+            // sanitize; GCS must open the writer first, so patch after).
+            Map<String, String> finalMetadata =
+                storageHandler.buildObjectMetadata(importBucket, sourceName, transform);
+            storage.update(BlobInfo.newBuilder(
+                    BlobId.of(request.getDestinationBucketName(), request.getDestinationObjectPath()))
+                .setMetadata(finalMetadata)
+                .build());
```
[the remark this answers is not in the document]
eschultink: can we set the metadata whne intiially creating the file out? that seems most correct

infra/modules/gcp-host/main.tf
```diff
@@ -24,11 +24,13 @@ locals {
...
-  # '_' (e.g. psoxy-dev-erik_GCAL_SOURCE); bucket keys are psoxy-dev-erik/GCAL/rules.yaml.
-  shared_resource_path = "${trimsuffix(trimprefix(local.config_parameter_prefix, "/"), "_")}/"
+  # Object-key prefixes are siblings, not nested: shared is `{env}/` and instance is
+  # `{env}-{INSTANCE}/` so IAM `env/*` does not also match instance objects. Secret IDs still use
+  # config_parameter_prefix with a trailing '_' (e.g. psoxy-dev-erik_GCAL_SOURCE).
+  resource_path_root = trimsuffix(trimprefix(local.config_parameter_prefix, "/"), "_")
+  shared_resource_path = "${local.resource_path_root}/"
   connector_instance_resource_path = { for k, v in merge(var.api_connectors, var.bulk_connectors, var.webhook_collectors) :
-    k => "${local.shared_resource_path}${replace(upper(k), "-", "_")}/"
+    k => "${local.resource_path_root}-${replace(upper(k), "-", "_")}/"
```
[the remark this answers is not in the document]
eschultink: that's OK; no one using those alpha features yet

java/core/src/main/java/co/worklytics/psoxy/impl/gen/LangChain4jGenMetadataBackend.java
```diff
@@ -0,0 +1,347 @@
...
+    private final ObjectMapper objectMapper;
+    private final GenMetadataPromptBudget promptBudget;
+    private final GenMetadataChatModelFactory chatModelFactory;
+    private final GenMetadataTokenUsageAccumulator tokenUsageAccumulator;
+    private final GenMetadataPromptBuilder promptBuilder;
+    private final GenMetadataResponseFormats responseFormats;
+
+    private final ConcurrentHashMap<String, ModelHandle> models = new ConcurrentHashMap<>();
+    private final ExecutorService chatExecutor =
+        Executors.newFixedThreadPool(CLOUD_MAX_CONCURRENT, chatThreadFactory());
```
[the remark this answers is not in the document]
eschultink: can we bound execution path with a max timeout?  (10s, pull from config?)

